PR #2, opened 2022-10-21
Make compile successful 

ChenaLee:


khanhntd (approved):
LGTM.

README.md
```diff
@@ -1,6 +1,4 @@
-## My Project
-
-TODO: Fill this README out!
+## Amazon CloudWatch Agent Test
```
SaxyPandaBear, line 1: I'd prefer if this title was more descriptive. The format that I've always stuck to for README files is:
```
Title of the repo/project
------------------------
Short tag line or elevator pitch for the purpose of the repo. 

## Subheader
Stuff

...
```

go.mod
```diff
@@ -0,0 +1,39 @@
+module github.com/aws/amazon-cloudwatch-agent-test
+
+replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20220502160831-c20ebe67c5ef
```
SaxyPandaBear, line 3: Do we need telegraf here?

go.mod
```diff
@@ -0,0 +1,39 @@
...
+
+go 1.18
+
+require (
+	github.com/aws/aws-sdk-go v1.44.118
```
SaxyPandaBear, line 8: The integration tests don't **need** the v1 SDK. Can you remove this and update imports as necessary? I think there are imports in the tests that use stuff like `aws.String()` but import from the v1 SDK when there are equivalent functions in v2. My e2e mem test PR included a change with that. Can follow that as a guide.

SaxyPandaBear (approved):
LGTM. Minor comments that don't need to be addressed immediately in this PR.

Follow up actions we need to take (not necessarily in this order):
1. Port over the GitHub Actions workflows for integration testing
2. Set up CI checks (build, lint, etc) and automate execution of them in PRs
3. Maybe CODEOWNERS to automate assigned reviewers for PRs
4. Figure out our strategy for when these integration tests will execute (we should NOT run them automatically for all PRs all the time because that allows people to exhaust our resources)

go.mod
ChenaLee: oh this might have been before I removed the dependency to cwa main repo. will retry

go.mod
ChenaLee: same as below

ChenaLee: yeah looks like it works now without telegraf. With agent repo dependency this had to be added

go.mod
ChenaLee: sure I can update

PR #3, opened 2022-10-21
Fix workflow for ec2 linux integration test

ChenaLee:
Changes for the git workflow files in the main repo here: https://github.com/ChenaLee/amazon-cloudwatch-agent/tree/e2e_test_package_separation.
I did thinks those files need to live there because many steps are using SHA of the agent repo commit.


README.md
```diff
@@ -1,8 +1,7 @@
 ## Amazon CloudWatch Agent Test
-
+git
```
SaxyPandaBear, line 2: Don't think this was intentional

terraform/ec2/localstack/main.tf
```diff
@@ -36,18 +36,18 @@ resource "aws_instance" "integration-test" {
...
+      "cd ~/amazon-cloudwatch-agent-test/localstack/ls_tmp",
       "openssl req -new -x509 -newkey rsa:2048 -sha256 -nodes -out snakeoil.pem -keyout snakeoil.key -config snakeoil.conf",
       "cat snakeoil.key snakeoil.pem > server.test.pem",
       "cat snakeoil.key > server.test.pem.key",
       "cat snakeoil.pem > server.test.pem.crt",
-      "cd ~/amazon-cloudwatch-agent/integration/localstack",
+      "cd ~/amazon-cloudwatch-agent-test/localstack",
       "docker-compose up -d --force-recreate",
-      "aws s3 cp ls_tmp s3://${var.s3_bucket}/integration-test/ls_tmp/${var.github_sha} --recursive"
+      "aws s3 cp ls_tmp s3://${var.s3_bucket}/integration-test/ls_tmp/${var.cwa_github_sha} --recursive"
```
SaxyPandaBear, line 50: So how this all worked in the main repo is that part of the workflow builds the binaries and executables and uploads them to S3. This is a prerequisite step to executing the integration tests. How are we planning on keeping integration test executions up to date here when the main agent code lives in the other repo? This terraform is very likely to change since we're ripping it out from where the rest of the agent code is.

README.md
ChenaLee: no, thanks

terraform/ec2/localstack/main.tf
ChenaLee: Not following. main repo github actions steps will check out this repo and use these terraform files. Every time either of them changes if changes are incompatible main repo's github workflow should fail. Am I missing something?

SaxyPandaBear: Okay then I misunderstood how this was going to be used. Makes sense. Disregard.

SaxyPandaBear (approved):

SaxyPandaBear:
Are you manually requesting review from amazon-auto? I think that's incorrect. If we want, we can request review from our GitHub team. We do this in the main repo via CODEOWNERS

ChenaLee:
> Are you manually requesting review from amazon-auto? I think that's incorrect. If we want, we can request review from our GitHub team. We do this in the main repo via CODEOWNERS

Yes, I didn't really know what that was though. How can I make sure people get notifications for review requests? Separate page for that?

PR #9, opened 2022-10-25
Fix test readme for successful ecs tests

ChenaLee:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/terraform/ec2/README.md
+++ b/terraform/ec2/README.md
@@ -240,7 +240,18 @@ Outputs:
     }
   ]
 }
+```  
+Additionally, add these actions for ECS tests to also work
+
 ```
+"ecs:RegisterTaskDefinition"
+"ecs:DescribeTaskDefinition"
+"ecs:UpdateService"
+"ecs:DeregisterTaskDefinition"
+"logs:DeleteLogGroup"
+
+```
+
 11. Once creation is done, go back to the IAM role and attach the policy you just created by searching for the policy name.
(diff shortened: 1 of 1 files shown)


terraform/ec2/README.md
```diff
@@ -240,7 +240,18 @@ Outputs:
     }
   ]
 }
+```  
+Additionally, add these actions for ECS tests to also work
```
khanhntd, line 244: Any reasons why we separate the ECS with general IAM Policies? Since 
* The general policy has some policies [for ECS](https://github.com/aws/amazon-cloudwatch-agent-test/pull/9/files#diff-b39e585f74930033489d2d564915e9ba35b95b0343c84f734ec8787f628596c9R171-R188)
* `logs:DeleteLogGroup` which is used for delete log group can also be used in normal EC2 test. Therefore, should it be placed in general policy ?

If its good, maybe moving this README  to more centralize place or create another centralized README since this README has shared some information for ECS.

ChenaLee: - I'm not sure why these were separate. All policies listed here are based on what made each of my tests fail. Maybe some ecs failures got in when I thought I was only scraping ec2 failures from cloudtrail
- Yep document cleanup overall is needed. Just out of scope for this PR.

SaxyPandaBear (approved):

khanhntd (approved):
LGTM.

PR #11, opened 2022-11-02
Add Basic Sanity Test for ProcStat

khanhntd:
*Issue #, if available:*

*Description of changes:*


.github/workflows/integrationTest.yml
```diff
@@ -1,799 +0,0 @@
-# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-# SPDX-License-Identifier: MIT
-
-name: Run Integration Tests
```
ChenaLee, line 4: This file was never in this repo. Why is it marked deleted?

.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+# Description of the issue
+_Describe the problem or feature in addition to a link to the issues._
+
+# Description of changes
```
ChenaLee, line 4: thanks for all these!

test/metric/query-json.json
```diff
@@ -1,23 +1,24 @@
...
-      },
-      "Period": 60,
-      "Stat": "Average"
+{
+  "MetricDataQueries":[
+    {
+      "Id": "m1",
+      "MetricStat": {
+        "Metric": {
+          "Namespace": "MetricValueBenchmarkTest",
```
ChenaLee, line 7: Did you mean to edit this? I think I actually added this intending to help people debug quickly as an example

test/metric/metric_value_query.go
```diff
@@ -34,9 +35,14 @@ func GetMetricFetcher(metricName string) (MetricValueFetcher, error) {
 }
 
 type MetricValueFetcher interface {
+	// Fetch returns the values associate with the metric within the past 10 minutes
```
ChenaLee, line 38: I understand return types were not exactly typed.. but I'm mostly on the side that believes comments shouldn't be needed if code is written in a certain way and should be avoided. Would it be helpful if return types were []MetricValue or so instead of this comment? ditto for all other comments added here. Some seems very redundant and/or can be addressed by changing method names

test/metric/query-json.json
khanhntd: Yes. It would be easier for them to use 
```
aws cloudwatch get-metric-data --cli-input-json file://query-json.json
```

.github/workflows/integrationTest.yml
khanhntd: Hmm !!! I believe it is there though !!!! https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/.github/workflows/integrationTest.yml

.github/pull_request_template.md
khanhntd: No problem !!!

test/metric/metric_value_query.go
khanhntd: I'm in the same boat with ya. However, I also believe we will never achieve "perfect code" that will help people understand at first sight. Therefore, comments. But yep I agree I intend to change it in "refactor" PR. If everyone agree,I can change it now too !

test/metric/query-json.json
SaxyPandaBear: But if you put the Start and End time here, this becomes stale.

test/metric/metric_value_query.go
SaxyPandaBear: If you agree that the comments are redundant, then you should remove them.

.github/workflows/integrationTest.yml
SaxyPandaBear: I see it in the repo.

SaxyPandaBear (approved):

sky333999 (approved):

PR #13, opened 2022-11-08
Add E2E Sanity Test For CollectD

khanhntd:
*Issue #, if available:*

*Description of changes:*


test/metric_value_benchmark/agent_configs/collectd_config.json
```diff
@@ -0,0 +1,17 @@
...
+      "debug": true,
+      "logfile": ""
+    },
+    "metrics": {
+      "namespace": "MetricValueBenchmarkTest",
+      "append_dimensions": {
+        "InstanceId": "${aws:InstanceId}"
+      },
+      "metrics_collected": {
+        "collectd": {}
```
SaxyPandaBear, line 14: Do the test host AMIs install collectd?

test/metric_value_benchmark/collectd_test.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (t *CollectDTestRunner) getExtraCommands() []string {
+	return []string{
+		"sudo mkdir -p /etc/collectd",
+		"sudo chown `whoami` /etc/collectd",
+		"echo 'user: secret' > /etc/collectd/auth_file",
+		fmt."sudo collectd -C ./extra_configs/collectd.conf",
+	}
+}
```
SaxyPandaBear, line 53: What is this for? Should this be baked into the AMI

test/metric_value_benchmark/base_test.go
```diff
@@ -58,10 +60,17 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
 		},
 	}
 
+	extraCommands := t.testRunner.getExtraCommands()
+	err := util.RunCommands(extraCommands)
+	if err != nil {
+		testGroupResult.TestResults[0].Status = status.FAILED
+		return testGroupResult, fmt.Errorf("Failed to run extra commands due to: %v", err.Error())
```
SaxyPandaBear, line 67: nit - `err.Error()` is a string. You should either change it to `err` and leave `%v` or change it to `%s` and leave `err.Error()`

test/metric_value_benchmark/collectd_test.go
khanhntd: So the [collectd](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-custom-metrics-collectd.html) has the option to send the data with encryption. Therefore, `secret` is used to verify signatures and to decrypt encrypted network packets. IMO, it should not be baked into the AMI since it needs to be configured in both [collectd.conf](https://github.com/aws/amazon-cloudwatch-agent-test/pull/13/files#diff-0c4c6e0c9498578e2b6b38316e2093258e2c060b84f7e83c2bf170e1ac8f21e6R13-R14)

SaxyPandaBear: Do we need to add this for testing?

test/metric_value_benchmark/base_test.go
khanhntd: Nice catch and thanks for reminding me. Fixed it !!!

test/metric_value_benchmark/collectd_test.go
khanhntd: Not entirely if we only consider the `metric value benchmark` However, I'm not seeing the harm of adding it too.

SaxyPandaBear: What is the value for adding this? This adds complexity to our testing

khanhntd: Would be simulating the flow for customers but after discussing in OH, I have deleted it !

test/metric_value_benchmark/base_test.go
```diff
@@ -58,25 +60,32 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
 		},
 	}
 
+	extraCommands := t.testRunner.getExtraCommands()
```
ChenaLee, line 63: this doesn't seem like it belongs to 'runAgent'? It doesn't really have to do with running the agent right..? or did I misunderstand

test/metric_value_benchmark/base_test.go
```diff
@@ -58,25 +60,32 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
...
+	err := util.RunCommands(extraCommands)
```
ChenaLee, line 64: I think getExtraCommands() is too generic. It doesn't really specify what kind of commands you mean and when it will be used. How about replacing it with "SetupBeforeAgentRun", and instead of getting commands, just call util.RunCommands within Setup..() function?

khanhntd: I love generic more than specific case. e.g `runExtraCommands` can run after or before CWA running and for after running CWA, it can stop `collectd`. I can create functions such as `SetupBeforeAgentRun` or `SetupAfterAgentRun` but its too hollow regarding the core functionality.

ChenaLee: But in the template though you only run this before CWA. Names should be easy to understand without new devs having to read where the functions will be used. It's unclear what these are 'extra' for. I think it's almost like naming things "doSomething()" Just a hidden closet to shove in everything that doesn't fit an existing stuff

test/metric_value_benchmark/base_test.go
khanhntd: Yup its has nothing to do with agent !!! That's why I kept it separate from agents and be generic as possible.

test/metric_value_benchmark/base_test.go
khanhntd: Hmm !!! Agree `Names should be easy to understand without new devs having to read where the functions will be used.` . Let's me think about how to refactor it a little bit more.

test/ca_bundle/ca_bundle_test.go
```diff
@@ -49,14 +49,14 @@ func TestBundle(t *testing.T) {
 		//before test run
 		log.Printf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget)
 		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget), func(t *testing.T) {
-			test.ReplaceLocalStackHostName(parameter.dataInput + configJSON)
-			test.CopyFile(parameter.dataInput+configJSON, configOutputPath)
-			test.CopyFile(parameter.dataInput+commonConfigTOML, commonConfigOutputPath)
-			test.StartAgent(configOutputPath, true)
+			util.ReplaceLocalStackHostName(parameter.dataInput + configJSON)
+			util.CopyFile(parameter.dataInput+configJSON, configOutputPath)
```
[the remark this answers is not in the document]
khanhntd: Hmm !!! Would it be better with agent with related functions and /aws package for all functionalities regarding cwl and cwm since I would expect more service such as IMDS or SSM soon.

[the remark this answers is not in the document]
ChenaLee: I don't think I understood. What are you suggesting?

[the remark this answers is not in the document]
khanhntd: So /agent with related functions, /aws for cwl,cwm (and extra services such as ssm)

sethAmazon:
I do not think I added collectd for all os. Please check. I think the collectd component does not work for arm instances. Please check.

test/metric_value_benchmark/collectd_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (t *CollectDTestRunner) setupBeforeAgentRun() error {
+	startCollectdCommands := []string{
+		"sudo mkdir -p /etc/collectd",
+		"sudo collectd -C ./extra_configs/collectd.conf",
+	}
+
+	return common.RunCommands(startCollectdCommands)
+}
```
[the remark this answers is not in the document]
khanhntd: The default configuration for `collectd` would includes many other plugins (e.g CPU, Memory, Interface, etc). Therefore, to limit the amount of metrics we set and less rely on the default enable plugins from collectd, it would better to setup the collectd configuration and setup all these steps beforehand.

test/performancetest/performance_test.go
```diff
@@ -76,9 +76,9 @@ func TestPerformance(t *testing.T) {
 	//run tests
 	for _, tps := range tpsVals {
 		t.Run(fmt.Sprintf("TPS run: %d", tps), func(t *testing.T) {
-			test.CopyFile(configFilePath, configOutputPath)
+			agent.CopyFile(configFilePath, configOutputPath)
```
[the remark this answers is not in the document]
khanhntd: Yes. It should be. Nice catch !!!

test/metric_value_benchmark/extra_configs/collectd.conf
```diff
@@ -0,0 +1,12 @@
+LoadPlugin network
```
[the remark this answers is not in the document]
khanhntd: This would be the [same reasons with the above questions though ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/13#discussion_r1035376992). 
```
to limit the amount of metrics we set and less rely on the default enable plugins from collectd, it would better to setup the collectd configuration, setup all these steps beforehand and provide more flexibility when testing collectd
```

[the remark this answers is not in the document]
SaxyPandaBear: This doesn't explain why we can't configure collectd in the image build.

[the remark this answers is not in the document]
khanhntd: In more simpler way, would it be better with `provide more flexibility when testing collectd` ? e.g testing with different plugins with collectd ? If that reason is not convinced enough, then I'm fine with configure collectd in the image build.

test/metric_value_benchmark/base_test.go
```diff
@@ -65,23 +67,29 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
...
+	common.CopyFile(agentConfigPath, configOutputPath)
+	err := common.StartAgent(configOutputPath, false)
 
 	if err != nil {
 		testGroupResult.TestResults[0].Status = status.FAILED
-		return testGroupResult, fmt.Errorf("Agent could not start due to: %v", err.Error())
+		return testGroupResult, fmt.Errorf("Agent could not start due to: %s", err.Error())
+	}
+
+	err = t.testRunner.setupAfterAgentRun()
```
SaxyPandaBear, line 78: When we introduce this, we increase the time it takes before we sleep. I don't think that will cause a big issue, but it's something that we have to keep in mind. There are a lot of tests now, they each take some time, and they all run sequentially.

test/metric_value_benchmark/collectd_test.go
```diff
@@ -0,0 +1,98 @@
...
+	// 		<Plugin cpu>
+	//			ReportByState = true
+	//			ReportByCpu = true
+	//			ValuesPercentage = true
+	//		</Plugin>
+	//		<Plugin network>
+	//			Server "127.0.0.1" "25826"
+	//		</Plugin>
+	startCollectdCommands := []string{
+		"sudo systemctl restart collectd",
```
SaxyPandaBear, line 63: Why do we need to restart collectd?

test/run_as_user/run_as_user_test.go
```diff
@@ -40,15 +40,23 @@ func TestRunAsUser(t *testing.T) {
...
+			common.CopyFile(parameter.dataInput, configOutputPath)
+			common.StartAgent(configOutputPath, true)
 			time.Sleep(agentRuntime)
 			log.Printf("Agent has been running for : %s", agentRuntime.String())
 			// Must read the pid file while agent is running
-			pidOutput := test.RunCommand(test.CatCommand + pidFile)
-			agentOwnerOutput := test.RunCommand(test.AppOwnerCommand + pidOutput)
+			pidOutput, err := common.RunCommand(common.CatCommand + pidFile)
+			if err != nil {
+				t.Fatalf("Error: %v", err)
```
SaxyPandaBear, line 50: Why `Fatal` here but `Error` below?

test/metric_value_benchmark/base_test.go
khanhntd: True. That's why I have some ideas of improving it and align with Chad's test plan. But I guess its not worth to do here

test/run_as_user/run_as_user_test.go
khanhntd: To be honest, I'm not sure. The `RunCommand` previously return Fatal so I preserve the order. However, in the end, `Fatalf` and `Errorf` will produce the same results but Fatal with fail earlier so its better to use Fatal here

test/metric/procstat.go
```diff
@@ -24,7 +24,7 @@ func (f *ProcStatMetricValueFetcher) Fetch(namespace, metricName string, stat St
 	dims = append(dims, f.getInstanceIdDimension())
 	values, err := f.fetch(namespace, metricName, dims, stat)
 	if err != nil {
-		log.Printf("Error while fetching metric value for %s: %v", metricName, err)
+		log.Printf("Error while fetching metric value for %s: %s", metricName, err.Error())
```
sky333999, line 26: For my understanding, why is this preferred?

test/metric/processes.go
```diff
@@ -0,0 +1,57 @@
...
+
+package metric
+
+import (
+	"log"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+)
+
+type ProcessesMetricValueFetcher struct {
```
sky333999: I thought processes was already merged.. does this need to be rebased?

test/metric/collectd.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+)
+
+type CollectDMetricValueFetcher struct {
+	baseMetricValueFetcher
+}
+
+var _ MetricValueFetcher = (*CollectDMetricValueFetcher)(nil)
+
+func (f *CollectDMetricValueFetcher) Fetch(namespace, metricName string, stat Statistics) (MetricValues, error) {
```
sky333999, line 22: This seems common for all plugins... does this ever differ for any plugin?
Same question for `isApplicable`

test/metric_value_benchmark/collectd_test.go
sky333999: Related, do we care about stopping it towards the end of the test run?

khanhntd: At the end, agent will receive new config. Therefore, it does not matter for me IMO. No particular reason, i guess I am only want to make sure collectd is starting after all. And if its fails, its will return some error during running github action.

test/metric/net.go
```diff
@@ -20,10 +20,11 @@ type NetMetricValueFetcher struct {
...
 
 func (f *NetMetricValueFetcher) Fetch(namespace, metricName string, stat Statistics) (MetricValues, error) {
-	dims := f.getMetricSpecificDimensions()
-	values, err := f.fetch(namespace, metricName, dims, stat)
+	dimensions := f.getMetricSpecificDimensions()
+	dimensions = append(dimensions, f.getInstanceIdDimension())
+	values, err := f.fetch(namespace, metricName, dimensions, stat)
 	if err != nil {
-		log.Printf("Error while fetching metric value for %s: %v", metricName, err)
+		log.Printf("Error while fetching metric value for %v: %v", metricName, err.Error())
```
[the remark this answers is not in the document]
khanhntd: No. Somehow I got rebase and forgot to fix this.

test/metric/procstat.go
khanhntd: Its cleaner for me IMO. I don't mind using `err`.

test/metric/processes.go
khanhntd: Not sure why its show me up here to. I have already rebased and merged though hmmm.

khanhntd: Its weird that I have rebased and the diff PR commits does reflect it but now here https://github.com/aws/amazon-cloudwatch-agent-test/compare/main...khanhntd:amazon-cloudwatch-agent-test:collectd_sanity

test/metric/collectd.go
SaxyPandaBear: I agree that they're redundant. I haven't put a lot of thought into how to consolidate, but I think that `isApplicable` is the easier one to clean up since we already have a list of supported metrics that we have to implement in the fetcher interface. We _should_ be able to just assume that `isApplicable` is consistent across all implementations.

test/metric/procstat.go
SaxyPandaBear: Can you explain why you think it's cleaner

SaxyPandaBear (approved):

sky333999 (approved):

PR #26, opened 2022-11-17
First ECS metric sanity test - ECS EC2 Launch Type Daemon Deployment - Container Insights

ChenaLee:
# Description of the issue
ECS metric test needs to be added

## Out of scope
- Some duplicated code for now between ECS and EC2 test. This will be removed in future commits. Trying to push what already works.
- Print result summary stopped working after adding init() for flag parsing. Will look into it and fix.

# Description of changes
Added container insights sanity test for ecs-ec2launchtype-daemondeployment

# License

# Tests
https://github.com/ChenaLee/amazon-cloudwatch-agent/actions/runs/3482812632/jobs/5827664654


environment/MetaData.go
```diff
@@ -0,0 +1,96 @@
...
+	flag.StringVar(&(dataString.EcsDeploymentStrategy), "ecsDeploymentStrategy", "", "Daemon/Replica/Sidecar")
+	flag.StringVar(&(dataString.EcsClusterArn), "clusterArn", "", "Used to restart ecs task to apply new agent config")
+	flag.StringVar(&(dataString.CwagentConfigSsmParamName), "cwagentConfigSsmParamName", "", "Used to set new cwa config")
+	flag.StringVar(&(dataString.EcsServiceName), "cwagentECSServiceName", "", "Used to restart ecs task to apply new agent config")
+}
+
+func fillComputeType(e *MetaData, data *MetaDataStrings) *MetaData {
+	computeType, ok := compute_type.FromString(data.ComputeType)
+	if !ok {
+		panic("Invalid compute type. Needs to be EC2/ECS/EKS. Compute Type is a required flag. :" + data.ComputeType)
```
khanhntd, line 51: **Nitpick**: We can use `log.panic` so it would have a better log and it also writes the logs to cloudwatch agent log too

environment/MetaData.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	ecsDeploymentStrategy, ok := ecs_deployment_type.FromString(data.EcsDeploymentStrategy)
+	if !ok {
+		log.Printf("Invalid deployment strategy %s. This might be because it wasn't provided for non-ECS tests", data.ComputeType)
+	} else {
+		e.EcsDeploymentStrategy = ecsDeploymentStrategy
+	}
+
+	e.EcsClusterArn = data.EcsClusterArn
```
khanhntd, line 76: For these meatadata, I would recommend use [ECS Metadata for better automated](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html)

environment/compute_type/compute_type.go
```diff
@@ -0,0 +1,24 @@
+package compute_type
```
khanhntd, line 1: **Nit pick**: License !!!!

terraform/ecs/linux/ec2_launch/daemon/iam.tf
```diff
@@ -0,0 +1,126 @@
+resource "aws_iam_role" "ecs_task_role" {
```
khanhntd, line 1: Welcome to Terraform World !!!

test/ssmParameter.go
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: MIT
+
+//go:build integration
+// +build integration
+
+package test
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ssm"
```
khanhntd, line 11: Would it be better to use SSM v2 ?

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -32,17 +33,61 @@ func (suite *MetricBenchmarkTestSuite) TearDownSuite() {
...
+// DONE: ec2 & ecs whole test
+// DONE: merge conflict resolution
+//TODO: test e2e
+// TODO: make CWA package not use personal test branch
+// TODO: maybe after this I can make a PR before coveredTestList cleanup. Make it simple & static for test list.
+// TODO: test this runAgentStrategy and then if it works, refactor the ec2 ones with this too. -> no do this later?
+// TODO: remove coveredTestList into another branch to separate PR
+//TODO: coveredTestList needs to be cleaned up. See my handwritten notes for ideas. (Todo)
+// Based on the above, make a factory.
+// Do this only for ecs for now, and a separate PR for ec2 changes? nah..not possible
```
khanhntd, line 60: Self-conflicting moment.

test/metric_value_benchmark/container_insights_test.go
```diff
@@ -0,0 +1,79 @@
...
+	return "./agent_configs/container_insights.json"
+}
+
+func (t *ContainerInsightsTestRunner) getAgentRunDuration() time.Duration {
+	return minimumAgentRuntime
+}
+
+func (t *ContainerInsightsTestRunner) getMeasuredMetrics() []string {
+	return []string{
+		"instance_memory_utilization", "instance_number_of_running_tasks", "instance_memory_reserved_capacity",
```
khanhntd, line 49: Would it be better to leave comments here with the public document that talks about metrics that CWAgent is gathering

test/ecs.go
```diff
@@ -0,0 +1,117 @@
...
+		return []ContainerInstance{}, err
+	}
+
+	describeContainerInstancesOutput, err := describeContainerInstances(clusterArn, containerInstanceArns)
+	if err != nil {
+		return []ContainerInstance{}, err
+	}
+
+	results := []ContainerInstance{}
+	for _, containerInstance := range describeContainerInstancesOutput.ContainerInstances {
```
khanhntd, line 60: Would it better to have an internal that calls [IMDS (from EC2)](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) and also [metadata from ECS](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html) so it would better to have less specific calls ?

khanhntd (commented):
LGTM overall to me and ignores many things that can be done in the next PR.

environment/ecs_launch_type/ecs_launch_type.go
```diff
@@ -0,0 +1,22 @@
+package ecs_launch_type
```
SaxyPandaBear, line 1: Nit - I don't think packages tend to include `_` or `-`. Kubernetes as an example https://github.com/kubernetes/kubernetes/tree/master/pkg

environment/MetaData.go
```diff
@@ -0,0 +1,96 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build integration
```
SaxyPandaBear, line 4: One thing I wanted to ask and doesn't necessarily pertain specifically to this PR: Do we still need the integration build tag? This is a separate package now, so I feel like this is unnecessary. We don't have to make that change in this PR but wanted to ask.

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 )
 
 require (
+	github.com/aws/aws-sdk-go v1.44.131 // indirect
```
SaxyPandaBear, line 21: Don't think we need to pull in v1 of the SDK here.

terraform/ecs/linux/ec2_launch/daemon/main.tf
```diff
@@ -0,0 +1,229 @@
...
+  name = "cwagent-integ-test-cluster-${random_id.testing_id.hex}"
+}
+
+######
+# EC2 for ECS EC2 launch type
+######
+data "aws_ami" "latest" {
+  most_recent = true
+  owners = ["amazon"]
+  #owners      = ["self", "506463145083"]
```
SaxyPandaBear, line 16: we don't need the comment here

terraform/ecs/linux/ec2_launch/daemon/variables.tf
```diff
@@ -0,0 +1,29 @@
+variable "region" {
+  type    = string
+  default = "us-west-2"
+}
+
+variable "cwagent_image_repo" {
+  type    = string
+  default = "public.ecr.aws/cloudwatch-agent/cloudwatch-agent"
```
SaxyPandaBear, line 8: So does this test against the public, globally deployed agent image or a local docker image build from the test pipeline? I tried looking for substitutions that point to our own testing image and didn't find any

test/ecs.go
```diff
@@ -0,0 +1,117 @@
...
+// SPDX-License-Identifier: MIT
+
+//go:build integration
+// +build integration
+
+package test
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecs"
```
SaxyPandaBear, line 11: There should be Go SDK V2 equivalents for this. There might be code changes necessary, but I don't think there's a good reason to include both SDKs

test/metric/metric_value_query.go
```diff
@@ -52,17 +58,34 @@ type MetricValueFetcher interface {
...
+}
 
-func (f *baseMetricValueFetcher) fetch(namespace, metricName string, metricSpecificDimensions []types.Dimension, stat Statistics) (MetricValues, error) {
+func (f *baseMetricValueFetcher) getEnv() *environment.MetaData {
+	return f.Env
+}
+
+func (f *baseMetricValueFetcher) setEnv(env *environment.MetaData) {
+	f.Env = env
+}
```
SaxyPandaBear, line 74: you should be able to just do `fetcher.Env = f.Env` above. I don't think that we need the getter/setter model here.

test/metric/metric_value_query.go
```diff
@@ -11,23 +11,29 @@ import (
...
 	&DiskIOMetricValueFetcher{},
+	&ContainerInsightsValueFetcher{},
+}
+
+type MetricFetcherFactory struct {
+	Env *environment.MetaData
 }
 
-func GetMetricFetcher(metricName string) (MetricValueFetcher, error) {
+func (f *MetricFetcherFactory) GetMetricFetcher(metricName string) (MetricValueFetcher, error) {
```
SaxyPandaBear, line 33: nit - I think it's confusing to use the same variable name for methods on different structs. `f` is used for `MetricFetchFactory` and `baseMetricValueFetcher` in the same file

test/metric/procstat.go
```diff
@@ -21,6 +21,7 @@ var _ MetricValueFetcher = (*ProcStatMetricValueFetcher)(nil)
 
 func (f *ProcStatMetricValueFetcher) Fetch(namespace, metricName string, stat Statistics) (MetricValues, error) {
 	dims := f.getMetricSpecificDimensions()
+	dims = append(dims, f.getInstanceIdDimension())
```
SaxyPandaBear, line 24: Do we always want to pull in metric dimensions as well as instance dimensions? If so, should we just go back to the interface and consolidate there so we don't add `append` everywhere?

So `getDimensions()` instead of `getMetricSpecificDimensions()` + `getInstanceIdDimensions()`?

test/metric_value_benchmark/base_test.go
```diff
@@ -34,6 +35,10 @@ type TestRunner struct {
 	testRunner ITestRunner
 }
 
+type BaseTestRunner struct {
+	MetricFetcherFactory *metric.MetricFetcherFactory
```
SaxyPandaBear, line 39: We can drop the variable name here

test/metric_value_benchmark/covered_test_list.go
```diff
@@ -0,0 +1,71 @@
...
+	"mem_inactive":          HostMetricCommonCoverage,
+	"mem_total":             HostMetricCommonCoverage,
+	"mem_used":              HostMetricCommonCoverage,
+	"mem_used_percent":      HostMetricCommonCoverage,
+}
+
+func getEcsDaemonMetricsToTest() []string {
+	metricsToTest := []string{}
+	for metricName, setupTypeTestCoverage := range supportedMetricsToTestCoverage {
+		if setupTypeTestCoverage.EcsDaemon == TEST_EXISTS {
```
SaxyPandaBear, line 66: What is the purpose of this support matrix if we're always filtering on ECS daemon set?

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -32,17 +33,61 @@ func (suite *MetricBenchmarkTestSuite) TearDownSuite() {
 	fmt.Println(">>>> Finished MetricBenchmarkTestSuite")
 }
 
-var testRunners = []*TestRunner{
-	{testRunner: &CPUTestRunner{}},
-	{testRunner: &MemTestRunner{}},
-	{testRunner: &ProcStatTestRunner{}},
-	{testRunner: &DiskIOTestRunner{}},
+// [DONE] make tests run
```
SaxyPandaBear, line 36: Do we need this comment block?

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -32,17 +33,61 @@ func (suite *MetricBenchmarkTestSuite) TearDownSuite() {
...
+	environment.RegisterEnvironmentMetaDataFlags(envMetaDataStrings)
 }
 
 func (suite *MetricBenchmarkTestSuite) TestAllInSuite() {
-	for _, testRunner := range testRunners {
-		testRunner.Run(suite)
+	env := environment.GetEnvironmentMetaData(envMetaDataStrings)
+	if env.ComputeType == compute_type.ECS {
+		log.Printf("Environment compute type is ECS")
+		var ecsTestRunners = []*ECSTestRunner{
```
SaxyPandaBear, line 72: should we define ECS and EC2 test runners at the top level? Also
```suggestion
		ecsTestRunners := []*ECSTestRunner{
```

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -32,17 +33,61 @@ func (suite *MetricBenchmarkTestSuite) TearDownSuite() {
...
+			{testRunner: &ContainerInsightsTestRunner{ECSBaseTestRunner{MetricFetcherFactory: &metric.MetricFetcherFactory{Env: env}}},
+				agentRunStrategy: &ECSAgentRunStrategy{}},
+		}
+		for _, ecsTestRunner := range ecsTestRunners {
+			ecsTestRunner.Run(suite, env)
+		}
+	} else {
+		var testRunners = []*TestRunner{
```
SaxyPandaBear, line 80: We don't need to declare as a var since we are instantiating it anyway
```suggestion
		testRunners := []*TestRunner{
```

test/metric/container_insights.go
```diff
@@ -0,0 +1,71 @@
...
+	//TODO currently assuming there's only one container
+	containerInstances, err := test.GetContainerInstances(&(f.getEnv().EcsClusterArn))
+	if err != nil {
+		log.Print(err)
+		return []types.Dimension{}
+	}
+
+	return []types.Dimension{
+		{
+			Name:  aws.String("ClusterName"),
```
SaxyPandaBear, line 59: Note that the type for `Name` is a `*string`, and the purpose of the `aws.String()` function is to return the pointer reference. This is used heavily and we can do this instead of passing in the pointer value directly as parameters to functions

test/metric_value_benchmark/container_insights_test.go
ChenaLee: Imo this is the source for public docs, not the other way around. I don't mind it but why do you think that will be helpful?

environment/ecs_launch_type/ecs_launch_type.go
ChenaLee: hmm gotcha

ChenaLee: [TODO]

environment/MetaData.go
ChenaLee: no, I was going to do that after test package separation and then got lazy. Will have to change in both public/private repo + in test repo. Thought I could punt that in favor of adding more tests & delivering alpha :D

go.mod
ChenaLee: oh idk what happened? it's indirect so must be something's dependency.

terraform/ecs/linux/ec2_launch/daemon/main.tf
ChenaLee: oooops [TODO]

terraform/ecs/linux/ec2_launch/daemon/variables.tf
ChenaLee: by default yeah looks like it. It should never be default though. This value is passed in from workflow: https://github.com/aws/amazon-cloudwatch-agent/pull/639/files#diff-9b9f9a38b889c9dcb55ba0330baa69c8fdd62ac1100b6eb59b6fc44008f5d320R727

test/ecs.go
ChenaLee: will try [TODO]

test/metric/metric_value_query.go
ChenaLee: where? I tried different things to get Env from all the fetchers, and making getEnv() was the only way. direct access to variable failed which is.. I guess reasonable

test/metric/metric_value_query.go
ChenaLee: I can change that [TODO]

test/metric/procstat.go
ChenaLee: We don't. Well this got moved in here because of ECS test. And now I'm actually realizing some configurations don't have instance id as dimension. (If append-dimension isn't specified). I'm in process of adding a simple test for behavior without append-dimension, and will have some modifications for this dimension logic while doing that. Mind if we punt this convo til my test PR?

test/metric_value_benchmark/base_test.go
ChenaLee: oh ok didn't know that's a thing [TODO]

test/metric_value_benchmark/covered_test_list.go
ChenaLee: uhhh idk why this file is still here. This isn't used :D.. this is for the future.

test/metric_value_benchmark/metrics_value_benchmark_test.go
ChenaLee: ty [TODO]

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -32,17 +33,61 @@ func (suite *MetricBenchmarkTestSuite) TearDownSuite() {
...
+// TODO: remove coveredTestList into another branch to separate PR
+//TODO: coveredTestList needs to be cleaned up. See my handwritten notes for ideas. (Todo)
+// Based on the above, make a factory.
+// Do this only for ecs for now, and a separate PR for ec2 changes? nah..not possible
+
+var envMetaDataStrings = &(environment.MetaDataStrings{})
+
+func init() {
+	environment.RegisterEnvironmentMetaDataFlags(envMetaDataStrings)
 }
```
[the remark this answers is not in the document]
ChenaLee: Yeah :( This is literally the only way. Where are you suggesting? I could get logs of function just fine. Trust me trying various things for this probably wasted the most amount of my time during this PR. Another way to do it is to just... kinda what we were doing before, and copy and paste in other (if any in the future) test suites, and make some mess trying to make it into a struct.

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	agentConfig := string(b)
+	ssmParamType := "String"
+
+	err = test.PutParameter(&(e.CwagentConfigSsmParamName), &agentConfig, &ssmParamType)
+	if err != nil {
+		return fmt.Errorf("Failed while reading config file : %s", err.Error())
+	}
+	log.Printf("Put parameter successful")
+
+	err = test.RestartDaemonService(&(e.EcsClusterArn), &(e.EcsServiceName))
```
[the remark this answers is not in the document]
ChenaLee: Really good point. Unclear to me. Couldn't find public docs about that. So far this hasn't failed to use the new config I'm putting, so I have to guess this might support strongly consistent read.

test/ssmParameter.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func PutParameter(name *string, value *string, paramType *string) error {
```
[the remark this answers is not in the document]
ChenaLee: 1. oh wooooow that's very cool [TODO]
2. just my laziness  :D I didn't really think much about it. do you hate it?

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	runAgent(e *environment.MetaData, configFilePath string) error
+}
+
+type ECSAgentRunStrategy struct {
+}
+
+func (r *ECSAgentRunStrategy) runAgent(e *environment.MetaData, configFilePath string) error {
+	b, err := os.ReadFile(configFilePath)
+	if err != nil {
+		return fmt.Errorf("Failed while reading config file")
```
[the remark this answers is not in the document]
ChenaLee: [TODO]

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed while reading config file : %s", err.Error())
+	}
+	log.Printf("Put parameter successful")
+
+	err = test.RestartDaemonService(&(e.EcsClusterArn), &(e.EcsServiceName))
+	if err != nil {
+		fmt.Print(err)
+	}
+	log.Printf("CWAgent service is restarted")
```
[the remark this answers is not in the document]
ChenaLee: [TODO]

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,105 @@
...
+				Status: status.FAILED,
+			},
+		},
+	}
+
+	err := t.agentRunStrategy.runAgent(e, t.testRunner.getAgentConfigFileName())
+
+	if err != nil {
+		fmt.Print(err)
+		return testGroupResult, fmt.Errorf("Failed to run agent with config for the given test")
```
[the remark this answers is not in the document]
ChenaLee: [TODO]

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,105 @@
...
+				Name:   "Starting Agent",
+				Status: status.FAILED,
+			},
+		},
+	}
+
+	err := t.agentRunStrategy.runAgent(e, t.testRunner.getAgentConfigFileName())
+
+	if err != nil {
+		fmt.Print(err)
```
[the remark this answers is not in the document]
ChenaLee: ok :( [TODO]

test/metric_value_benchmark/covered_test_list.go
```diff
@@ -0,0 +1,71 @@
...
+type SetupTypeTestCoverage struct {
+	EcsDaemon  TestCoverageStatus
+	EcsSidecar TestCoverageStatus
+	EcsReplica TestCoverageStatus
+	Ec2        TestCoverageStatus
+}
+
+var HostMetricCommonCoverage = SetupTypeTestCoverage{EcsDaemon: TEST_EXISTS, EcsSidecar: TEST_EXISTS, EcsReplica: UNSUPPORTED_FEATURE, Ec2: TEST_EXISTS}
+
+var supportedMetricsToTestCoverage = map[string]SetupTypeTestCoverage{
```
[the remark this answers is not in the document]
ChenaLee: let's just ignore this file for now. It's not used and I have some thoughts on how to clean this up. I thought I deleted this file from this branch but I must have not. I will remove [TODO]

test/metric_value_benchmark/container_insights_test.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return testResult
+	}
+
+	if !isAllValuesGreaterThanOrEqualToZero(metricName, values) {
+		return testResult
+	}
+
+	// TODO: Range test with >0 and <100
+	// TODO: Range test: which metric to get? api reference check. should I get average or test every single datapoint for 10 minutes? (and if 90%> of them are in range, we are good)
```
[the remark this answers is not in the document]
ChenaLee: yeahhh we should do some kinda range test whenever possible

test/ecs.go
```diff
@@ -0,0 +1,117 @@
...
+	return listContainerInstancesOutput.ContainerInstanceArns, nil
+}
+
+func GetContainerInstanceId(containerInstanceArn *string) string {
+	return strings.Split(*containerInstanceArn, "/")[2]
+}
+
+func GetClusterName(clusterArn *string) string {
+	return strings.Split(*clusterArn, ":cluster/")[1]
+}
```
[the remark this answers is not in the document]
ChenaLee: ditto above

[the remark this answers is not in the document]
ChenaLee: [TODO]

test/metric/container_insights.go
```diff
@@ -0,0 +1,71 @@
...
+	baseMetricValueFetcher
+}
+
+var _ MetricValueFetcher = (*ContainerInsightsValueFetcher)(nil)
+
+func (f *ContainerInsightsValueFetcher) Fetch(namespace, metricName string, stat Statistics) (MetricValues, error) {
+	dimensions := f.getMetricSpecificDimensions()
+	values, err := f.fetch(namespace, metricName, dimensions, stat)
+	if err != nil {
+		log.Printf("Error while fetching metric value for %v: %v", metricName, err.Error())
```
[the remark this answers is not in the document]
ChenaLee: [TODO]

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	agentConfig := string(b)
+	ssmParamType := "String"
+
+	err = test.PutParameter(&(e.CwagentConfigSsmParamName), &agentConfig, &ssmParamType)
+	if err != nil {
+		return fmt.Errorf("Failed while reading config file : %s", err.Error())
+	}
+	log.Printf("Put parameter successful")
```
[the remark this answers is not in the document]
ChenaLee: [TODO]

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -32,17 +33,61 @@ func (suite *MetricBenchmarkTestSuite) TearDownSuite() {
...
 
 func (suite *MetricBenchmarkTestSuite) TestAllInSuite() {
-	for _, testRunner := range testRunners {
-		testRunner.Run(suite)
+	env := environment.GetEnvironmentMetaData(envMetaDataStrings)
+	if env.ComputeType == compute_type.ECS {
+		log.Printf("Environment compute type is ECS")
```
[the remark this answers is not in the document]
ChenaLee: [TODO]

environment/MetaData.go
```diff
@@ -0,0 +1,96 @@
...
+	EcsClusterArn             string
+	CwagentConfigSsmParamName string
+	EcsServiceName            string
+}
+
+func registerComputeType(dataString *MetaDataStrings) {
+	flag.StringVar(&(dataString.ComputeType), "computeType", "", "EC2/ECS/EKS")
+}
+func registerECSData(dataString *MetaDataStrings) {
+	flag.StringVar(&(dataString.EcsLaunchType), "ecsLaunchType", "", "EC2 or Fargate")
```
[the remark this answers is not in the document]
ChenaLee: Other possible options failed and I didn't wanna touch it after it finally worked after.. a couple days of trying to make this work. We have other data like instance stuff that does need to be passed in iirc. Configuration file doesn't necessarily specify environment stuff either

environment/MetaData.go
```diff
@@ -0,0 +1,96 @@
...
+func fillComputeType(e *MetaData, data *MetaDataStrings) *MetaData {
+	computeType, ok := compute_type.FromString(data.ComputeType)
+	if !ok {
+		panic("Invalid compute type. Needs to be EC2/ECS/EKS. Compute Type is a required flag. :" + data.ComputeType)
+	}
+	e.ComputeType = computeType
+	return e
+}
+
+func fillECSData(e *MetaData, data *MetaDataStrings) *MetaData {
```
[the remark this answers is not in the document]
ChenaLee: I know it looks stupid. I initially tried to do it in one method but had to be done this way because of how flags registration works. It's like registration needs to happen first, and then variables get passed in when tests actually start. So if you try to do it both in init(), you have no actual variable, they are all empty, so turning it into these enums fail.

I know. I hated it too.

test/metric/container_insights.go
ChenaLee: That's a good point. Will do [TODO]

environment/MetaData.go
ChenaLee: oh tyty [TODO]

environment/MetaData.go
ChenaLee: ah yeah I don't think this would work? These variables in the link exists in containers, but our tests don't run in containers.

environment/compute_type/compute_type.go
ChenaLee: ooops [TODO]

terraform/ecs/linux/ec2_launch/daemon/iam.tf
ChenaLee: huh?

test/ssmParameter.go
ChenaLee: yeaah Andrew had a similar comment. I will have to go over every non v2 stuff [TODO]

test/ecs.go
ChenaLee: These tests don't run on container instances. Agent runs there but tests execute from github's own runners.

test/metric/container_insights.go
```diff
@@ -0,0 +1,71 @@
...
+		"instance_filesystem_utilization":   {},
+		"instance_network_total_bytes":      {},
+		"instance_cpu_utilization":          {},
+		"instance_cpu_reserved_capacity":    {},
+	}
+}
+
+func (f *ContainerInsightsValueFetcher) isApplicable(metricName string) bool {
+	_, exists := f.getPluginSupportedMetric()[metricName]
+	return exists
```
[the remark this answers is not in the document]
ChenaLee: Nvm getPluginSupporterMetric() isn't accessible from base fetcher

test/ecs.go
khanhntd: We can run that. However, its would not be the same case for Fargate (but based on the aim for this PR, should we only focus on EC2).

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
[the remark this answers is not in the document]
ChenaLee: Errorf is the only option for fmt

ChenaLee (commented):
Addressed comments

test/metric/metric_value_query.go
```diff
@@ -11,24 +11,30 @@ import (
...
-	"github.com/aws/amazon-cloudwatch-agent-test/test"
 )
 
 var metricValueFetchers = []MetricValueFetcher{
 	&CPUMetricValueFetcher{},
 	&MemMetricValueFetcher{},
 	&ProcStatMetricValueFetcher{},
 	&DiskIOMetricValueFetcher{},
-	&NetMetricValueFetcher{},
+  &NetMetricValueFetcher{},
```
khanhntd, line 26: **Nit**: Space

terraform/ecs/linux/ec2_launch/daemon/main.tf
```diff
@@ -0,0 +1,228 @@
...
+  network_configuration {
+    security_groups  = [aws_security_group.ecs_security_group.id]
+    subnets          = toset(data.aws_subnets.default.ids)
+    assign_public_ip = true
+  }
+
+  depends_on = [aws_iam_role_policy_attachment.ecs_task_execution_role]
+}
+
+resource "null_resource" "validator" {
```
khanhntd, line 219: We can change to `ssh` into EC2 instead of `local-exe` that similar to other EC2 test. Since there are no use case that shares between EC2 and Fargate (except Prometheus)

test/metric/container_insights.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+)
+
+type ContainerInsightsValueFetcher struct {
+	baseMetricValueFetcher
+}
+
+var _ MetricValueFetcher = (*ContainerInsightsValueFetcher)(nil)
+
+func (f *ContainerInsightsValueFetcher) Fetch(namespace, metricName string, stat Statistics) (MetricValues, error) {
```
khanhntd, line 23: **Nit**: Namespace should be unique for container insights (e.g AWS/ECS/ContainerInsights). Therefore, should we strive for a function call (determine Namespace based on EKS/ECS/EC2)

test/metric/metric_value_query.go
khanhntd: TBH, I don't like `getter/setter` model since its too hollow for me and does not give any value much. IIRC you can access variable as long  as its uppercase when outside of function.

test/metric/metric_value_query.go
```diff
@@ -53,17 +59,34 @@ type MetricValueFetcher interface {
...
+
+func (f *baseMetricValueFetcher) setEnv(env *environment.MetaData) {
+	f.Env = env
+}
+
+func (f *baseMetricValueFetcher) getInstanceIdDimension() types.Dimension {
 	ec2InstanceId := test.GetInstanceId()
-	instanceIdDimension := types.Dimension{
+
+	//TODO For now they can stay. Later host metrics fetchers might need to be flexible on how to get instance Id
```
khanhntd, line 80: IMO, it should be the same considers its both EC2.

test/metric_value_benchmark/base_test.go
```diff
@@ -34,6 +35,10 @@ type TestRunner struct {
 	testRunner ITestRunner
 }
 
+type BaseTestRunner struct {
+	*metric.MetricFetcherFactory
```
khanhntd, line 39: Should it be `IECSTestRunner` ?

test/metric_value_benchmark/container_insights_test.go
khanhntd: It would be better for new dev to know what we support for customers currently. So if the values are mismatch, we can instantly fixing it.

test/ssmParameter.go
```diff
@@ -0,0 +1,54 @@
...
+		Overwrite: &isOverwriteAllowed,
+	})
+
+	return err
+}
+
+func getSsmClient() (*ssm.Client, context.Context, error) {
+	if ssmClient == nil {
+		ssmCtx = context.Background()
+		cfg, err := config.LoadDefaultConfig(ssmCtx)
```
khanhntd, line 46: IMO, the others client (e.g [cloudwatch](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/cwm_util.go#L73)) should share the same configuration without any different (e.g credentials, etc)

test/ssmParameter.go
```diff
@@ -0,0 +1,54 @@
...
+var (
+	ssmCtx    context.Context
+	ssmClient *ssm.Client
+)
+
+func PutStringParameter(name, value string) error {
+	return putParameter(name, value, types.ParameterTypeString)
+}
+
+func putParameter(name string, value string, paramType types.ParameterType) error {
```
khanhntd, line 26:
```suggestion
func putParameter(name, value string, paramType types.ParameterType) error {
```

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -0,0 +1,103 @@
...
+	"os"
+	"time"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/environment"
+	"github.com/aws/amazon-cloudwatch-agent-test/test"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/metric"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/status"
+)
+
+type IECSTestRunner interface {
```
khanhntd, line 20: This share a lots with `basedTestRunner`. Should we create a common interface and `runStrategy` can be a part of it

test/metric_value_benchmark/container_insights_test.go
SaxyPandaBear: Fixing it here doesn't fix it in the agent. This is just testing.

test/metric_value_benchmark/metrics_value_benchmark_test.go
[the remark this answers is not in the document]
SaxyPandaBear: I'm okay with leaving it. I don't like `init` but no point in spending more time on it

test/ssmParameter.go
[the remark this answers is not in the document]
SaxyPandaBear: Yes I hate it :) IMHO we don't need to pass a pointer for a primitive type as a parameter. That introduces null pointer errors when we don't need to. We can always grab the pointer with `&` as needed.

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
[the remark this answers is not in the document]
SaxyPandaBear: Huh. Weird. Good to know

test/metric/container_insights.go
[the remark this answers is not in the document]
SaxyPandaBear: We could make it accessible from the base fetcher and just have it return an empty array, then all of the implementing child structs can override it.

test/ecs.go
ChenaLee: 1. We can't know exact ec2 instances running without calling this describecontainerinstances, if using autoscaling (which this test is)
2. metadata from ecs thingy seems to be callable using environment variable within each container/task, which aren't accessible in github's host?

test/metric/container_insights.go
ChenaLee: I don't understand. Namespace should be unique? Unique in.. what? Also this entire fetch logic is changed here https://github.com/aws/amazon-cloudwatch-agent-test/pull/41 so let's talk over there

test/metric/metric_value_query.go
ChenaLee: fetch logic changed https://github.com/aws/amazon-cloudwatch-agent-test/pull/26

test/metric/metric_value_query.go
ChenaLee: I don't think so? For plain ec2 tests we are creating an ec2 instance, ssh into it, and run both test and agent on that created ec2 instance. For ecs tests ec2 instances exist but we aren't ssh'ing into it.

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
ChenaLee: Yep this was intentional to split PRs. 1) needed to make it work for ecs to even know what I'm going to do  & 2) it was probably going to be extra line changes in this PR and take longer to make sure nothing breaks. I plan to merge this with a shared base runner with injectable structs

test/ssmParameter.go
ChenaLee: You want me to pull existing stuff out?

test/ssmParameter.go
ChenaLee: Updated!

test/metric_value_benchmark/base_test.go
ChenaLee: which one? why? This is currently used for ec2 tests. Note: my other comment about later plan to merge & why I haven't merged the two

terraform/ecs/linux/ec2_launch/daemon/main.tf
ChenaLee: what's the benefit? also not extensible if there's a test case with multiple ec2 instances behind the same ecs cluster.

test/metric/metric_value_query.go
ChenaLee: fixed!

test/metric/container_insights.go
SaxyPandaBear: > Therefore, should we strive for a function call

What does this mean? The test calls this and passes in `ECS/ContainerInsights`. That is sufficient.

khanhntd: In Container Insights, there always a default namespace (e.g [EKS Container Insights](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/main/plugins/processors/k8sdecorator/structuredlogsadapter/metricruletagger.go#L13) and [ECS Container Insights](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/main/plugins/processors/ecsdecorator/ecsdecorator.go#L16-L18). Moreover, we have a function to detect if we run in [EC2, ECS, EKS](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/main/translator/translate/logs/metrics_collected/prometheus/emfprocessor/ruleMetricNamespace.go#L23-L39) so want to ask if we could use/improve this general function to detect the namespace beforehand.

khanhntd (approved):
LGTM.

environment/MetaData.go
```diff
@@ -0,0 +1,96 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
SaxyPandaBear, line 1: I think that this comment got missed somewhere. The file name shouldn't be capitalized.

khanhntd (approved):

terraform/ecs/linux/ec2_launch/daemon/main.tf
```diff
@@ -0,0 +1,228 @@
...
+  provisioner "local-exec" {
+    command = <<-EOT
+      echo "Validating metrics/logs"
+      cd ../../../../..
+      go test ${var.test_dir} -timeout 0 -computeType=ECS -ecsLaunchType=EC2 -ecsDeploymentStrategy=DAEMON -cwagentConfigSsmParamName=${local.cwagent_config_ssm_param_name} -clusterArn=${aws_ecs_cluster.cluster.arn} -cwagentECSServiceName=${aws_ecs_service.cwagent_service.name} -v --tags=integration
```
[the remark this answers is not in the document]
ChenaLee: yes, that's what it's doing. feel free to change :P

PR #29, opened 2022-11-18
Add Run As User Test

sethAmazon:
# Description of the issue
Need to confirm run as user works for private agent repo.

# Description of changes
Add run as user test

# License

# Tests
go test ./test/run_as_user/ -p 1 -timeout 30m -v --tags=integration

### DIFF
--- a/test/agent_util_linux.go
+++ b/test/agent_util_linux.go
@@ -14,6 +14,11 @@ import (
 	"time"
 )
 
+const (
+	CatCommand      = "cat "
+	AppOwnerCommand = "ps -u -p "
+)
+
 func CopyFile(pathIn string, pathOut string) {
 	log.Printf("Copy File %s to %s", pathIn, pathOut)
 	pathInAbs, err := filepath.Abs(pathIn)
@@ -107,12 +112,13 @@ func RunShellScript(path string, args ...string) error {
 	return nil
 }
 
-func RunCommand(cmd string) {
+func RunCommand(cmd string) (output string) {
 	out, err := exec.Command("bash", "-c", cmd).Output()
 
 	if err != nil {
 		log.Fatalf("Error occurred when executing %s: %s | %s", cmd, err.Error(), string(out))
 	}
+	return string(out)
 }
 
 func ReplaceLocalStackHostName(pathIn string) {
--- a/test/run_as_user/resources/root.json
+++ b/test/run_as_user/resources/root.json
@@ -1,0 +1,18 @@
+{
+  "agent": {
+    "run_as_user": "root",
+    "logfile": ""
+  },
+  "metrics": {
+    "metrics_collected": {
+      "disk": {
+        "measurement": [
+          "used_percent"
+        ],
+        "resources": [
+          "*"
+        ]
+      }
+    }
+  }
+}
--- a/test/run_as_user/run_as_user_test.go
+++ b/test/run_as_user/run_as_user_test.go
@@ -1,0 +1,64 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build linux && integration
+// +build linux,integration
+
+package run_as_user
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/test"
+)
+
+const (
+	// Let the agent run for 15 seconds. This will give agent enough time to change user
+	agentRuntime     = 15 * time.Second
+	configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	pidFile          = "/opt/aws/amazon-cloudwatch-agent/var/amazon-cloudwatch-agent.pid"
+	root             = "root"
+	cwagent          = "cwagent"
+)
+
+type input struct {
+	user      string
+	dataInput string
+}
+
+func TestBundle(t *testing.T) {
+
+	parameters := []input{
+		{dataInput: "resources/default.json", user: root},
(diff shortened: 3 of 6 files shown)


sky333999 (approved):

test/run_as_user/run_as_user_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	for _, parameter := range parameters {
+		t.Run(fmt.Sprintf("resource file location %s user %s", parameter.dataInput, parameter.user), func(t *testing.T) {
+			test.CopyFile(parameter.dataInput, configOutputPath)
+			test.StartAgent(configOutputPath, true)
+			time.Sleep(agentRuntime)
+			log.Printf("Agent has been running for : %s", agentRuntime.String())
+			// Must read the pid file while agent is running
+			pidOutput := test.RunCommand(test.CatCommand + pidFile)
```
khanhntd, line 48: Personally, `runCommand` is generic for me enough to run a command. Therefore, I don't think we need a generic variable for it too. So a command would be enough `cat /opt/aws/amazon-cloudwatch-agent/var/amazon-cloudwatch-agent.pid` for a passing input here

test/run_as_user/run_as_user_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+type input struct {
+	user      string
+	dataInput string
+}
+
+func TestBundle(t *testing.T) {
+
+	parameters := []input{
+		{dataInput: "resources/default.json", user: root},
```
khanhntd, line 36: Are there any clashing issues we need to aware about? Same as `ca_bundle`. If yes, are there any ways we could prevent it or improve it further.

test/run_as_user/run_as_user_test.go
```diff
@@ -0,0 +1,64 @@
...
+		t.Run(fmt.Sprintf("resource file location %s user %s", parameter.dataInput, parameter.user), func(t *testing.T) {
+			test.CopyFile(parameter.dataInput, configOutputPath)
+			test.StartAgent(configOutputPath, true)
+			time.Sleep(agentRuntime)
+			log.Printf("Agent has been running for : %s", agentRuntime.String())
+			// Must read the pid file while agent is running
+			pidOutput := test.RunCommand(test.CatCommand + pidFile)
+			agentOwnerOutput := test.RunCommand(test.AppOwnerCommand + pidOutput)
+			processOwner := outputContainsTarget(agentOwnerOutput, parameter.user)
+			test.StopAgent()
```
khanhntd, line 51: IMO, it would be better to confirm if the agent really stops (same as we do starting) - e.g by checking status

test/run_as_user/resources/root.json
```diff
@@ -0,0 +1,18 @@
...
+    "run_as_user": "root",
+    "logfile": ""
+  },
+  "metrics": {
+    "metrics_collected": {
+      "disk": {
+        "measurement": [
+          "used_percent"
+        ],
+        "resources": [
```
khanhntd, line 12: I will avoid using `resources=*` with disk. Better to reduces the amount of metrics we send to CW (e.g Mem Metrics, etc) to validate the test

test/run_as_user/run_as_user_test.go
```diff
@@ -0,0 +1,64 @@
...
+	user      string
+	dataInput string
+}
+
+func TestBundle(t *testing.T) {
+
+	parameters := []input{
+		{dataInput: "resources/default.json", user: root},
+		{dataInput: "resources/root.json", user: root},
+		{dataInput: "resources/cwagent.json", user: cwagent},
```
khanhntd, line 38: IMO, it would be better if we can validate with `unit test` and reduces the amount of `integration test`? (e.g if we has unit test that validates `root` and `cwagent` that is being set in the toml config or json config, does it matter if we confirm root as default and `root` anymore?

sethAmazon: There are already unit tests for config.

test/run_as_user/run_as_user_test.go
sethAmazon: No since we do run tests one by one.

test/run_as_user/run_as_user_test.go
sethAmazon: I like having cat as a const since it should be reusable.

test/run_as_user/run_as_user_test.go
khanhntd: Yeah so that's why should we test for `root` anymore?

sethAmazon: I'm not sure I understand the question. Are you saying we should not test the default? Are you saying we should not test setting the user to root?

khanhntd: Yes. Test as "user" cwagent is enough for `integration test` since I dont see any value in testing 'root' when user cwagent runs successfuly

sethAmazon: I spoke with the team and they think it is better to leave in the root tests.

test/run_as_user/resources/root.json
sethAmazon: We only run for 15 seconds instead of the flush time of 60 seconds thus cw should not get any metrics. We are only testing process is the correct user.

test/agent_util_linux.go
```diff
@@ -107,12 +112,13 @@ func RunShellScript(path string, args ...string) error {
 	return nil
 }
 
-func RunCommand(cmd string) {
+func RunCommand(cmd string) (ouput string) {
```
adam-mateen, line 115: I recommend just:
```func RunCommand(cmd string) string {```
Since you do not use the variable `ouput` within the function.

jefchien (approved):

PR #45, opened 2022-12-03
Add E2E Sanity Test for Statsd

khanhntd:
# Description of the issue
Add E2E Sanity Test for Statsd 

# Description of changes
* Added integration test to validate that the agent publishes metrics for the Statsd input plugin

# License

# License

# Tests
Run integration  test on my fork https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/3609860857/jobs/6083399608
```
null_resource.integration_test (remote-exec): 2022/12/03 20:01:06 ==============================
null_resource.integration_test (remote-exec): 2022/12/03 20:01:06 ==============Statsd==============
null_resource.integration_test (remote-exec): 2022/12/03 20:01:06 ==============Successful==============
null_resource.integration_test (remote-exec): statsd_counter   Successful
```


sky333999:
LGTM; Some comments from collected PR apply here as well.

test/metric_value_benchmark/statsd_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return testResult
+	}
+
+	if !isAllValuesGreaterThanOrEqualToZero(metricName, values) {
+		return testResult
+	}
+
+	// TODO: Range test with >0 and <100
+	// TODO: Range test: which metric to get? api reference check. should I get average or test every single datapoint for 10 minutes? (and if 90%> of them are in range, we are good)
```
SaxyPandaBear, line 88: How will this statsd test change when we add tests that validate metric values? Should we refactor now or later?

SaxyPandaBear (dismissed):

sky333999 (approved):

test/metric_value_benchmark/statsd_test.go
khanhntd: The good thing about service input (e.g statsd, collectd, emf) is that we can control the value and its deterministic. While the other regular inputs does not. As you know, the statsd and collectd can be aggregated into SEH distribution, (for the old agent) but for the [new agent](https://github.com/khanhntd/private-amazon-cloudwatch-agent-staging/blob/main/plugins/outputs/cloudwatch/aggregator.go#L36-L43) it not has been adopted into place yet. Therefore, the only thing that we can validate now is static value instead of dynamic value (e.g sending histogram with statsd, etc). So to answer your question, we can validate static value and do it now but if we want to test wholefully, we will need to wait for the QS to add the aggregator

SaxyPandaBear (approved):

sky333999 (approved):

PR #64, opened 2022-12-20
Add E2E Sanity Test for Prometheus

sky333999:
# Description of changes
Add E2E Sanity Test for Prometheus

# License

# Tests
* https://github.com/sky333999/amazon-cloudwatch-agent/actions/runs/3741202922

### DIFF
--- a/test/metric/metric_value_query.go
+++ b/test/metric/metric_value_query.go
@@ -21,6 +21,7 @@ import (
 )
 
 var metricValueFetchers = []MetricValueFetcher{
+	&PrometheusMetricValueFetcher{},
 	&SwapMetricValueFetcher{},
 	&CPUMetricValueFetcher{},
 	&MemMetricValueFetcher{},
@@ -61,7 +62,7 @@ type MetricValueFetcher interface {
 	isApplicable(metricName string) bool
 
 	// getMetricSpecificDimensions returns the dimensions that needs to be scraped by each plugin
-	getMetricSpecificDimensions() []types.Dimension
+	getMetricSpecificDimensions(metricName string) []types.Dimension
 
 	// getPluginSupportedMetric returns the supported metrics for each plugin
 	// https://github.com/aws/amazon-cloudwatch-agent/blob/6451e8b913bcf9892f2cead08e335c913c690e6d/translator/translate/metrics/config/registered_metrics.go
--- a/test/metric/metric_value_query.go
+++ b/test/metric/metric_value_query.go
@@ -93,6 +93,10 @@ func (f *baseMetricValueFetcher) getInstanceIdDimension() types.Dimension {
 	}
 }
 
+func (f *baseMetricValueFetcher) getMetricSpecificDimensions(string) []types.Dimension {
+	return []types.Dimension{}
+}
+
 func (f *baseMetricValueFetcher) fetch(namespace, metricName string, metricSpecificDimensions []types.Dimension, stat Statistics) (MetricValues, error) {
 	dimensions := metricSpecificDimensions
 	metricToFetch := types.Metric{
--- a/test/metric/prometheus.go
+++ b/test/metric/prometheus.go
@@ -1,0 +1,91 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build linux && integration
+// +build linux,integration
+
+package metric
+
+import (
+	"log"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+)
+
+type PrometheusMetricValueFetcher struct {
+	baseMetricValueFetcher
+}
+
+var _ MetricValueFetcher = (*PrometheusMetricValueFetcher)(nil)
+
+func (f *PrometheusMetricValueFetcher) Fetch(namespace, metricName string, stat Statistics) (MetricValues, error) {
+	dimensions := f.getMetricSpecificDimensions(metricName)
+	values, err := f.fetch(namespace, metricName, dimensions, stat)
+	if err != nil {
+		log.Printf("Error while fetching metric value for %v: %v", metricName, err.Error())
+	}
+	return values, err
+}
+
+func (f *PrometheusMetricValueFetcher) isApplicable(metricName string) bool {
+	prometheusSupportedMetric := f.getPluginSupportedMetric()
+	_, exists := prometheusSupportedMetric[metricName]
+	return exists
+}
+
+func (f *PrometheusMetricValueFetcher) getPluginSupportedMetric() map[string]struct{} {
+	return map[string]struct{}{
+		"prometheus_test_counter":       {},
+		"prometheus_test_gauge":         {},
+		"prometheus_test_summary_count": {},
+		"prometheus_test_summary_sum":   {},
+		"prometheus_test_summary":       {},
+	}
+}
+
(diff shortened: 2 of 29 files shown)


test/metric_value_benchmark/swap_test.go
```diff
@@ -52,6 +52,10 @@ func (t *SwapTestRunner) getMeasuredMetrics() []string {
 	}
 }
 
+func (t *SwapTestRunner) setupBeforeAgentRun() error {
```
williazz, line 55: good catch

test/metric_value_benchmark/prometheus_test.go
```diff
@@ -0,0 +1,128 @@
...
+func (t *PrometheusTestRunner) getTestName() string {
+	return "Prometheus"
+}
+
+func (t *PrometheusTestRunner) getAgentConfigFileName() string {
+	return "prometheus_config.json"
+}
+
+func (t *PrometheusTestRunner) getAgentRunDuration() time.Duration {
+	return time.Minute
```
williazz, line 74: Question: how does this accurately reflect the run duration?

test/metric_value_benchmark/prometheus_test.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Printf("%v", err)
+		return testResult
+	}
+
+	if !isAllValuesGreaterThanOrEqualToZero(metricName, values) {
+		return testResult
+	}
+
+	// TODO: Range test with >0 and <100
+	// TODO: Range test: which metric to get? api reference check. should I get average or test every single datapoint for 10 minutes? (and if 90%> of them are in range, we are good)
```
williazz, line 124: Question: have you considered just using the max?

williazz (approved):
It's really clean and I like how you tested these changes on your fork

test/metric/metric_value_query.go
```diff
@@ -61,7 +62,7 @@ type MetricValueFetcher interface {
 	isApplicable(metricName string) bool
 
 	// getMetricSpecificDimensions returns the dimensions that needs to be scraped by each plugin
-	getMetricSpecificDimensions() []types.Dimension
+	getMetricSpecificDimensions(metricName string) []types.Dimension
```
taohungyang, line 65: I was a bit confused at the beginning when seeing most of the tests we don't even pass in a string to `getMetricSpecificDimensions`, but I see the reason from the prometheus test. I guess another way might be just implement a default `getMetricSpecificDimensions` in prometheus test without calling it and create another helper function to return the dimensions based on the metric name. Not blocking this though

taohungyang (approved):

test/metric_value_benchmark/prometheus_test.go
sky333999: Sorry, didnt get the question. This is to determine how long to run the agent before stopping it to allow a few scrapes. I just kept this the same as others.

test/metric_value_benchmark/prometheus_test.go
sky333999: This is just a placeholder TODO from other tests. I think we still need to understand how to derive these benchmarks more broadly.

test/metric/metric_value_query.go
sky333999: I see how that can be confusing. But thought more plugins could have this usecase in the future so thought its better to update the interface method itself.

test/metric_value_benchmark/net_test.go
```diff
@@ -44,6 +44,10 @@ func (m *NetTestRunner) getAgentRunDuration() time.Duration {
 	return minimumAgentRuntime
 }
 
+func (m *NetTestRunner) setupBeforeAgentRun() error {
```
jefchien, line 47: nit: I think you can have the function return nil in `BaseTestRunner`, so you don't need to implement it for each of these runners. 

Super quick example: https://go.dev/play/p/EjvuJWSpJ_n

jefchien (dismissed):

test/metric_value_benchmark/prometheus_test.go
williazz: Oh I just meant does this give you the timestamp, not the duration?

williazz (approved):
I like the improved code reuse on these latest changes

test/metric_value_benchmark/prometheus_test.go
sky333999: Gotcha.. yeah no timestamp.. this just says how long the agent should run for before being stopped

test/metric_value_benchmark/prometheus_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (t *PrometheusTestRunner) validatePrometheusMetric(metricName string) status.TestResult {
+	testResult := status.TestResult{
+		Name:   metricName,
+		Status: status.FAILED,
+	}
+
+	fetcher, err := t.MetricFetcherFactory.GetMetricFetcher(metricName)
+	if err != nil {
+		log.Printf("%v", err)
```
khanhntd, line 105: **NIt**: Duplicate [log errors ](https://github.com/aws/amazon-cloudwatch-agent-test/blob/939d1e01e55916dce7ae954721ad4cf018647ba9/test/metric/metric_value_query.go#L41-L51)

test/metric_value_benchmark/prometheus_test.go
```diff
@@ -0,0 +1,124 @@
...
+		return testResult
+	}
+
+	values, err := fetcher.Fetch(namespace, metricName, metric.AVERAGE)
+	if err != nil {
+		log.Printf("%v", err)
```
khanhntd, line 111: **Nit**: Same log errors

test/metric_value_benchmark/base_test.go
```diff
@@ -41,6 +42,14 @@ type BaseTestRunner struct {
 	*metric.MetricFetcherFactory
 }
 
+func (t *BaseTestRunner) setupBeforeAgentRun() error {
+	return nil
+}
+
+func (t *BaseTestRunner) setupAfterAgentRun() error {
```
khanhntd, line 49: **Nit**: Would expect the same with [getMetricSpecificDImension](https://github.com/aws/amazon-cloudwatch-agent-test/blob/939d1e01e55916dce7ae954721ad4cf018647ba9/test/metric/mem.go#L53-L55) since some are having no dimensions

test/metric/prometheus.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (f *PrometheusMetricValueFetcher) isApplicable(metricName string) bool {
+	prometheusSupportedMetric := f.getPluginSupportedMetric()
+	_, exists := prometheusSupportedMetric[metricName]
+	return exists
+}
+
+func (f *PrometheusMetricValueFetcher) getPluginSupportedMetric() map[string]struct{} {
+	return map[string]struct{}{
+		"prometheus_test_counter":       {},
```
khanhntd, line 39: So we would not do histogram or untyped?

sky333999: We need to enhance the framework to account of metrics we dont expect. Currently we only handle expected metrics.
Histogram & untyped would be ones we dont expect.

khanhntd: Weird. I forgot to update my comments. Yes agree on that. If yes, it would be better to have document on what type of Prometheus metrics type we support though.

sky333999: Ill create a task for tracking the enhancement. By document, did you mean just add a comment here? If yes, I just added a comment.

SaxyPandaBear (approved):

khanhntd (approved):
LGTM.

jefchien (approved):

PR #67, opened 2022-12-25
Add E2E sanity test for disk

williazz:
# Description of the issue
Adds an e2e sanity test for disk plugin. This test mounts an EFS to the EC2 instance to make `fstype` and `path` predictable. 

In a future PR, I will migrate the EFS mounting process from terraform to the AMI building pipeline

# License

# Tests
[Ran on hacked_ec2 branch of my fork](https://github.com/williazz/amazon-cloudwatch-agent/actions/runs/3801322858/jobs/6466811131)

I also tested the `install-efs-utils` script on Amazon Linux, Ubunutu, Debian, and SLES


terraform/ec2/linux/main.tf
```diff
@@ -26,6 +26,45 @@ locals {
   private_key_content = var.ssh_key_name != "" ? var.ssh_key_value : tls_private_key.ssh_key[0].private_key_pem
 }
 
+#####################################################################
+# Create EFS
+#####################################################################
+resource "aws_efs_file_system" "efs" {
+  creation_token = "predictable-${random_id.testing_id.hex}"
+  tags           = {
+    Name = "predictable-${random_id.testing_id.hex}"
```
SaxyPandaBear, line 35: nit - this is not a great name. What is predictable? We're tagging the resource by naming it here.

terraform/ec2/linux/main.tf
```diff
@@ -26,6 +26,45 @@ locals {
...
+
+  provisioner "remote-exec" {
+    # https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-helper-ec2-linux.html
+    inline = [
+      "sudo yum install -y amazon-efs-utils",
+      "sudo mkdir ${var.efs_mount_point}",
+      "sudo mount -t efs -o tls ${aws_efs_file_system.efs.dns_name} ${var.efs_mount_point}/",
+    ]
+  }
+}
```
SaxyPandaBear, line 66: This will not work for a distro of linux that does not have yum, e.g.: Ubuntu.

test/metric/disk.go
```diff
@@ -0,0 +1,62 @@
...
+		"disk_used":         {},
+		"disk_used_percent": {},
+	}
+}
+
+func (f *DiskMetricValueFetcher) getMetricSpecificDimensions(string) []types.Dimension {
+	return []types.Dimension{
+		{
+			Name:  aws.String("path"),
+			Value: aws.String("/home/ec2-user/efs-mount-point"),
```
SaxyPandaBear, line 55: Eventually would be a problem for windows testing but should be fine on linux

terraform/ec2/linux/main.tf
williazz: Good catch. I'm trying to think of a clean and agnostic way to install `amazon-efs-utils`

williazz: I added a script to use yum or apt-get to install amazon-efs-utils, still considering other edge cases for linux

test/metric_value_benchmark/agent_configs/disk_config.json
```diff
@@ -0,0 +1,31 @@
...
+          "*"
+        ],
+        "measurement": [
+          "free",
+          "inodes_free",
+          "inodes_total",
+          "inodes_used",
+          "total",
+          "used",
+          "used_percent"
```
adam-mateen, line 25: Instead of just testing a single JSON config with ALL metrics enabled, it would be better to iteratively test combinations of metrics.
1. only configure metric A, and verify only A is reported with recent timestamp.
2. only configure metric B, and verify only B is reported with recent timestamp.
3. .... repeat for metrics A-Z.
4. Then configure metric A and B, and verify only those 2 get reported.
5. I realize the number of combinations will be too large and make test time a pain, so please think of ways to reduce the combinations... Maybe iterate from 0 to total_num_metrics, and than randomly select current_count number of metrics.

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -64,6 +64,7 @@ func getEc2TestRunners(env *environment.MetaData) []*TestRunner {
 	if ec2TestRunners == nil {
 		factory := &metric.MetricFetcherFactory{Env: env}
 		ec2TestRunners = []*TestRunner{
+			{testRunner: &DiskTestRunner{BaseTestRunner{MetricFetcherFactory: factory}}},
```
adam-mateen, line 67: Using the word "benchmark" to describe these tests is confusing to me. These tests do not compare the metric value against anything.

>verb
evaluate or check (something) by comparison with a standard.
"we are benchmarking our performance against external criteria"

adam-mateen (dismissed):

test/metric_value_benchmark/metrics_value_benchmark_test.go
williazz: @adam-mateen agreed

williazz: added

test/metric_value_benchmark/agent_configs/disk_config.json
williazz: @adam-mateen that's an interesting point. I'll convert that into a github issue so that we can contain our discussion over there

williazz: Issue #71

terraform/ec2/linux/install-efs-utils.sh
```diff
@@ -0,0 +1,32 @@
+# install amazon EFS client on linux, agnostic to yum or apt
```
SaxyPandaBear, line 1: We already talked about this separately, but we should update our AMIs that we build for testing to install the EFS utils instead of doing it in our terraform for test setup

test/metric/disk.go
```diff
@@ -0,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
SaxyPandaBear, line 1: I don't think we need this file anymore

terraform/ec2/linux/main.tf
williazz: Added support for `zypper`. Among the three, all linux distros should now be supported

terraform/ec2/linux/install-efs-utils.sh
```diff
@@ -0,0 +1,32 @@
...
+  sudo apt-get update -y
+  sudo apt-get -y install git binutils
+  git clone https://github.com/aws/efs-utils
+  cd ./efs-utils
+  ./build-deb.sh
+  sudo apt-get -y install ./build/amazon-efs-utils*deb
+elif command -v yum >/dev/null; then
+  sudo yum update -y
+  sudo yum install sudo yum install -y amazon-efs-utils
+elif command -v zypper >/dev/null; then
```
sethAmazon, line 14: What os uses zypper? Are we going to add tests for this os?

terraform/ec2/linux/main.tf
```diff
@@ -26,6 +26,51 @@ locals {
...
+
+resource "aws_efs_mount_target" "mount" {
+  file_system_id = aws_efs_file_system.efs.id
+  subnet_id = aws_instance.cwagent.subnet_id
+  security_groups = [aws_security_group.ec2_security_group.id]
+}
+
+resource "null_resource" "mount_efs" {
```
sethAmazon, line 45: Do we need the null resource? Should we add the depends on what the null resource depends on instead of creating a null resource.

test/metric_value_benchmark/agent_configs/disk_config.json
khanhntd: Interesting point. However, why all combinations though? All combinations are the subset of testing all metrics? Unless there are higher gate to deal with that plugins (e.g testing small subset of metrics in first part and testing others metrics in performance metrics)

test/metric_value_benchmark/disk_test.go
```diff
@@ -0,0 +1,80 @@
...
+		Name:   metricName,
+		Status: status.FAILED,
+	}
+
+	fetcher, err := t.MetricFetcherFactory.GetMetricFetcher(metricName)
+	if err != nil {
+		return testResult
+	}
+
+	values, err := fetcher.Fetch(namespace, metricName, metric.AVERAGE)
```
sethAmazon, line 68: My understanding from reading the code is we get average metric stat. If the stat is there we pass the test?  Is there a way to control time? I think we will be finding values from past runs instead of only this run.

terraform/ec2/linux/main.tf
```diff
@@ -26,6 +26,51 @@ locals {
...
+    source = "install-efs-utils.sh"
+    destination = "/tmp/install-efs-utils.sh"
+  }
+
+  provisioner "remote-exec" {
+    # https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-helper-ec2-linux.html
+    inline = [
+      "chmod +x /tmp/install-efs-utils.sh",
+      "/tmp/install-efs-utils.sh",
+      "sudo mkdir ${var.efs_mount_point}",
```
khanhntd, line 68: Will the `efs_mount_point` change according to OS? If not, then IMO we don't need the variable though since its a static mount or partition.

terraform/ec2/linux/install-efs-utils.sh
```diff
@@ -0,0 +1,32 @@
+# install amazon EFS client on linux, agnostic to yum, apt, or zypper + rpm
```
khanhntd, line 1: **Nit pick**: Amazon License though

terraform/ec2/linux/install-efs-utils.sh
khanhntd: Agree. If adding Zypper, then it would be better to have OpenSUSE or SLES OS too. If not, we are over-engineering this though.

terraform/ec2/linux/install-efs-utils.sh
```diff
@@ -0,0 +1,32 @@
...
+# https://docs.aws.amazon.com/efs/latest/ug/installing-amazon-efs-utils.html
+
```
khanhntd, line 3: Overall, the step for Suse and Debian/Ubuntu is to clone and install the package, while Amazon Linux does have a similar step with Suse without using zypper. If that's the case, should we apply the general installation for all OS and check if zypper is there, we would use zypper to refresh (to avoid OpenSuse case)?

terraform/ec2/linux/install-efs-utils.sh
```diff
@@ -0,0 +1,32 @@
...
+  sudo zypper install -y git rpm-build make
+  git clone https://github.com/aws/efs-utils
+  cd ./efs-utils
+  make rpm
+  sudo zypper --no-gpg-checks install -y build/amazon-efs-utils*rpm
+else
+  echo "Unable to install amazon-efs-client"
+  exit 1
+fi
```
khanhntd, line 32: Maybe moving this file to a share resources folder?

terraform/ec2/linux/main.tf
```diff
@@ -26,6 +26,51 @@ locals {
...
+  ]
+
+  connection {
+    type = "ssh"
+    user = "ec2-user"
+    private_key = local.private_key_content
+    host = aws_instance.cwagent.public_ip
+  }
+
+  provisioner "file" {
```
khanhntd, line 58: I wonder should we setup all the resources (including cloning) instead of [letting at null resource integration test](https://github.com/williazz/amazon-cloudwatch-agent-test/blob/main/terraform/ec2/linux/main.tf#L32-L43)? Since it would reduce the necessary to use "file" with scp, want to ask your opinion on the design script though.

terraform/ec2/linux/install-efs-utils.sh
williazz: We suport SLES according to [public cwa docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Install-CloudWatch-Agent.html), and `amazon-efs-utils` insallation uses zypper on SLES and OpenSUS according to this [doc](https://docs.aws.amazon.com/efs/latest/ug/installing-amazon-efs-utils.html)

khanhntd: I think Seth meant to adding a OS use this to confirm this behavior ! In your case, its OpenSus and SLES (if we are not confirming it and breaking at some points, its a tech debt though)

terraform/ec2/linux/install-efs-utils.sh
williazz: Could you clarify what you mean?

terraform/ec2/linux/main.tf
williazz: No, it will not change. I only made it a variable for code reuse

terraform/ec2/linux/install-efs-utils.sh
khanhntd: Adding Amazon License when creating new files though
```
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT
```

terraform/ec2/linux/main.tf
khanhntd: Then it safe to delete it though IMO. Make it static and let's people less worry about it would be better experience.

terraform/ec2/linux/main.tf
williazz: > Should we add the depends on what the null resource depends on instead of creating a null resource.

Could you break down what this means?

test/metric_value_benchmark/agent_configs/disk_config.json
williazz: please discuss in #71 since this is out of scope

terraform/ec2/linux/main.tf
williazz: Not sure what you mean by "setup all the resources" instead of using the null resource.

If this answers your question, I plan to move the efs mounting process from terraform to the AMI building pipelines next week, which should dramatically reduce the over-engineering in this PR

terraform/ec2/linux/main.tf
williazz: To help my learning, is that a design or stylistic principle?

terraform/ec2/linux/install-efs-utils.sh
williazz: Good idea thank you

terraform/ec2/linux/main.tf
khanhntd: It's design though. I will leave the final decision to others though.

terraform/ec2/linux/install-efs-utils.sh
khanhntd: For every script based on my knowledge.

SaxyPandaBear (approved):

terraform/ec2/linux/main.tf
khanhntd: Then LGTM.

terraform/ec2/linux/install-efs-utils.sh
williazz: Since public docs state that cwa suports SLES, then we will definitely add tests for SLES eventually. 

As for tech debt, I agree that this is script is a tech debt. I plan to migrate the efs mounting process from terraform to the AMI building stage, which should be a much cleaner solution for everyone.

khanhntd (approved):

PR #68, opened 2022-12-28
No append test & refactor for dimension tests

ChenaLee:
# Description of the issue
Missing tests for behavior when no append dimensions were specified in config

# Description of changes
- added no append dimension test
- refactor for various dimension tests flexibility

# License

# Tests
Integration test. Note: There seems to be 30 minute max timeout no matter the go test flag I provide. Need to investigate that soon.


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/aws/amazon-cloudwatch-agent-test
 go 1.18
 
 require (
+	github.com/aws/aws-sdk-go v1.44.159
```
SaxyPandaBear, line 6: I don't think we need to pull in v1 of the SDK

terraform/ec2/linux/main.tf
```diff
@@ -49,9 +49,8 @@ resource "null_resource" "integration_test" {
       "echo sha ${var.cwa_github_sha}",
       "cloud-init status --wait",
       "echo clone and install agent",
-      "git clone ${var.github_test_repo}",
+      "git clone --branch ${var.github_test_repo_branch} ${var.github_test_repo}",
```
SaxyPandaBear, line 52: Thank you. I need this 😂

test/metric/dimension/default/container_insights.go
```diff
@@ -4,33 +4,23 @@
 //go:build linux && integration
 // +build linux,integration
 
-package metric
+package _default
```
SaxyPandaBear, line 7: red flag. what is the purpose of the package change?

test/metric_value_benchmark/mem_test.go
```diff
@@ -10,41 +10,49 @@ import (
...
+	"github.com/aws/amazon-cloudwatch-agent-test/test/metric/dimension"
+
 	"github.com/aws/amazon-cloudwatch-agent-test/test/status"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/test_runner"
 )
 
 type MemTestRunner struct {
-	BaseTestRunner
+	test_runner.BaseTestRunner
+	Base test_runner.BaseTestRunner
```
SaxyPandaBear, line 21: Why does the struct need two copies of the BaseTestRunner?

test/metric/dimension/provider.go
```diff
@@ -0,0 +1,94 @@
...
+
+type ExpectedDimensionValue struct {
+	Value *string
+}
+
+func (d *ExpectedDimensionValue) IsKnown() bool {
+	if d.Value == nil {
+		return false
+	}
+	return true
```
SaxyPandaBear, line 24: Is this equivalent to `return d.Value != nil`

test/metric/dimension/instanceid_provider.go
```diff
@@ -0,0 +1,70 @@
...
+	Provider
+}
+
+var _ IProvider = (*ECSInstanceIdDimensionProvider)(nil)
+
+func (p *ECSInstanceIdDimensionProvider) IsApplicable() bool {
+	if (p.env.ComputeType == computetype.ECS) {
+		return true
+	}
+	return false
```
SaxyPandaBear, line 27: equivalent to `return p.env.ComputeType == computetype.ECS`

test/metric/dimension/provider.go
ChenaLee: oh lol yeah

test/metric_value_benchmark/mem_test.go
ChenaLee: apparently that's not how it works. if I wanna reference member variables from Base then I need to specify Base, but if I wanna just use the methods defined for base then I cannot specify Base. I guess in that case I will have to do Base.call()

test/metric/dimension/instanceid_provider.go
ChenaLee: yes

ChenaLee: Done

test/metric/dimension/provider.go
ChenaLee: Done

test/metric_value_benchmark/mem_test.go
SaxyPandaBear: 🙁 well, in that case, I would rather drop the first one and then update usage everywhere to do `Base.function()`. I don't like that, but it's better than specifying it twice. 

Or, do you need member variables from base or can you expose it via a function?

sky333999: Im still a little confused why we need both here... was it a compilation error or a runtime error? Because I atleast tried `t.DimensionFactory.GetDimensions` instead of `t.Base.DimensionFactory.GetDimensions` and it didnt complain in my IDE.

test/metric_value_benchmark/mem_test.go
```diff
@@ -10,41 +10,49 @@ import (
...
 	return "mem_config.json"
 }
 
-func (m *MemTestRunner) getAgentRunDuration() time.Duration {
-	return minimumAgentRuntime
+func (m *MemTestRunner) GetAgentRunDuration() time.Duration {
+	return test_runner.MinimumAgentRuntime
+}
+
+func (t *MemTestRunner) SetupAfterAgentRun() error {
```
sky333999, line 51: Dont think we need these overrides since returning nil is the default behavior. In fact it might be nice to make the above `GetAgentRunDuration` a default on `BaseTestRunner` as well

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -45,37 +44,40 @@ func init() {
...
+				env:              *env,
+			},
 		}
 	}
 	return ecsTestRunners
 }
 
-func getEc2TestRunners(env *environment.MetaData) []*TestRunner {
+func getEc2TestRunners(env *environment.MetaData) []*test_runner.TestRunner {
+	factory := dimension.GetDimensionFactory(*env)
```
sky333999, line 66: any reason to move this line above the if block?

test/metric_append_dimension/no_append_dimensions_test.go
```diff
@@ -0,0 +1,97 @@
...
+	log.Printf("metric values are %v", values)
+	if err != nil {
+		return testResult
+	}
+
+	if !isAllValuesGreaterThanOrEqualToZero(metricName, values) {
+		return testResult
+	}
+
+	// TODO: Range test with >0 and <100
```
sky333999, line 92: Should we delete these comments since I dont think this test would ever have benchmarks?

test/metric/dimension/container_insights_provider.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type ContainerInsightsDimensionProvider struct {
+	Provider
+}
+
+var _ IProvider = (*ContainerInsightsDimensionProvider)(nil)
+
+func (p *ContainerInsightsDimensionProvider) IsApplicable() bool {
+	if (p.Provider.env.ComputeType == computetype.ECS) {
```
sky333999, line 24: could simplify this to just return the result of `p.Provider.env.ComputeType == computetype.ECS` similar to the one below

test/metric/dimension/custom_provider.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+var _ IProvider = (*CustomDimensionProvider)(nil)
+
+func (p *CustomDimensionProvider) IsApplicable() bool {
+	return true
+}
+
+func (p *CustomDimensionProvider) GetDimension(instruction Instruction) types.Dimension {
+	log.Printf("CustomDimProvider")
```
sky333999, line 26: were these temp prints or did you mean to keep these?

test/metric/dimension/instanceid_provider.go
```diff
@@ -0,0 +1,67 @@
...
+		Name:  aws.String("InstanceId"),
+		Value: aws.String(containerInstances[0].EC2InstanceId),
+	}
+}
+
+type InstanceIdDimensionProvider struct {
+	Provider
+}
+
+var _ IProvider = (*ECSInstanceIdDimensionProvider)(nil)
```
sky333999, line 52: should this be `InstanceIdDimensionProvider`?

test/metric/dimension/instanceid_provider.go
```diff
@@ -0,0 +1,67 @@
...
+		return types.Dimension{}
+	}
+
+	return types.Dimension{
+		Name:  aws.String("InstanceId"),
+		Value: aws.String(containerInstances[0].EC2InstanceId),
+	}
+}
+
+type InstanceIdDimensionProvider struct {
```
sky333999, line 48: Might be better to explicitly call this `EC2InstanceIdDimensionProvider`

test/metric/dimension/instanceid_provider.go
ChenaLee: crab thank you

test/metric/dimension/instanceid_provider.go
ChenaLee: How about local? rn we assume ec2 test cases always run on the same ec2 instance where agent is installed.. but that's an assumption

test/metric/dimension/custom_provider.go
ChenaLee: yeah that's not too useful I guess

test/metric/dimension/container_insights_provider.go
ChenaLee: oh wow andrew missed this!

test/metric_append_dimension/no_append_dimensions_test.go
ChenaLee: you are very good at catching things :(

test/metric_value_benchmark/metrics_value_benchmark_test.go
ChenaLee: There was a reason when I first wrote something in this file but no not anymore

test/metric_value_benchmark/mem_test.go
ChenaLee: yes, I forgot to delete them. this was before I realized I missed that was possible. had build breaks that was fixable without doing this

ChenaLee: For some of these things I feel like there should be a better way - because right now this is kinda technically has problems of inheritance. But not something I wanna worry about rn

test/metric_value_benchmark/prometheus_test.go
```diff
@@ -99,12 +103,58 @@ func (t *PrometheusTestRunner) validatePrometheusMetric(metricName string) statu
 		Status: status.FAILED,
 	}
 
-	fetcher, err := t.MetricFetcherFactory.GetMetricFetcher(metricName)
-	if err != nil {
+	dims, failed := t.Base.DimensionFactory.GetDimensions([]dimension.Instruction{})
```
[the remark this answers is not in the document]
ChenaLee: yeah probably. i was trying to make dims and failed accessible from outside of switch but nils complain. I will see if I can do anything about it


also for the comment below: wow that didn't work when I tried so that's why I did this Base thing. dang.

test/metric_value_benchmark/mem_test.go
ChenaLee: wow that didn't work when I tried so that's why I did this Base thing. dang. It did complain when I did go test but somehow not anymore. I must have misinterpreted something

test/metric/dimension/container_insights_provider.go
SaxyPandaBear: :(((((((((((( fine I'll look harder next time

test/metric/dimension/instanceid_provider.go
sky333999: Yeah local might also be ok. I was thinking when someone looks at this one type in isolation, it doesnt give much context immediately as to where its running etc. so would just be better to be explicit in the name itself.
Non-blocker though.

test/metric/dimension/host_provider.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type HostDimensionProvider struct {
+	Provider
+}
+
+var _ IProvider = (*HostDimensionProvider)(nil)
+
+func (p *HostDimensionProvider) IsApplicable() bool {
+	if (p.env.ComputeType == computetype.EC2) {
```
sky333999, line 23: one more :(

test/metric/dimension/instanceid_provider.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type ECSInstanceIdDimensionProvider struct {
+	Provider
+}
+
+var _ IProvider = (*ECSInstanceIdDimensionProvider)(nil)
+
+func (p *ECSInstanceIdDimensionProvider) IsApplicable() bool {
+	if p.env.ComputeType == computetype.ECS {
```
sky333999, line 24: two more :(

sky333999 (approved):

test/metric/dimension/host_provider.go
ChenaLee: done!

test/metric/dimension/instanceid_provider.go
ChenaLee: done!

sky333999 (approved):

SaxyPandaBear (approved):

PR #70, opened 2022-12-30
Add Terraform setup to setup all the resources before hand and look up those resources when running tests

khanhntd:
# Description of the issue
Instead of setup all the resources during running which can cause some drawbacks:
* Increase time for setup all those resources (e.g in EKS, create EKS cluster would take more than 30 minutes)
* Create resources that terraform cannot destroy which cause unnecessary resources (e.g Security Group)
* Share the common resources between tests (e.g IAM, SG, etc)

# Description of changes
* [Add a common module to share all the constant variables](https://github.com/aws/amazon-cloudwatch-agent-test/pull/70/files)
* Add a [setup module to create all the modules beforehand](https://github.com/aws/amazon-cloudwatch-agent-test/pull/70/files#diff-cff421649d96486b6eaafff40bf87ddd341b5b4cbaf627deb4cba21c364d8a21)
* Add init environment [setup for nvidia ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/70/files#diff-8013d42f0fa4551baa4c2d82140622bfe42be2cb0e03ecdf69ab4536866b2d8d)
* Separate [ECS Fargate and ECS EC2 ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/70/files#diff-56a236e4b6e258f5d0557652161f06c17f2668c5aa9adec128a949321ca5bcfe)

# Note
* Next step is create a PR to adhere the changes  and also run the setup once for the integration test open source account to setup all the required resources
# License

# Tests
By apply this changes, I have run this on my fork
* Performance Tracking Test: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4070853647/jobs/7012314979
* Other tests https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4070754614


terraform/common/variables.tf
```diff
@@ -0,0 +1,18 @@
...
+}
+
+variable "cwagent_image_tag" {
+  type    = string
+  default = "latest"
+}
+
+variable "vpc_security_group" {
+  type    = string
+  default = "vpc_security_group"
```
SaxyPandaBear, line 17: Should this be the default?

terraform/setup/providers.tf
```diff
@@ -0,0 +1,67 @@
...
+  hash_key     = "LockID"
+  billing_mode = "PAY_PER_REQUEST"
+
+  attribute {
+    name = "LockID"
+    type = "S"
+  }
+
+  depends_on = [aws_s3_bucket.setup-remote-state-s3-bucket]
+}
```
SaxyPandaBear, line 67: What is this for

test/nvidia_gpu/metrics_nvidia_gpu_window_test.go
```diff
@@ -24,7 +24,14 @@ const (
 	numberofWindowsAppendDimensions = 1
 )
 
-var expectedNvidiaGPUWindowsMetrics = []string{"Memory % Committed Bytes In Use", "nvidia_smi utilization_gpu", "nvidia_smi utilization_memory", "nvidia_smi power_draw", "nvidia_smi temperature_gpu"}
+var (
+	envMetaDataStrings              = &(environment.MetaDataStrings{})
```
SaxyPandaBear, line 28: This conflicts with the declaration in the linux test doesn't it?

terraform/ec2/README.md
```diff
@@ -275,7 +275,16 @@ Outputs:
...
 
-## 7. Run the integration test action on your fork
+## 7. Setup the integration test resources
+1. Navigate to your fork
+2. Run the following command to setup all the resources being used for integration test (e.g vpc, iam, etc)
+```shell
+cd terraform/setup # assuming you are still in the ./terraform
+terraform init
+terraform apply --auto-approve 
+```
```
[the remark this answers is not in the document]
khanhntd: Its a one-time setup for resources can be share (e.g vpc) and can be used all over again without determine/affect the test's outcome or make the test more deterministics. Therefore, we don't need to destroy the resources and used it over again across the workflow

terraform/ec2/localstack/variables.tf
```diff
@@ -30,15 +33,10 @@ variable "cwa_test_github_sha" {
 
 variable "github_test_repo" {
   type    = string
-  default = ""
+  default = "https://github.com/aws/amazon-cloudwatch-agent-test.git  "
```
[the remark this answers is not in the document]
khanhntd: Nice catch. It does not affect the test outcome but we can quickly fix it.

terraform/setup/providers.tf
khanhntd: Its for a distributed system across dev workspace. The issue [this script is solving ](https://github.com/aws/amazon-cloudwatch-agent-test/blob/b3ea48c5c5891bcfe07762cf9da871eebac9e47f/terraform/setup/providers.tf#L13-L67) is that: since the setup is a one-time script and not a part of the test, multiple developers can run into a conflict: return an error when an existing resources are already there, therefore, two issues appear:
* The dev will have an outdate resources even if they update this script 
* Force them to delete those resources manually.
So create a share terraform state configuration will resolve these two issues.

Coming to the main question, [ the dynamo table is ](https://github.com/aws/amazon-cloudwatch-agent-test/blob/b3ea48c5c5891bcfe07762cf9da871eebac9e47f/terraform/setup/providers.tf#L17) only exists to avoid a race condition when updating the terraform state configuration

terraform/setup/iam.tf
```diff
@@ -0,0 +1,81 @@
...
+      "ecs:ListServices",
+      "ecs:DescribeTaskDefinition",
+      "ecr:GetAuthorizationToken",
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
+      "ssm:Describe*",
+      "ssm:Get*",
+      "ssm:List*",
+      "s3:*",
```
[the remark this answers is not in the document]
khanhntd: For S3 I am not sure. However, questions for everyone, would it be better to use * for all resources in AWS or would it be better to list the specific actions in the resources? I would prefer the former one since it would make the dev less awareness regarding the actions being used but want your thought on that before implementing it.

test/nvidia_gpu/metrics_nvidia_gpu_window_test.go
khanhntd: Yes you are right. Let's me confirm that real quick.

sethAmazon:
Is there a change needed to the integration test yaml

khanhntd:
> Is there a change needed to the integration test yaml


Yes I have stated this in the PR: `Next step is create a PR to adhere the changes and also run the setup once for the integration test open source account to setup all the required resources`

terraform/ecs_ec2/daemon/vpc.tf
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
khanhntd, line 1: This is only separate from Chena ECS EC2 Test.

terraform/ec2/README.md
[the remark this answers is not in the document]
khanhntd: The `common` folder is a shared common constant between [these terraform](https://github.com/aws/amazon-cloudwatch-agent-test/blob/6f8c6538f088d6d3b7b8e0b0fcbe83f7f195a40f/terraform/common/output.tf). Example would be if you create an EC2 Instance Profile with name `cwa-e2e-iam-instance-profile`; therefore, it will look up them with the same name (e.g [EC2 Terraform looking up](https://github.com/aws/amazon-cloudwatch-agent-test/blob/6f8c6538f088d6d3b7b8e0b0fcbe83f7f195a40f/terraform/ec2/linux/iam.tf#L4-L6))

sethAmazon (approved):

PR #85, opened 2023-01-17
Fix Compile For No Append Dimensions Test And Allow Running On Darwin

sethAmazon:
# Description of the issue
```
null_resource.integration_test (remote-exec): test/metric_append_dimension/metrics_append_dimension_test.go:49:46: unknown field 'Base' in struct literal of type NoAppendDimensionTestRunner
```

# Description of changes
Fix compile issue by using BaseTestRunner instead of Base and allowing Darwin compiling. 

# License

# Tests
Do not see a compile issue in intelij.


sethAmazon:
We do not need the integration tag since this is not in the base repo. The reason for the integration tag was, so we did not run an integration test, a unit test, when it was in the main repo. Now we can remove it, so this package will compile in IntelliJ without setting the build tag.

internal/common/agent_util_unix.go
```diff
@@ -1,8 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT
 
-//go:build linux && integration
-// +build linux,integration
+//go:build unix
```
khanhntd, line 4: Should we do **!windows** instead?

khanhntd (commented):
Would be better to have the successful run here to confirm if all the changes are good to delivered. Other than that, LGTM.

internal/common/agent_util_unix.go
sethAmazon: I think either is fine.

sethAmazon:
I was only looking for this to be in a compilable state. Did not run it. Will run it after merge.

SaxyPandaBear (dismissed):
If we merge this and don't update the GHA workflow in the main repo, it won't run any tests, right? Is there another PR for the fix there?

sethAmazon:
Both repos automatically take the latest version of the test repo.

SaxyPandaBear:
I keep forgetting the test execution got moved here

khanhntd:
If you are removing the tag, please also update the corresponding actions in the terraform script to remove the tag (e.g sanity or [integration test](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/terraform/ec2/linux/main.tf#L129))

sethAmazon:
Removed tags

khanhntd (approved):
LGTM.

williazz (approved):

PR #93, opened 2023-01-23
Aggregate

ChenaLee:
# Description of the issue
Aggregate dimension sanity test is an alpha requirement. Needed to be done asap.

# Description of changes
Added a test for aggregate dimension.

# License

# Tests
Created one new e2e test and verified it passes.
(Last line says fail but that's just because I made the suite to mark fail no matter what after all the tests run. This is for easy/quick rerun)

Screen Shot 2023-01-23 at 14 57 05


test/metric/dimension/provider.go
```diff
@@ -17,11 +17,11 @@ type ExpectedDimensionValue struct {
 }
 
 func (d *ExpectedDimensionValue) IsKnown() bool {
-	return d.Value == nil
+	return d.Value != nil
```
sethAmazon, line 20: Nit: I think you may need to rebase on main. I think this fix is already in.

test/metric_dimension/metrics_aggregate_dimension_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build linux && integration
```
sethAmazon, line 4: Nit: integration is no longer needed since this is not in the same dir as the agent's unit tests. For we should use !windows since this should work on darwin.

test/metric_dimension/metrics_aggregate_dimension_test.go
```diff
@@ -0,0 +1,74 @@
...
+// +build linux,integration
```
sethAmazon, line 5: Nit: // +build having this is no longer recommened.

test/metric_dimension/metrics_append_dimension_test.go
```diff
@@ -46,7 +46,8 @@ func getTestRunners(env *environment.MetaData) []*test_runner.TestRunner {
 	if testRunners == nil {
 		factory := dimension.GetDimensionFactory(*env)
 		testRunners = []*test_runner.TestRunner{
-			{TestRunner: &NoAppendDimensionTestRunner{Base: test_runner.BaseTestRunner{DimensionFactory: factory}}},
+			{TestRunner: &NoAppendDimensionTestRunner{test_runner.BaseTestRunner{DimensionFactory: factory}}},
```
sethAmazon, line 49: Think the append dimension tests could be added to this test suite?

khanhntd: The append dimension tests would cover a lot though. There are 2 use cases that would need to test:
* The global append dimension
* The custom dimension which would only exist for some plugin
This would be an outside scope so yeah maybe something to consider.

khanhntd (commented):
Would need a rebase before able to continue reviewing though. Thanks for working on this too Chena !!

terraform/ec2/README.md
```diff
@@ -157,8 +157,10 @@ Outputs:
         "elasticfilesystem:CreateMountTarget",
         "elasticfilesystem:DeleteFileSystem",
         "elasticfilesystem:DeleteMountTarget",
+        "elasticfilesystem:DescribeMountTargets",
```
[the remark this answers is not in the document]
ChenaLee: ugh I told Billy about it yesterday and he beat me to it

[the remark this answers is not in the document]
ChenaLee: will have to do conflict resolution anyway

test/metric/dimension/provider.go
ChenaLee: yes will have to rebase. Somebody told me it wasn't fixed yet & my repo was failing even with newest commits when I was doing this :)

test/metric_dimension/agent_configs/one_aggregate_dimension.json
```diff
@@ -0,0 +1,23 @@
...
+    "debug": true,
+    "logfile": ""
+  },
+  "metrics": {
+    "namespace": "MetricAggregateDimensionTest",
+    "append_dimensions": {
+      "InstanceId": "${aws:InstanceId}",
+      "InstanceType": "${aws:InstanceType}"
+    },
+    "aggregation_dimensions": [[]],
```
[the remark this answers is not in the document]
ChenaLee: same dimension? no when you do this you get metrics with no other dimension than metric names. See validation!

test/metric_dimension/metrics_aggregate_dimension_test.go
ChenaLee: yeah but for now everything is assuming tests run with integration flag (terraform, all files) so not something I wanna change in this PR. Am I missing something?

test/metric_dimension/metrics_aggregate_dimension_test.go
ChenaLee: ty! will take a look at reasons

test/metric_dimension/metrics_aggregate_dimension_test.go
```diff
@@ -0,0 +1,74 @@
...
+func (t *OneAggregateDimensionTestRunner) GetTestName() string {
+	return "OneAggregatedDimension"
+}
+
+func (t *OneAggregateDimensionTestRunner) GetAgentConfigFileName() string {
+	return "one_aggregate_dimension.json"
+}
+
+func (t *OneAggregateDimensionTestRunner) GetMeasuredMetrics() []string {
+	return []string{"cpu_time_active", "cpu_time_guest"}
```
sky333999, line 45: I think even the cpu E2E may overlap with the same metrics and same namespace etc.. so is there a chance of a false positive?

test/metric_dimension/metrics_append_dimension_test.go
```diff
@@ -4,13 +4,13 @@
 //go:build linux && integration
```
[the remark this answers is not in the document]
ChenaLee: Same comment as I made to Seth. Could you take a look and tell me what I'm missing please?

test/metric_dimension/metrics_append_dimension_test.go
ChenaLee: Oh we totally can. I just had added one test case that was the easiest to check bare minimum of does it actually not add anything. Zhenyu had assigned Thomas to work on adding Append Dimension tests since he implemented the feature, but I don't think he got to it. I will talk to Zhenyu

test/metric_dimension/metrics_aggregate_dimension_test.go
sky333999: I think Seth might have removed that flag. I think when you rebase, we can confirm.

test/metric_dimension/metrics_append_dimension_test.go
taohungyang: yeah...I have a pending task to add that but didn't get time to push the PR. And yes the plan is to add at least one more test here for append dimensions.
Just to confirm that this PR is for aggregation but not append dimension, right? (Since I see `append_dimensions` in the new config.json, but, if I understand it correctly, that's just a requirement to use `aggregation_dimensions`, right?)

test/metric_dimension/metrics_aggregate_dimension_test.go
ChenaLee: ah yes he did

test/metric_dimension/metrics_aggregate_dimension_test.go
ChenaLee: This configuration was generating metrics with no dimensions. Like namespace and metric name exists but nothing else. And so validation also sent no dimension with the get request, and got the metric. cpu e2e wouldn't have generated metrics with no dimensions in my understanding?

ChenaLee:
Addressed comments

sethAmazon (approved):

SaxyPandaBear (dismissed):

test/metric_dimension/metrics_aggregate_dimension_test.go
sky333999: True.. maybe not an immediate concern but in future if someone tweaks the dimensions or something, its hard for someone to know its impact on other tests.
Which leads me to a more broader unrelated to your PR question, why not just use a diff metric namespace for each test.

SaxyPandaBear (approved):

sky333999 (approved):

sethAmazon (approved):

test/metric_dimension/metrics_aggregate_dimension_test.go
ChenaLee: yeah true. General coverage issue for all these tests. Should we add a backlog item or something to increase coverage in the future?

adam-mateen: Yes to using _something_ unique so that test1 is not impacted by metrics reported for test2.
Each test using a unique namespace seems good to me.
Is there a limit to the number of namespaces?

PR #99, opened 2023-01-24
Do not use prom_metric_type and comment out summary checks for prometheus

sky333999:
# Description of the issue
This PR addresses two issues that were highlighted when running the prometheus test on our staging repo (the test was working fine on our public repo).
1. Summary metrics need some additional work for CCWA to behave the same as CWA. Hence the current assertions will fail for CCWA.
2. `prom_metric_type` is auto-added as a label/dimension by `awsemfexporter` but it is not considered as a valid dimension when doing the dimension filtering. Hence if someone uses `prom_metric_type` as a dimension filter in their config, it currently wont work. This is a very rare usecase but this test does happen to do it. 

# Description of changes
1. We have a backlog item to track this. So until then, commenting out the checks for summary metrics.
2. I'll create a backlog item to track the fix but tweaking the test for now to NOT use `prom_metric_type` as a dimension filter.

# License

# Tests
Ran prometheus test on fork [successfully](https://github.com/sky333999/private-amazon-cloudwatch-agent-staging/actions/runs/3991080364/jobs/6845533725).

### DIFF
--- a/test/metric_value_benchmark/agent_configs/prometheus_config.json
+++ b/test/metric_value_benchmark/agent_configs/prometheus_config.json
@@ -20,10 +20,10 @@
               "label_matcher": "^yes$",
               "dimensions": [
                 [
-                  "prom_metric_type"
+                  "prom_type"
                 ],
                 [
-                  "prom_metric_type",
+                  "prom_type",
                   "quantile"
                 ]
               ],
--- a/test/metric_value_benchmark/prometheus_test.go
+++ b/test/metric_value_benchmark/prometheus_test.go
@@ -26,27 +26,27 @@ var _ test_runner.ITestRunner = (*PrometheusTestRunner)(nil)
 //go:embed agent_configs/prometheus.yaml
 var prometheusConfig string
 
-const prometheusMetrics = `prometheus_test_untyped{include="yes"} 1
+const prometheusMetrics = `prometheus_test_untyped{include="yes",prom_type="untyped"} 1
 # TYPE prometheus_test_counter counter
-prometheus_test_counter{include="yes"} 1
+prometheus_test_counter{include="yes",prom_type="counter"} 1
 # TYPE prometheus_test_counter_exclude counter
-prometheus_test_counter_exclude{include="no"} 1
+prometheus_test_counter_exclude{include="no",prom_type="counter"} 1
 # TYPE prometheus_test_gauge gauge
-prometheus_test_gauge{include="yes"} 500
+prometheus_test_gauge{include="yes",prom_type="gauge"} 500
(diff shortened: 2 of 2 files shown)


test/metric_value_benchmark/agent_configs/prometheus_config.json
```diff
@@ -20,10 +20,10 @@
               "label_matcher": "^yes$",
               "dimensions": [
                 [
-                  "prom_metric_type"
+                  "prom_type"
                 ],
                 [
```
khanhntd, line 25: **Nit**: If comment out summary, should it be the same for dimension "quantile"? Yes, it will work but I assume it would be better to come along when we officially support summary with CCWA

khanhntd (approved):
LGTM. Only small nit picks.

test/metric_value_benchmark/agent_configs/prometheus_config.json
sky333999: Agree.. I thought Id keep the changes minimal to revert later on. Since the metrics to measure itself doesnt have summary, the quantile or the extra dimension below wont be used.
If itll be cleaner to comment out in other places as well, I can update it.

test/metric_value_benchmark/prometheus_test.go
```diff
@@ -26,27 +26,27 @@ var _ test_runner.ITestRunner = (*PrometheusTestRunner)(nil)
...
+prometheus_test_summary{include="yes",quantile="1",prom_type="summary"} 5.5
 # TYPE prometheus_test_histogram histogram
-prometheus_test_histogram_sum{include="yes"} 300
-prometheus_test_histogram_count{include="yes"} 75
-prometheus_test_histogram_bucket{include="yes",le="0"} 1
-prometheus_test_histogram_bucket{include="yes",le="0.5"} 2
-prometheus_test_histogram_bucket{include="yes",le="2.5"} 3
-prometheus_test_histogram_bucket{include="yes",le="5"} 4
-prometheus_test_histogram_bucket{include="yes",le="+Inf"} 5
+prometheus_test_histogram_sum{include="yes",prom_type="histogram"} 300
```
sethAmazon, line 43: Is the prom_type because of a bug in emf exporter? I was able to find prom_metric_type as a label in emf, and this worked for the prod agent.

sethAmazon (approved):

test/metric_value_benchmark/prometheus_test.go
sky333999: Yes. They add it as a label in the awsemfexporter but dont consider it when filtering for dimensions, hence it is a minor bug with the exporter.

ChenaLee (approved):

PR #101, opened 2023-01-25
Add A Canary Test

sethAmazon:
# Description of the issue
We need a way to auto-deploy a canary on release of agent. 

# Description of changes
Add the terraform required to add a canary instance. 

# License

# Tests
Tested in private repo release.


test/canary/resources/canary_config.json
```diff
@@ -0,0 +1,11 @@
+{
```
sethAmazon, line 1: This is placeholder config. Change when @jefchien and @adam-mateen figure out what metrics the canary should use.

terraform/ec2/linux/main.tf
```diff
@@ -142,6 +144,56 @@ resource "null_resource" "integration_test" {
   ]
 }
 
+// Only run on canary test
```
khanhntd, line 147: Would it be best to separate the canary workflow though? Since I believe the canary is for confirm if the latest agent is able to run. Therefore, does not need some resources such as `efs` and it won't stop here since we have container use case too

terraform/ec2/README.md
```diff
@@ -169,6 +169,7 @@ Outputs:
         "s3:GetObject",
         "s3:GetObjectAcl",
         "s3:PutObject",
+        "s3:DeleteObject",
```
khanhntd, line 172: Why we need deletion for s3 though?

test/canary/resources/canary_config.json
adam-mateen: This is fine with me.
The purpose of this is to test downloading and installing from S3 yields the expected version.

test/canary/canary_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func init() {
+	environment.RegisterEnvironmentMetaDataFlags(envMetaDataStrings)
+}
+
+// This only copies the canary config and starts the agent. No validation is done here.
+// Sanity test will run earlier in the terraform execution
+func TestCanary(t *testing.T) {
+	common.CopyFile(configInputPath, configOutputPath)
+	common.StartAgent(configOutputPath, true)
```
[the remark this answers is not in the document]
sethAmazon: Yes this happens in the remote exec before this test is run.

terraform/ec2/README.md
sethAmazon: Delete old canary instance. See other pr for private repo.

terraform/ec2/linux/main.tf
```diff
@@ -142,6 +144,56 @@ resource "null_resource" "integration_test" {
...
+      private_key = local.private_key_content
+      host        = aws_instance.cwagent.public_ip
+    }
+  }
+
+  #Run sanity check and integration test
+  provisioner "remote-exec" {
+    inline = [
+      "echo prepare environment",
+      "export LOCAL_STACK_HOST_NAME=${var.local_stack_host_name}",
```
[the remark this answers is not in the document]
sethAmazon: removed

test/canary/canary_test.go
```diff
@@ -0,0 +1,25 @@
+package canary
```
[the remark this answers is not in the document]
sethAmazon: added

terraform/ec2/linux/main.tf
sethAmazon: EFS is needed for consistent disk metrics across runs. If disk is required metric to track, it is good to have efs. As for adding an ecs canary, I would say extend the ecs terraform code. Adding a new terraform package for canary means copying lots of code and for most things changing in both places on change. There is very little difference between canary and a regular integration test.

test/canary/canary_test.go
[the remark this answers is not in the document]
adam-mateen: Can you also check the output of `/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent --version` and verify it is the new version.

You could also check `rpm -qa | grep amazon-cloudwatch-agent`.

Not sure the right way to know what the expected new version is...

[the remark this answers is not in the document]
adam-mateen: If not feasible, we can leave it as a manual step in the MCM.

terraform/ec2/linux/main.tf
khanhntd: I understand your point but somehow, I would disagree with that. 
* The whole point of canary is to verify if the agent is running normally with the latest version, its not confirming the metrics are sending. The metrics are sending is part of the integration test/feature testing/plugin testing 
* I would disagree regarding extending the ecs terraform code. Even though I agree the major use case currently are EC2, we are moving to container (which is a trending based on Kevin). so there is a possible that we will only use ECS/EKS (be it on EC2 or Fargate) to do it canary and its covers EC2
*  Somehow I'm not convinced though regarding `difference  between canary and a regular integration test`.

test/canary/resources/canary_config.json
khanhntd: If its only running the agent, maybe create a SSM config and let's the agent pull so that we don't have to create a separate config/test for it.

terraform/ec2/linux/main.tf
sethAmazon: @jefchien is setting up alarms to the canary. That is the validation.

test/canary/canary_test.go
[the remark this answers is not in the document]
sethAmazon: Added

test/canary/resources/canary_config.json
sethAmazon: Talked with team, we decided to go with config.json instead of ssm since that will add an extra layer of obfuscation to the config. We would rather it be in the test dir code because it is easier to find.

khanhntd: No I meant create a default template config in the terraform test, create it also in the ssm by using ssm resources and running the agent with /amazon-cloudwatch-agent-ctl --config ssm-config so we don't need to have a separate script for it.

terraform/ec2/linux/main.tf
```diff
@@ -96,7 +96,10 @@ resource "null_resource" "integration_test" {
       "echo clone and install agent",
       "git clone --branch ${var.github_test_repo_branch} ${var.github_test_repo}",
       "cd amazon-cloudwatch-agent-test",
-      "aws s3 cp s3://${var.s3_bucket}/integration-test/binary/${var.cwa_github_sha}/linux/${var.arc}/${var.binary_name} .",
+      // If canary fetch the latest binary else fetch commit binary
+      var.is_canary ?
+      "aws s3 cp s3://${var.s3_bucket}/release/amazon_linux/${var.arc}/latest/${var.binary_name} ."
+      : "aws s3 cp s3://${var.s3_bucket}/integration-test/binary/${var.cwa_github_sha}/linux/${var.arc}/${var.binary_name} .",
```
adam-mateen, line 102: Looks fine to me.
But if we are passing the bucket to the `go test` command later on, then the test code could download the RPM instead of doing it here, right?

adam-mateen (dismissed):

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -66,7 +67,7 @@ type BaseTestRunner struct {
 	DimensionFactory dimension.Factory
 }
 
-func (t *ECSTestRunner) Run(s *MetricBenchmarkTestSuite, e *environment.MetaData) {
+func (t *ECSTestRunner) Run(s test_runner.ITestSuite, e *environment.MetaData) {
```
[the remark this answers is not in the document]
sethAmazon: This fixes a compile error. Unrelated to the canary. I can remove.

test/canary/canary_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+// This only copies the canary config and starts the agent and confirm version is correct.
+// Sanity test will run earlier in the terraform execution
+func TestCanary(t *testing.T) {
+	metaData := environment.GetEnvironmentMetaData(envMetaDataStrings)
+	common.CopyFile(configInputPath, configOutputPath)
+	common.StartAgent(configOutputPath, true)
+	common.RunCommand("aws s3 cp s3://" + metaData.Bucket + "/release/CWAGENT_VERSION .")
+	expectedVersion, _ := common.RunCommand(common.CatCommand + agentVersionPath)
+	actualVersion, _ := common.RunCommand(common.CatCommand + agentVersionPath)
```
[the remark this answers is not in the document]
khanhntd: Same. Intend to use create a validator to avoid using the shell commands. However, need to have a POC before consult with Seth and Chena

internal/common/agent_util_unix.go
```diff
@@ -14,12 +14,18 @@ import (
...
+	ConfigOutputPath   = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	Namespace          = "CWAgent"
+	Host               = "host"
+	AgentLogFile       = "/opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log"
+	installRetryNumber = 10
+	installRetryTime   = 30 * time.Second
+	debInstall         = "deb"
+	rpmInstall         = "rpm"
+	debInstallCommand  = "sudo dpkg -i -E ./amazon-cloudwatch-agent.deb"
+	rpmInstallCommand  = "sudo rpm -U ./amazon-cloudwatch-agent.rpm"
```
sethAmazon, line 28: This will ensure that we fail if the new version is not grater than the current version. Thus, we fail the canary deployment if you gave the wrong version number.

test/canary/resources/canary_config.json
adam-mateen: Canary test does not need to verify downloading configuration files from SSM.
That could be test earlier during the "integration test" phase.

If/when we release CCWA to SSM. In other words we push new SSM Documents that customers use to download and configure CCWA, then we would want a canary test that pulls them and verified they are the expected Documents.

adam-mateen (approved):

test/canary/resources/canary_config.json
khanhntd: To clarify, the canary test are only running the CWA. Therefore, we don't need a **separate go file**  to run it, so an easiest way would be create a default config in terraform and using terraform to put it in the SSM configuration

test/canary/canary_test.go
```diff
@@ -0,0 +1,45 @@
...
+	environment.RegisterEnvironmentMetaDataFlags(envMetaDataStrings)
+}
+
+// This downloads the latest agent, copies the canary config, starts the agent, confirm version is correct, and validates no errors in logs.
+// Sanity test runs earlier in the terraform code execution
+func TestCanary(t *testing.T) {
+	// Canary set up
+	common.CopyFile(configInputPath, configOutputPath)
+	common.StartAgent(configOutputPath, true)
+
```
jefchien, line 34: Don't we still need to download the agent version from S3?

terraform/ec2/linux/main.tf
```diff
@@ -24,6 +24,9 @@ resource "aws_key_pair" "aws_ssh_key" {
 locals {
   ssh_key_name        = var.ssh_key_name != "" ? var.ssh_key_name : aws_key_pair.aws_ssh_key[0].key_name
   private_key_content = var.ssh_key_name != "" ? var.ssh_key_value : tls_private_key.ssh_key[0].private_key_pem
+  // Canary downloads latest binary. Integration test downloads binary connect to git hash.
+  downloadBinary = var.is_canary ? "aws s3 cp s3://${var.s3_bucket}/release/amazon_linux/${var.arc}/latest/${var.binary_name} ."
+    : "aws s3 cp s3://${var.s3_bucket}/integration-test/binary/${var.cwa_github_sha}/linux/${var.arc}/${var.binary_name} ."
```
jefchien, line 29: super minor nit: You could also just store the S3 URI and then in `inline` have it do the s3 cp. Like `aws s3 cp ${local.binary_uri} .`

```suggestion
  binary_uri = var.is_canary ? "s3://${var.s3_bucket}/release/amazon_linux/${var.arc}/latest/${var.binary_name}"
    : "s3://${var.s3_bucket}/integration-test/binary/${var.cwa_github_sha}/linux/${var.arc}/${var.binary_name}"
```

adam-mateen (approved):

jefchien (dismissed):

terraform/ec2/linux/main.tf
```diff
@@ -24,6 +24,9 @@ resource "aws_key_pair" "aws_ssh_key" {
 locals {
   ssh_key_name        = var.ssh_key_name != "" ? var.ssh_key_name : aws_key_pair.aws_ssh_key[0].key_name
   private_key_content = var.ssh_key_name != "" ? var.ssh_key_value : tls_private_key.ssh_key[0].private_key_pem
+  // Canary downloads latest binary. Integration test downloads binary connect to git hash.
+  binary_uri = var.is_canary ? "${var.s3_bucket}/release/amazon_linux/${var.arc}/latest/${var.binary_name} ."
```
jefchien, line 28: nit: `.` isn't part of the URI.

jefchien (approved):

jefchien (approved):

adam-mateen (approved):

PR #103, opened 2023-01-27
Generalize AWS service API calls 

khanhntd:
# Description of the issue
Create a general interface for each of the service (e.g ec2, ecs, ...) and share the same resources (e.g config, context). Moreover, return error for all API call and let's the test case decide the outcome of the API call.

# Description of changes
* Create a [AWS Config structure](https://github.com/aws/amazon-cloudwatch-agent-test/pull/103/files#diff-3a6948fe838e6def59426c2b59860eb1c7fa73c8b5bd2468a7132859ed6f5f77) that shares the same resources (e.g ctx, config) for other API calls. 
* Add [DynamoDB API Call to AWS Config Structure](https://github.com/aws/amazon-cloudwatch-agent-test/pull/103/files#diff-06922a0a54f2d4ec5d262269aa687bb154b542cde78bc15b5af79c9c5e5e8458)
* Instead of validate in the service call itself, all API will return errors (e.g [CWL API](https://github.com/aws/amazon-cloudwatch-agent-test/pull/103/files#diff-481de9e23b7d675ebe3f425909e7ce4585e30859cddc0183817e636124dd07f2L72))
* Delete some functions since the exist one has cover enough use case (e.g [SendItem](https://github.com/aws/amazon-cloudwatch-agent-test/pull/103/files#diff-ec56f4b9b9afc2dd6d2e563863f82a56c58e2965989bef7a3937cad38a5fd03eL203-L234) - checking if package is exist and add the item; however, [UpdateItem](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html) already done that)

# License

# Tests
By running the changes on my fork, here is the result of it: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4036724893


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/aws/amazon-cloudwatch-agent-test
 go 1.18
 
 require (
+	github.com/aws/aws-sdk-go v1.44.188
```
SaxyPandaBear, line 6: Why do we need to introduce v1 of the SDK?

internal/awsservice/cloudwatchlogs.go
```diff
@@ -6,37 +6,73 @@ package awsservice
...
+
+	// DeleteLogGroup cleans up a log group  by name. This gracefully handles
+	// ResourceNotFoundException errors from calling the APIs
+	DeleteLogGroup(logGroup string)
+
+	// IsLogGroupExist confirms whether the logGroupName exists or not
+	IsLogGroupExist(logGroup string) bool
+}
+
+type cloudwatchLogConfig struct {
```
SaxyPandaBear, line 51: nit - why is this named config? This isn't a configuration struct. It's a wrapper around an SDK client

internal/awsservice/config.go
```diff
@@ -0,0 +1,46 @@
...
+type awsServiceAPI struct {
+	cxt context.Context
+
+	Ec2API   ec2API
+	EcsAPI   ecsAPI
+	CwmAPI   cwmAPI
+	CwlAPI   cwlAPI
+	SsmAPI   ssmAPI
+	ImdsAPI  imdsAPI
+	DnmdbAPI dnmdbAPI
```
SaxyPandaBear, line 25: What is the benefit in doing this? What are we generalizing? Why can't we just use the SDK client directly?

khanhntd: By implementing this, we have the following benefits:
* Every time we call the SDK Client, we need to call getClients (get context and get SDK client e.g [CloudWatchLogs](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/internal/awsservice/cloudwatchlogs.go#L34)). Therefore, wasteful call on every API
* Resources are not being shared and every time we create a new SDK, we would need to create [another variables](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/internal/awsservice/cloudwatchlogs.go#L25-L26)shares the same resources (e.g ctx, config). Therefore, wasteful resources being allocated
* Allow the developers to look at the interface to know what we currently have and what we allow the developer to use without dive deep into the usage of each call 
* etc

internal/awsservice/cloudwatchlogs.go
khanhntd: Hmm. I guess at that time I use Config as a name for  the implementation details. However, I'm fine to changing to `{{service}}.SDK`

go.mod
khanhntd: Nice catch. We don't though, I mixed it up with another commit from another branch. Sorry for this.

internal/awsservice/config.go
SaxyPandaBear: > Therefore, wasteful call on every API

What is the waste? We are not running out of host resources running this code. What is the point of pre-optimizing when there is no problem? 

> Allow the developers to look at the interface to know what we currently have and what we allow the developer to use without dive deep into the usage of each call

This is the point of the existing SDK. Why do we need to add a layer on top of that? 

> etc

Such as..?

jefchien: I also don't really understand why this is needed. `context.Background()` isn't an expensive resource (it's a singleton https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/context/context.go;l=208). If we were using the context to do something (cancellation, timeouts), then it might make sense, but I don't think we're saving anything by doing this.

Some convenience functions might be useful, but those can be done without the interfaces and layers.

sethAmazon: Maybe a compromise is an utl func that returns the generated sdk since the sdk is generated the same way every single time. This can give us the reduce in lines of code without having to write an abstraction layer.

khanhntd: Hmm !!!! Yeah you rights are right. Maybe I'm a obsess with the design a little bit. I'm fine with not using an abstraction layer too.

terraform/performance/main.tf
```diff
@@ -67,15 +68,13 @@ resource "null_resource" "integration_test" {
...
-      "echo prepare environment",
       "export AWS_REGION=${var.region}",
       "export PATH=$PATH:/snap/bin:/usr/local/go/bin",
-      "echo run integration test",
       "cd ~/amazon-cloudwatch-agent-test",
       "export SHA=${var.cwa_github_sha}",
       "export SHA_DATE=${var.cwa_github_sha_date}",
       "export PERFORMANCE_NUMBER_OF_LOGS=${var.performance_number_of_logs}",
-      "go test ${var.test_dir} -p 1 -timeout 1h -v"
+      "go test ${var.test_dir} -p 1 -timeout 1h  -v"
```
adam-mateen, line 77: unnecessary added whitespace

terraform/performance/main.tf
```diff
@@ -49,6 +49,7 @@ resource "null_resource" "integration_test" {
       "echo sha ${var.cwa_github_sha}",
       "cloud-init status --wait",
       "echo clone and install agent",
+      "rm -rf amazon-cloudwatch-agent-test",
```
adam-mateen, line 52: Instead of hardcoding the folder name, can you use `${var.github_test_repo`?

khanhntd: Forgot to delete it.Yes

terraform/performance/main.tf
khanhntd: Done !!! Small fixed !!!

sky333999:
So what's the final vote on whether we push this or not? I see two against and one on-the-fence. Im on-the-fence as well.

It does clean up the code but its going to hard to maintain abstractions over time (especially if someone introduces new clients or usages, they would need to know that some layer of abstraction already exists or at the least, the reviewers need to know to point it out).

internal/awsservice/ec2.go
```diff
@@ -4,46 +4,18 @@
...
-func GetInstancePrivateIpDns(instanceId string) (*string, error) {
-	instanceData, err := DescribeInstances([]string{instanceId})
-	if err != nil {
-		return nil, err
-	}
-
-	return instanceData.Reservations[0].Instances[0].PrivateDnsName, nil
-}
+// GetInstancePrivateIpDns returns Instance Private IP Address
+func GetInstancePrivateIpDns(instanceId string) (string, error) {
```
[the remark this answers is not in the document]
khanhntd:
```
Is it here just as a utility for future?
```
Yes that what I'm assuming. Therefore, not deleting it before final confirmation with Thomas.

test/performancetest/resources/config.json
```diff
@@ -4,16 +4,8 @@
...
-			[
-				"InstanceId"
-			]
-		],
 		"append_dimensions": {
-			"AutoScalingGroupName": "${aws:AutoScalingGroupName}",
-			"ImageId": "${aws:ImageId}",
-			"InstanceId": "${aws:InstanceId}",
-			"InstanceType": "${aws:InstanceType}"
+			"InstanceId": "${aws:InstanceId}"
```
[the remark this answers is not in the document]
khanhntd: There are not currently even with the `append_dimensions`. The only thing that going to be published are the metric and the value of it.

test/performancetest/performance_test.go
```diff
@@ -126,34 +119,31 @@ func TestPerformance(t *testing.T) {
...
 			LogGroupName:  "{instance_id}",
-			LogStreamName: logStream,
+			LogStreamName: fmt.Sprintf("{instance_id}/tmp%d", i+1),
 			Timezone:      "UTC",
 		})
 
-		logStreams = append(logStreams, logStream)
 	}
 
 	log.Printf("Writing config file with %d logs to ./resources/config%d.json\n", logNum, logNum)
```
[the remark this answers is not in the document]
khanhntd: I don't think there is any t testing to use `t.logf` here

test/metric_value_benchmark/ecs_daemon_base_test_runner.go
```diff
@@ -35,14 +36,14 @@ type ECSAgentRunStrategy struct {
 func (r *ECSAgentRunStrategy) runAgent(e *environment.MetaData, configFilePath string) error {
 	b, err := os.ReadFile(configFilePath)
 	if err != nil {
-		return fmt.Errorf("Failed while reading config file")
+		return fmt.Errorf("failed while reading config file")
```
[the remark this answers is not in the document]
khanhntd: Yes. We can leverage the golangci-link that I have introduced to [apply the lint](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/.golangci.yml)

test/performancetest/performance_test.go
```diff
@@ -51,20 +51,13 @@ func TestPerformance(t *testing.T) {
...
-	configFilePath, logStreams, err := GenerateConfig(logNum)
+	configFilePath, err := GenerateConfig(logNum)
 	if err != nil {
 		t.Fatalf("Error: %v", err)
 	}
 
-	//defer deleting log group and streams
-	//defer deleting log group first because golang handles defers in LIFO order
-	//and we want to delete the log group after deleting the log streams
 	defer awsservice.DeleteLogGroup(instanceId)
```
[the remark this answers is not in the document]
khanhntd:
```
Does deleting a log group by default also delete any log streams in it?
```
Yes. I'm even not sure why we `DeleteLogGroupAndLogStream` to begin with. However, in the performance test. the LogStream are being generated dynamically. Therefore, if using  `DeleteLogGroupAndLogStream` it will create throttle on backend side. To simplify things, only use `DeleteLogGroup`

khanhntd:
> So what's the final vote on whether we push this or not? I see two against and one on-the-fence. Im on-the-fence as well.

The original change is using the interface. That's why we have two against of using interface. However, I have changed back to the original way we are doing it.

```
It does clean up the code but its going to hard to maintain abstractions over time (especially if someone introduces new clients or usages, they would need to know that some layer of abstraction already exists or at the least, the reviewers need to know to point it out).
```

That's why the `source-code editor` always have hint to reduce the abstraction as much as possible
```
func IsMetricSampleCountWithinBound(metricName string, namespace string, dimensions []types.Dimension, startTime time.Time, endTime time.Time, lowerBoundInclusive int, upperBoundInclusive int, periodInSeconds int32) bool
IsMetricSampleCountWithinBound checking if certain metric's sample count is within the predefined bound interval
```
However, yes there is no guarantee of  that. That's why I introduce the interface to make it more simpler but not bring much values what we currently have.

internal/awsservice/dynamodb.go
```diff
@@ -0,0 +1,61 @@
...
+package awsservice
+
+import (
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+
+func AddPacketIntoDatabase(databaseName, conditionExpression string, packet interface{}, expressAttributesNames map[string]string) error {
```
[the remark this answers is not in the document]
khanhntd: I'm fine with that too

test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -97,27 +95,15 @@ func TestNumberMetricDimension(t *testing.T) {
...
-		}
-	}
-	dimensionFilter[appendDimension-1] = types.DimensionFilter{
-		Name:  aws.String(instanceId),
-		Value: aws.String(ec2InstanceId),
+			err = awsservice.ValidateMetric(parameter.metricName, namespace, dimensionFilter)
+			if err != nil {
+				t.Fatalf("Validate metrics failed: %v", err)
+			}
+		})
```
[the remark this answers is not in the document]
khanhntd: Yup sure. Nice catch !!! Thanks Kaushik

test/performancetest/performance_query_utils.go
```diff
@@ -103,8 +103,8 @@ func GetConfigMetrics(configPath string) ([]string, []string, error) {
 	return metricNames, ids, nil
 }
 
-// GenerateGetMetricInputStruct generates the struct required to make a query request to cloudwatch's GetMetrics
-func GenerateGetMetricInputStruct(ids, metricNames []string, instanceId string, timeDiff int) (*cloudwatch.GetMetricDataInput, error) {
+// GetMetrics generates the struct required to make a query request to cloudwatch's GetMetrics
```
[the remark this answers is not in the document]
khanhntd: Intend to delete it in the next PR but fine to me !

internal/awsservice/cloudwatchmetrics.go
```diff
@@ -4,76 +4,43 @@
...
-func ValidateSampleCount(t *testing.T, metricName, namespace string, dimensions []types.Dimension,
-	startTime time.Time, endTime time.Time,
-	lowerBoundInclusive int, upperBoundInclusive int, periodInSeconds int32) bool {
-	cwmClient, clientContext, err := GetCloudWatchMetricsClient()
-	if err != nil {
-		t.Fatalf("Error occurred while creating CloudWatch Logs SDK client: %v", err.Error())
-	}
-
+// isMetricSampleCountWithinBoundInclusive checking if certain metric's sample count is within the predefined bound interval
+func IsMetricSampleCountWithinBoundInclusive(
```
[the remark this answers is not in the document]
khanhntd: Fair to me !! Thanks Kaushik for recommending.

PR #105, opened 2023-01-27
Add Basic Canary Config

sethAmazon:
# Description of the issue
Canary has no metrics about the agent. 

# Description of changes
Add procstat metrics about the agent. 

# License

# Tests
Manual run the canary test.

### DIFF
--- a/test/canary/resources/canary_config.json
+++ b/test/canary/resources/canary_config.json
@@ -1,9 +1,31 @@
 {
   "metrics": {
+    "namespace": "CanaryTest",
+    "aggregation_dimensions" : [[]],
     "metrics_collected": {
-      "mem": {
-        "measurement": [
-          "mem_used_percent"
+      "procstat": [
+        {
+          "pid_file": "/opt/aws/amazon-cloudwatch-agent/var/amazon-cloudwatch-agent.pid",
+          "measurement": [
+            "cpu_usage",
+            "memory_data",
+            "memory_rss"
+          ]
+        }
+      ]
+    }
+  },
+  "logs": {
+    "logs_collected": {
+      "files": {
+        "collect_list": [
+          {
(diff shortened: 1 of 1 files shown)


test/canary/resources/canary_config.json
```diff
@@ -1,9 +1,28 @@
 {
   "metrics": {
```
jefchien, line 2: Need to include aggregate dimensions so we can get a metric without dimensions. Looking at the canary metrics from its brief run, it adds a metric with a couple dimensions including the host IP, which makes it difficult to use.

test/canary/resources/canary_config.json
```diff
@@ -1,9 +1,28 @@
...
     "metrics_collected": {
-      "mem": {
-        "measurement": [
-          "mem_used_percent"
+      "procstat": [
+        {
+          "pid_file": "/opt/aws/amazon-cloudwatch-agent/var/amazon-cloudwatch-agent.pid",
+          "measurement": [
+            "cpu_usage",
+            "memory_data"
```
jefchien, line 10: We also typically use `memory_rss`.

test/canary/resources/canary_config.json
```diff
@@ -1,9 +1,28 @@
...
+          ]
+        }
+      ]
+    }
+  },
+  "logs": {
+    "logs_collected": {
+      "files": {
+        "collect_list": [
+          {
```
SaxyPandaBear, line 20: nit - is there any reason we wouldn't want log group retention for this?

test/canary/resources/canary_config.json
sethAmazon: Added

jefchien (approved):

SaxyPandaBear (approved):
I don't think we need it for 14 days. Maybe 3 or less. But it's fine

PR #107, opened 2023-01-27
Add test for global append dimension.

taohungyang:
# Description of the issue
This test makes sure ec2tagger tags metrics with the expected dimensions when users configure `append_dimensions` at the global metrics level.

# Description of changes
- Add a test for `append_dimensions` at global metrics level

# License

# Tests
Test succeeded against my fork 
https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/4046731013/jobs/6963800567


test/metric_dimension/global_append_dimensions_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Value: dimension.UnknownDimensionValue(),
+		},
+		{
+			Key:   "InstanceId",
+			Value: dimension.UnknownDimensionValue(),
+		},
+		{
+			Key:   "InstanceType",
+			Value: dimension.UnknownDimensionValue(),
+		},
```
SaxyPandaBear, line 65: Does this work? We're using append dimensions in the config so these should have actual values. When we query for an unknown dimension here, it's going to return nothing, isn't it?

test/metric_dimension/global_append_dimensions_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (t *GlobalAppendDimensionsTestRunner) GetAgentConfigFileName() string {
+	return "global_append_dimension.json"
+}
+
+func (t *GlobalAppendDimensionsTestRunner) GetMeasuredMetrics() []string {
+	return []string{"cpu_time_active"}
+}
+
+func (t *GlobalAppendDimensionsTestRunner) validateNoAppendDimensionMetric(metricName string) status.TestResult {
```
SaxyPandaBear, line 47: nit - function name doesn't reflect what this is doing.

internal/awsservice/imds.go
```diff
@@ -11,8 +11,20 @@ import (
...
+func GetImageId() string {
+	return GetImdsMetadata().ImageID
+}
+
+func GetInstanceType() string {
+	return GetImdsMetadata().InstanceType
+}
+
+// TODO: Refactor Structure and Interface for more easier follow that shares the same session
+func GetImdsMetadata() *imds.GetInstanceIdentityDocumentOutput {
```
SaxyPandaBear, line 27: We should cache the results of the metadata call so we don't ping IMDS a lot.

test/metric_dimension/global_append_dimensions_test.go
taohungyang: I checked the providers. As you can see here:
https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/metric/dimension/instanceid_provider.go#L59-L62
It's only when the value is unknown that the provider (instanceID/imageId/InstanceType) will retrieve the actual value from imds and use it to query the metrics

SaxyPandaBear: Ohh I guess I misunderstood how it works

internal/awsservice/imds.go
taohungyang: Added a variable to cached the results. Thanks

sethAmazon: Do we plan on calling this multiple times within the same test execution, that is needs to be cached?

sethAmazon (dismissed):

internal/awsservice/imds.go
```diff
@@ -11,8 +11,25 @@ import (
...
+	if indentityDoc != nil {
```
jefchien, line 30: nit: misspelled
```suggestion
	if identityDoc != nil {
```

taohungyang: Good eye...fixed it.

ChenaLee (approved):
This is awesome. Thanks!

PR #108, opened 2023-01-28
Delete uploading EFS and install EFS  from local source

khanhntd:
# Description of the issue
Delete uploading EFS and install EFS since the AMI Image build has been installed with the script. The reason for this would be:
* [Fixing the issue of the script creating package lock in Ubuntu OS](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4018285752/jobs/6903773647) (Debian too)  
* Abstract the script and built into the AMI layer.

# License

# Tests
By running the test on my fork https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4034318601


terraform/ec2/linux/variables.tf
```diff
@@ -93,15 +88,6 @@ variable "cwa_github_sha_date" {
   type    = string
   default = ""
 }
-variable "performance_number_of_logs" {
```
sethAmazon, line 96: This may need a change in the integ test yaml since terraform throws errors when you send a var not in the variables.

terraform/performance/variables.tf
```diff
@@ -34,11 +34,6 @@ variable "install_agent" {
   default     = "go run ./install/install_agent.go rpm"
 }
 
-variable "ca_cert_path" {
```
sethAmazon, line 37: Are you certain this is not still needed?

terraform/ec2/linux/main.tf
```diff
@@ -57,18 +57,11 @@ resource "null_resource" "mount_efs" {
...
 
-  provisioner "file" {
-    source = "install-efs-utils.sh"
-    destination = "/tmp/install-efs-utils.sh"
-  }
-
   provisioner "remote-exec" {
     # https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-helper-ec2-linux.html
     inline = [
-      "chmod +x /tmp/install-efs-utils.sh",
```
sethAmazon, line 68: @williazz will this work?

terraform/ec2/linux/variables.tf
khanhntd: Yes that's fine. EC2Linux does not use it. Only [Performance](https://github.com/aws/amazon-cloudwatch-agent/blob/main/.github/workflows/integrationTest.yml#L557-L574) use it

terraform/ec2/linux/main.tf
khanhntd: Yes. It will work (double check with integration test)

terraform/ec2/linux/main.tf
```diff
@@ -57,18 +57,11 @@ resource "null_resource" "mount_efs" {
...
-      "/tmp/install-efs-utils.sh",
-      "sudo mkdir ${var.efs_mount_point}",
-      "sudo mount -t efs -o tls ${aws_efs_file_system.efs.dns_name} ${var.efs_mount_point}/",
+      "sudo mkdir ~/efs-mount-point",
```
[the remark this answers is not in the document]
khanhntd: It's depends on the host you are in. For AL2, it would be `~/efs-mount-point` since there are the steps to be executed after SSH into the host.

terraform/ec2/linux/main.tf
williazz: not sure why that works but okay

terraform/performance/variables.tf
```diff
@@ -69,11 +64,6 @@ variable "cwa_github_sha" {
   default = ""
 }
 
-variable "cwa_test_github_sha" {
```
williazz, line 72: was this not being used anywhere?

williazz: nevermind, i see that we replaced with `github_test_repo_branch`

terraform/ec2/linux/main.tf
williazz: after offline call with khan, i see that we moved the install-efs-utils script to the AMI building pipelines

williazz (approved):
Thanks for moving `install-efs-utils` to the AMI pipelines. Still need to move efs-mount there next sprint

sethAmazon (approved):

PR #109, opened 2023-01-30
Add stress test and add metrics workload for both performance/stress

khanhntd:
# Description of the issue
_Describe the problem or feature in addition to a link to the issues._

# Description of changes

# License

# Tests


generator/test_case_generator.go
```diff
@@ -39,8 +41,9 @@ var testTypeToTestDirMap = map[string][]string{
 		"./test/collection_interval",
 		"./test/metric_dimension",
 	},
-	"ec2_performance": {
-		"./test/performancetest",
+	"ec2_performance": {},
```
khanhntd, line 44: Temporarily disable performance test and add it again in the next PR.

SaxyPandaBear (commented):
This has all of the changes from #70. Which one needs to be reviewed?

internal/awsservice/cloudwatchmetrics.go
```diff
@@ -66,12 +66,12 @@ func ValidateMetrics(t *testing.T, metricName, namespace string, dimensionsFilte
 
 }
 
-func ValidateSampleCount(t *testing.T, metricName, namespace string, dimensions []types.Dimension,
+func ValidateSampleCount(metricName, namespace string, dimensions []types.Dimension,
 	startTime time.Time, endTime time.Time,
 	lowerBoundInclusive int, upperBoundInclusive int, periodInSeconds int32) bool {
 	cwmClient, clientContext, err := GetCloudWatchMetricsClient()
 	if err != nil {
-		t.Fatalf("Error occurred while creating CloudWatch Logs SDK client: %v", err.Error())
```
sethAmazon, line 74: Is this okay?

internal/awsservice/cloudwatchmetrics.go
```diff
@@ -95,10 +94,6 @@ func ValidateSampleCount(t *testing.T, metricName, namespace string, dimensions
 		dataPoints = dataPoints + int(*datapoint.SampleCount)
 	}
 
-	t.Logf("Number of datapoints for start time %v with endtime %v and period %d "+
```
sethAmazon, line 98: Why remove?

sethAmazon:
When performance testing the internal, we learned logs are expensive. Should our logs performance test hit localstack? This may not be necessary since our internal performance out put is higher and runs indefinitely instead of only the time it takes to execute the test.

internal/awsservice/cloudwatchmetrics.go
khanhntd: Yes. I intend to generalize the call which does not rely on test and only return error or value instead of bool https://github.com/aws/amazon-cloudwatch-agent-test/pull/103. Also all of the calls for `ValidateSampleCount` is [there](https://github.com/aws/amazon-cloudwatch-agent-test/pull/109/files#diff-b9fb7a5de31d3d0134ed0bd2138c43847925d56c93a78db9f988a7d85cfabbc6R101)

khanhntd:
> When performance testing the internal, we learned logs are expensive. Should our logs performance test hit localstack? This may not be necessary since our internal performance out put is higher and runs indefinitely instead of only the time it takes to execute the test.

Its expensive but it should be worthwhile in exchange for gaining customers. We only need to define the custom major use case for logs (e.g running 1000 data logs and see what customers should expect) additional would be how if we increase 1 log, what would be the CPU request, Memory request in exchange for it

chadpatel:
Could this have been made in to smaller PRs.  Maybe starting with infra + hello world test, and then eventually adding the more complex tests.  This seems like it should be 2 or 3 PRs.  It is hard to review this many changes

khanhntd:
> Could this have been made in to smaller PRs. Maybe starting with infra + hello world test, and then eventually adding the more complex tests. This seems like it should be 2 or 3 PRs. It is hard to review this many changes

Some changes are moving to the general function. However, IMO, I don't think its a large scale change enough to introduce into many PRs since the only purpose is to validate the metric  (sample count, value, etc). If I am wrong, please correct me.

adam-mateen:
> > When performance testing the internal, we learned logs are expensive. Should our logs performance test hit localstack? This may not be necessary since our internal performance out put is higher and runs indefinitely instead of only the time it takes to execute the test.
> 
> Its expensive but it worthwhile to help us in know the characteristics of CWAgent, same for customers too. We only need to define the customer major use case for logs (e.g running 1000 data logs and see what customers should expect) additional would be how if we increase 1 log, what would be the CPU request, Memory request in exchange for it

Agree with everyone ;). I recommend start with the simple form of the test which uses the real CloudWatch Logs backend. If/when we decide cost is prohibitive we add a more complex solution using LocalStack, or even making it a test option on which backend to use.

khanhntd:
> Agree with everyone ;). I recommend start with the simple form of the test which uses the real CloudWatch Logs backend. If/when we decide cost is prohibitive we add a more complex solution using LocalStack, or even making it a test option on which backend to use.

The criteria for me when choosing test double would be:
* Determinism 
* Dependency construction
* Execution time

Cost maybe not the factor here since failing is a good thing as long as we able to identify the root cause ! Using `local stack` would make it harder to maintain or improve

internal/common/logs.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, logPath := range logPaths {
+		filePath := logPath //necessary weird golang thing
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err = writeToLogs(filePath, agentRunDuration, dataRate)
```
adam-mateen, line 32: It looks like you have a single `err` variable, and multiple goroutines will set it.
Unpredictible return value.

Maybe use an error counter that sums errors from all the files.

khanhntd: I use this in [metrics section but forgot to do that in logs](https://github.com/aws/amazon-cloudwatch-agent-test/pull/109/files#diff-781bb3810ede78771b4deb63bb6bb797a2d06434f0d844429b5ad16a20304e0fR19-R43).

internal/common/metrics.go
```diff
@@ -19,13 +18,10 @@ func StartSendingMetrics(receivers []string, agentRunDuration time.Duration, dat
 	var (
 		err      error
 		multiErr error
-		wg       sync.WaitGroup
```
adam-mateen: comment above still mentions "create wait group".

internal/common/metrics.go
```diff
@@ -37,8 +33,6 @@ func StartSendingMetrics(receivers []string, agentRunDuration time.Duration, dat
 			multiErr = multierr.Append(multiErr, err)
 		}(receiver, agentRunDuration)
 	}
-
-	wg.Wait()
```
adam-mateen: I think the wait group is still needed. Otherwise the returned `error` won't be accurate. The function returns before any `multiErr` variable can get updated.

validator/validators/stress/stress_validator.go
```diff
@@ -28,38 +28,38 @@ var (
...
+				"procstat_num_fds":     float64(19),
+				"net_bytes_sent":       float64(524000),
+				"net_packets_sent":     float64(520),
 			},
 		},
 		"10000": {
 			"statsd": {
-				"procstat_cpu_usage":   float64(14),
-				"procstat_memory_rss":  float64(95000000),
+				"procstat_cpu_usage":   float64(200),
```
adam-mateen, line 55: Quite the increase from 14% to 200%.
Why?

adam-mateen (approved):

test/statsd_stress/parameters.yml
```diff
@@ -0,0 +1,66 @@
+# Receivers that agent needs to tests
+receivers: ["statsd"]
+# Processors that agent needs to tests
+processors: []
+# Exporters that agent needs to tests
+exporters: []
```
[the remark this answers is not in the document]
khanhntd: Yes. It will works. The receivers/processors/exporters are only are referring plugins and does not generate from YAML or TOML but input directly. However, it is used to know what metrics load should we refer to.

internal/common/logs.go
```diff
@@ -0,0 +1,162 @@
...
+
+		case <-endTimeout:
+			return nil
+		}
+	}
+}
+
+// getLogFilePaths parses the cloudwatch agent config at the specified path and returns a list of the log files that the
+// agent will monitor when using that config file
+func getLogFilePaths(configPath string) ([]string, error) {
```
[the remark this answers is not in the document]
khanhntd: Yes. Agree it won't work for GLOB syntax. This function is leveraged from intern but I have some enhancement for next PR https://github.com/aws/amazon-cloudwatch-agent-test/pull/115

test/performancetest/resources/config.json
```diff
@@ -4,16 +4,8 @@
 		"run_as_user": "root"
 	},
 	"metrics": {
-		"aggregation_dimensions": [
-			[
-				"InstanceId"
```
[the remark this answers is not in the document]
khanhntd: Overall, aggregation with only InstanceID will return `InstanceID, AutoScalingGroupName, InstanceType, ImageId` and dimension of `InstanceID` which introduces extra dimension here. I'm fine with keeping it here too since I have updating it and decide to not using this feature for more visibility (it also does not make sense to combine multiple process together if using `aggregation_dimensions`)

validator/validators/stress/stress_validator.go
```diff
@@ -0,0 +1,231 @@
...
+
+var _ models.ValidatorFactory = (*StressValidator)(nil)
+
+func NewStressValidator(vConfig models.ValidateConfig) models.ValidatorFactory {
+	return &StressValidator{
+		vConfig: vConfig,
+	}
+}
+
+func (s *StressValidator) InitValidation() (err error) {
```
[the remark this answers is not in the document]
khanhntd: Duly note: will address it it the next PR.

validator/validators/stress/stress_validator.go
khanhntd: I increased from 1 minutes to 10 minutes. Therefore, every minutes it holds 10000 metrics and process it with the force flush interval of 60 before able to deliver it. Therefore, increasing the CPU Usage when aggregating with SEH, and holding more metrics in batch which increase value in Memory metrics

internal/common/metrics.go
khanhntd: Forgot to delete the go function. Therefore, we don't need to use wg here though since every sending metrics inside the metrics are a ticker.

validator/validators/stress/stress_validator.go
```diff
@@ -0,0 +1,231 @@
...
+			"statsd": {
+				"procstat_cpu_usage":   float64(14),
+				"procstat_memory_rss":  float64(95000000),
+				"procstat_memory_swap": float64(0),
+				"procstat_memory_vms":  float64(818000000),
+				"procstat_memory_data": float64(95000000),
+				"procstat_num_fds":     float64(9),
+				"net_bytes_sent":       float64(7800),
+				"net_packets_sent":     float64(24),
+			},
```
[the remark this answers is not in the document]
khanhntd: > Although I wonder if this is dependent on EC2 instance size...?

Yes based on my knowledge. However, I have not found a way for we to decouple from `Instance Size` but as a goal of our project, we prefer backward compatibility (including performance); therefore, it does not matter the Instance Size IIRC. FYI, I am using [t3x.large](https://instances.vantage.sh/aws/ec2/t3a.xlarge)

PR #122, opened 2023-02-08
Add E2E Sanity Test For Ethtool 

khanhntd:
# Description of the issue
Add E2E Sanity Test for Ethtool 

# Description of changes
* Added integration test to validate that the agent publishes metrics for the Ethtool input plugin

# License

# Tests
Ran on my fork: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4164762303
```
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 Values are all greater than or equal to zero for ethtool_linklocal_allowance_exceeded
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 >>>>>>>>>>>>>><<<<<<<<<<<<<<
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 >>>>>>>>>>>>>>Successful<<<<<<<<<<<<<<
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 ==============Ethtool==============
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 ==============Successful==============
null_resource.integration_test (remote-exec): ethtool_bw_in_allowance_exceeded       Successful
null_resource.integration_test (remote-exec): ethtool_bw_out_allowance_exceeded      Successful
null_resource.integration_test (remote-exec): ethtool_pps_allowance_exceeded         Successful
null_resource.integration_test (remote-exec): ethtool_conntrack_allowance_exceeded   Successful
null_resource.integration_test (remote-exec): ethtool_linklocal_allowance_exceeded   Successful
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 ==============================
null_resource.integration_test (remote-exec): 2023/02/11 04:10:42 >>>>>>>>>>>>>>><<<<<<<<<<<<<<<
null_resource.integration_test (remote-exec): >>>> Finished MetricBenchmarkTestSuite
null_resource.integration_test (remote-exec): --- PASS: TestMetricValueBenchmarkSuite (61.16s)
null_resource.integration_test (remote-exec):     --- PASS: TestMetricValueBenchmarkSuite/TestAllInSuite (61.16s)
null_resource.integration_test (remote-exec): PASS
```

# Tests
[_Describe what tests you have done._
](https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4149627810/jobs/7178735034)

### DIFF
--- a/test/metric/query-json.json
+++ b/test/metric/query-json.json
@@ -5,20 +5,27 @@
       "MetricStat": {
         "Metric": {
           "Namespace": "MetricValueBenchmarkTest",
-          "MetricName": "procstat_memory_swap",
+          "MetricName": "ethtool_bw_in_allowance_exceeded",
           "Dimensions": [
             {
               "Name": "InstanceId",
-              "Value": "i-X"
+              "Value": "i-014d094dcdfff54b5"
+            },
+            {
+              "Name": "driver",
+              "Value": "ena"
+            },
+            {
+              "Name": "interface",
+              "Value": "ens5"
             }
           ]
         },
         "Period": 60,
-        "Stat": "Average",
-        "Unit": "Bytes"
+        "Stat": "Average"
       }
     }
   ],
-  "StartTime": "2022-10-03T15:24:0000",
-  "EndTime": "2022-12-5T15:37:0000"
+  "StartTime": "2023-2-11T15:24:0000",
+  "EndTime": "2023-5-10T15:37:0000"
 }
--- a/test/metric/query-json.json
+++ b/test/metric/query-json.json
@@ -5,27 +5,20 @@
       "MetricStat": {
         "Metric": {
           "Namespace": "MetricValueBenchmarkTest",
-          "MetricName": "ethtool_bw_in_allowance_exceeded",
+          "MetricName": "procstat_memory_swap",
           "Dimensions": [
             {
               "Name": "InstanceId",
-              "Value": "i-014d094dcdfff54b5"
-            },
-            {
-              "Name": "driver",
-              "Value": "ena"
-            },
-            {
-              "Name": "interface",
-              "Value": "ens5"
+              "Value": "i-X"
             }
           ]
         },
         "Period": 60,
-        "Stat": "Average"
+        "Stat": "Average",
+        "Unit": "Bytes"
       }
     }
   ],
-  "StartTime": "2023-2-11T15:24:0000",
-  "EndTime": "2023-5-10T15:37:0000"
+  "StartTime": "2022-10-03T15:24:0000",
+  "EndTime": "2022-12-5T15:37:0000"
 }
--- a/test/metric_value_benchmark/ethtool_test.go
+++ b/test/metric_value_benchmark/ethtool_test.go
@@ -21,13 +21,13 @@ type EthtoolTestRunner struct {
 	test_runner.BaseTestRunner
 }
 
-var _ test_runner.ITestRunner = (*NetTestRunner)(nil)
+var _ test_runner.ITestRunner = (*EthtoolTestRunner)(nil)
 
 func (m *EthtoolTestRunner) Validate() status.TestGroupResult {
 	metricsToFetch := m.GetMeasuredMetrics()
 	testResults := make([]status.TestResult, len(metricsToFetch))
 	for i, name := range metricsToFetch {
-		testResults[i] = m.validateNetMetric(name)
+		testResults[i] = m.validateEthtoolMetric(name)
 	}
 
 	return status.TestGroupResult{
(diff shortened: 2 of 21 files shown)


test/metric/query-json.json
```diff
@@ -19,6 +19,6 @@
       }
     }
   ],
-  "StartTime": "2022-10-03T15:24:0000",
-  "EndTime": "2022-12-5T15:37:0000"
+  "StartTime": "2023-2-11T15:24:0000",
```
sethAmazon, line 22: Why does this need to change. Will it need to change again in the future?

sethAmazon (dismissed):

test/metric/query-json.json
khanhntd: Its used to validate the metrics are in CloudWatchMetrics by using query-json.json. There are no harm of changing it

williazz:
I noticed you fixed the network interface problems in your last commit. Please update your workflow run when you get the chance :)

test/metric/query-json.json
SaxyPandaBear: Why do we need to change it? If it's for manual validation, we should take out these values so we can script it.

test/metric_value_benchmark/ethtool_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/aws/amazon-cloudwatch-agent-test/test/status"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/test_runner"
+)
+
+type EthtoolTestRunner struct {
+	test_runner.BaseTestRunner
+}
+
+var _ test_runner.ITestRunner = (*NetTestRunner)(nil)
```
SaxyPandaBear, line 24: This is the wrong struct

SaxyPandaBear (changes requested):

test/metric_value_benchmark/ethtool_test.go
khanhntd: Nice catch !! Forgot about this. Thanks Andrew !

test/metric_value_benchmark/ethtool_test.go
```diff
@@ -0,0 +1,107 @@
...
+	test_runner.BaseTestRunner
+}
+
+var _ test_runner.ITestRunner = (*EthtoolTestRunner)(nil)
+
+func (m *EthtoolTestRunner) Validate() status.TestGroupResult {
+	metricsToFetch := m.GetMeasuredMetrics()
+	testResults := make([]status.TestResult, len(metricsToFetch))
+	for i, name := range metricsToFetch {
+		testResults[i] = m.validateNetMetric(name)
```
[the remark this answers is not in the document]
khanhntd: Yup. Thanks Kaushik. Nice catch !

sky333999 (dismissed):

test/metric/query-json.json
khanhntd: I'm good with not changing it. IMO, it would be better to add these flags into the script since developers need to know whether the flag exist but if added in the scripts, we only need to fix the time. It would be reduce time though.

williazz:
> I don't think its make any difference since adding a ENI would create another ENI but by default, the ENI is attach to the underlying OS (e.g Linux has eth0, adding one will create another eth1); however, the test case only validate the eth0.

```
│ Error: attaching EC2 Network Interface (eni-0aa9b4f05899572e1/i-086c63fce6be847a6): InvalidParameterCombination: You may not attach a network interface to an instance if they are not in the same availability zone
│ 	status code: 400, request id: 7a7b9cf7-a9b6-4108-9848-621b908dd7f3
│ 
│   with aws_network_interface.test,
│   on vpc.tf line 19, in resource "aws_network_interface" "test":
│   19: resource "aws_network_interface" "test" {
```

In your workflow run, `metric_value_benchmark` failed because of `aws_network_interface.test`, which you deleted in your last commit. I just wanted to confirm that it's no longer broken after your fix

khanhntd:
Run Integration Test successfully after the changes https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4165040024

test/metric_value_benchmark/ethtool_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	var (
+		dims   []types.Dimension
+		failed []dimension.Instruction
+	)
+
+	for _, iface := range ifaces {
+		if iface.Name == "eth0" || iface.Name == "ens5" {
+			dims, failed = m.DimensionFactory.GetDimensions([]dimension.Instruction{
```
SaxyPandaBear, line 72: Do we only care about these interfaces, and does it matter which one we get? The way this is set up, if we have both eth0 and ens5, we'll only pick one of them.

SaxyPandaBear (approved):

test/metric_value_benchmark/ethtool_test.go
khanhntd: Its depends on the firmware/BIOS/ etc quoting from the [sources](https://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-net_id.c#n20) for naming the interfaces. For more information, here is the [document for it ](https://unix.stackexchange.com/questions/134483/why-is-my-ethernet-interface-called-enp0s10-instead-of-eth0). However, in either way, we only have one naming convention for each OS (being it eth0 or ens5). So one is fine. However, there are plenty of ways to make it more general eth0 name ([example](https://www.itzgeek.com/how-tos/mini-howtos/change-default-network-name-ens33-to-old-eth0-on-ubuntu-16-04.html)) but its rather complex IMO. For your information, eth0 is a generic name for Linux and related families while ens5 is general for Ubuntu and Debian

khanhntd:
Run the fork again and here is the result of it https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4167248274

sky333999 (approved):

PR #123, opened 2023-02-08
Add family to enable Nvidia GPU test 

khanhntd:
# Description of the issue
Since family is not added as part of json [test_case_generator](https://github.com/aws/amazon-cloudwatch-agent-test/pull/123/files#diff-443c6e2df274e2dbc6fe7fce96a09a4b3ce8f62f587ee860c8dd56c8da56ff08R19), the NVIDIA_GPU test are not enabled since its only run when and only when family is linux. Therefore, adding linux to adding [back our NVIDIA test](https://github.com/aws/amazon-cloudwatch-agent/blob/main/.github/workflows/integrationTest.yml#L438)

# Description of changes

# License

# Tests
Running it in my fork https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4120644891


sethAmazon (approved):

sethAmazon:
How is family being used?

khanhntd:
`family` in the past is used for signal the parent of each OS. e.g Linux and Ubuntu can share the same parent while Windows does not.

sethAmazon:
Would linux be too generic, should it be centos/rhel/ubuntu ... instead of linux, since the package manager commands are different. I would be fine with all the centos based distros like al, ol and rhel all say centos.

khanhntd:
> Would linux be too generic, should it be centos/rhel/ubuntu ... instead of linux, since the package manager commands are different. I would be fine with all the centos based distros like al, ol and rhel all say centos.

Under the hood, they share the same application layer right. However, I'm fine with that too as long as we have use case for it.

PR #132, opened 2023-02-15
Restrict test hosts to HTTPS outbound, remove state bucket setup

SaxyPandaBear:
# Description of the issue
`terraform apply` would fail because the bucket it would try to create reuses an existing name. S3 bucket names are global. 

# License

# Tests
`terraform destroy --auto-approve` followed by `terraform apply --auto-approve`

```
module.common.random_id.testing_id: Creating...
...
Apply complete! Resources: 12 added, 0 changed, 0 destroyed.
```

---
Ran integration tests
```
null_resource.integration_test (remote-exec): --- PASS: TestRotatingLogsDoesNotSkipLines (72.25s)
null_resource.integration_test (remote-exec): PASS
null_resource.integration_test (remote-exec): ok  	github.com/aws/amazon-cloudwatch-agent-test/test/cloudwatchlogs	157.917s
null_resource.integration_test: Creation complete after 4m59s [id=863806897853973871]
```

Still able to submit PLE and PMD requests after restricting HTTPs access. Confirmed the security group setting manually by attempting to curl an https and http endpoint.


khanhntd (dismissed):

khanhntd (commented):
The purpose of this file would be to creating a resources that can be shared with multiple machine. However, if we are not able to use it since the s3 bucket  is global, then I think we could delete the [whole file](https://github.com/aws/amazon-cloudwatch-agent-test/blob/56341b6fe72e027d9db48f82e815875b03ad097e/terraform/setup/providers.tf#L27-L68)

williazz (dismissed):
is `terraform destroy` not disposing of the resources correctly? why was there a duplicate?

SaxyPandaBear:
> why was there a duplicate?

S3 bucket names are global across all AWS accounts.

terraform/setup/vpc.tf
```diff
@@ -10,9 +10,25 @@ resource "aws_security_group" "ec2_security_group" {
...
+  // Allow access to IMDS
+  egress {
+    from_port = 80
+    to_port   = 80
+    protocol = "TCP"
+    cidr_blocks = ["169.254.169.254/32"]
+  }
+
+  // Allow access to EFS. https://docs.aws.amazon.com/efs/latest/ug/troubleshooting-efs-mounting.html#mount-hangs-fails-timeout
+  egress {
```
khanhntd, line 28: [Its inbound right? ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/132/files/56341b6fe72e027d9db48f82e815875b03ad097e..bd6485bdeab50fea3af2ae4af4bea5cc917644e5#diff-3d3018c027b1f2ed0412ba5496014b7f2ce074be95bf6899a5543ad60031c68aR61-R66)

khanhntd (approved):

terraform/setup/vpc.tf
SaxyPandaBear: Inbound is already there. At the bottom of the doc, it says that if you don't allow all traffic outbound, you have to allow NFS outbound in the security group.

khanhntd: Interesting. Thanks for letting me know.

terraform/setup/vpc.tf
```diff
@@ -10,9 +10,25 @@ resource "aws_security_group" "ec2_security_group" {
   vpc_id = data.aws_vpc.default.id
 
   egress {
```
williazz, line 12: what is the purpose of these port mappings?

terraform/setup/vpc.tf
```diff
@@ -10,9 +10,25 @@ resource "aws_security_group" "ec2_security_group" {
...
-    from_port   = 0
-    to_port     = 0
-    protocol    = "-1"
+    from_port   = 443
```
williazz, line 13: what is the purpose of these port mappings?

SaxyPandaBear: 443 is for HTTPS. So we're restrict outbound access to HTTPS, except for HTTP for communicating with IMDS, and NFS for the EFS mounting.

terraform/setup/vpc.tf
```diff
@@ -10,9 +10,25 @@ resource "aws_security_group" "ec2_security_group" {
...
+    protocol    = "TCP"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  // Allow access to IMDS
+  egress {
+    from_port = 80
+    to_port   = 80
+    protocol = "TCP"
+    cidr_blocks = ["169.254.169.254/32"]
```
williazz, line 24: What information did you use to choose this specific cidr block?

williazz (approved):

PR #137, opened 2023-02-20
Validate Prometheus EMF log schema

SaxyPandaBear:
# Description of changes
1. Refactored CWL validation to be more generic
2. Introduced a JSON schema to validate raw EMF logs against

# License

# Tests
Ran on my fork. https://github.com/SaxyPandaBear/amazon-cloudwatch-agent/actions/runs/4285132843/jobs/7463467268


internal/awsservice/cloudwatchlogs.go
```diff
@@ -134,15 +73,27 @@ func DeleteLogGroup(logGroupName string) {
...
 }
 
-// ValidateLogsInOrder takes a log group, log stream, a list of specific log lines and a timestamp.
+// ValidateLogs takes a log group, log stream, a list of specific log lines and a timestamp.
 // It should query the given log stream for log events, and then confirm that the log lines that are
 // returned match the expected log lines. This also sanitizes the log lines from both the output and
 // the expected lines input to ensure that they don't diverge in JSON representation (" vs ')
-func ValidateLogsInOrder(t *testing.T, logGroup, logStream string, logLines []string, since time.Time) {
-	log.Printf("Checking %s/%s since %s for %d expected logs", logGroup, logStream, since.UTC().Format(time.RFC3339), len(logLines))
+func ValidateLogs(t *testing.T, logGroup, logStream string, since, until *time.Time, validator func(logs []string) bool) {
```
khanhntd, line 80: This would be out of scope for the PR: would it be better to return error instead of utilizing `test` as a helper?

khanhntd (commented):
Another thing to consider would be: do we need to validate the EMF logs from `metric_value_benchmark` with ECS Container Insights/ EMF Structured logs/ Prometheus ? Since Container Insights would use the logs to generate metrics in CI dashboard (not the EMF metrics).

internal/awsservice/cloudwatchlogs.go
SaxyPandaBear: We could, but what is the benefit of that? The purpose of this function is to validate logs from CloudWatch. We don't always need to wrap everything in an error and check it in the main test code. I don't see the issue with this.

SaxyPandaBear:
I'll add validation on the EMF plugin test

internal/awsservice/cloudwatchlogs.go
SaxyPandaBear: I ended up making the change in order to properly reuse it in the emf plugin test

test/metric_value_benchmark/emf_test.go
```diff
@@ -90,7 +100,52 @@ func (t *EMFTestRunner) validateEMFMetric(metricName string) status.TestResult {
...
+			keyErrors, e := rs.ValidateBytes(context.Background(), []byte(l))
+			if e != nil {
+				log.Println("failed to execute schema validator:", e)
+				return false
+			} else if len(keyErrors) > 0 {
+				log.Printf("failed schema validation: %v\n", keyErrors)
+				return false
+			}
+
+			if !strings.Contains(l, "\"EMFCounter\": 5") {
```
khanhntd, line 141:
```suggestion
			if !strings.Contains(l, "\"CloudWatchMetrics\"") {
```
in case of changing EMF format in the future IMO.

test/metric_value_benchmark/agent_resources/emf_counter.json
```diff
@@ -0,0 +1,72 @@
...
+    "EMFCounter",
+    "InstanceId"
+  ],
+  "additionalProperties": true,
+  "definitions": {
+    "emfDefinition": {
+      "type": "object",
+      "additionalProperties": true,
+      "properties": {
+        "Timestamp": {},
```
khanhntd, line 32: Why would this be an empty type though?

test/ecs/ecs_metadata/resources/emf_prometheus_redis_schema.json
```diff
@@ -0,0 +1,53 @@
...
+    "container_name": {},
+    "golang_version": {},
+    "instance": {},
+    "job": {},
+    "prom_metric_type": {},
+    "redis_exporter_build_info": {},
+    "version": {}
+  },
+  "required": [
+    "ClusterName",
```
khanhntd, line 30: Does it need to require all of this value though? Only some is important here `prom_metric_type`, `instance`, `job` and related information of ECS. Other than that, I don't see any important dimensions required by the tests and also by the features.

internal/awsservice/cloudwatchlogs.go
```diff
@@ -134,30 +71,50 @@ func DeleteLogGroup(logGroupName string) {
...
-// ValidateLogsInOrder takes a log group, log stream, a list of specific log lines and a timestamp.
-// It should query the given log stream for log events, and then confirm that the log lines that are
-// returned match the expected log lines. This also sanitizes the log lines from both the output and
-// the expected lines input to ensure that they don't diverge in JSON representation (" vs ')
-func ValidateLogsInOrder(t *testing.T, logGroup, logStream string, logLines []string, since time.Time) {
-	log.Printf("Checking %s/%s since %s for %d expected logs", logGroup, logStream, since.UTC().Format(time.RFC3339), len(logLines))
-	cwlClient, clientContext, err := getCloudWatchLogsClient()
+// ValidateLogs queries a given LogGroup/LogStream combination given the start and end times, and executes an
+// arbitrary validator function on the found logs.
+func ValidateLogs(logGroup, logStream string, since, until *time.Time, validator func(logs []string) bool) (bool, error) {
```
khanhntd, line 76: Would it be better to change to `startTime` and `endTime`  instead of since and until?

khanhntd (commented):
Overall, LGTM. Only some small nit picks. It would be best to add the validate logs in `container insights` [too](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/metric_value_benchmark/container_insights_test.go).

test/metric_value_benchmark/emf_test.go
SaxyPandaBear: No, the point is validating the metric value that is in the log.

test/metric_value_benchmark/agent_resources/emf_counter.json
SaxyPandaBear: I don't think that it makes much difference to validate the exact types here or not. If you feel strongly about it I can change it.

test/ecs/ecs_metadata/resources/emf_prometheus_redis_schema.json
SaxyPandaBear: I took exactly what we emit and am validating that.

internal/awsservice/cloudwatchlogs.go
SaxyPandaBear: StartTime and EndTime are the field attributes in the API call / SDK, but that doesn't mean we have to change it here. If you feel strongly about it, I can change it

internal/awsservice/cloudwatchlogs.go
```diff
@@ -134,30 +72,50 @@ func DeleteLogGroup(logGroupName string) {
...
+	log.Printf("Checking %s/%s\n", logGroup, logStream)
+
```
ChenaLee, line 79: I know this is an existing code, but it makes no sense why validation is in this package. This package seems like a SAO.

SaxyPandaBear: I wrote this code originally with the intent of obfuscating the SDK client and just having validation functions that implicitly used them. I wanted to simplify the testing so that it was easy to write a new test that uses a validation function which makes calls to AWS, without getting into the business of passing around SDK clients. 

Obviously, things got shifted around and that doesn't make sense anymore. I can rip it out now, or can do it in another PR. Up to you

internal/awsservice/cloudwatchlogs.go
```diff
@@ -134,30 +72,50 @@ func DeleteLogGroup(logGroupName string) {
...
 		LogStreamName: aws.String(logStream),
-		StartTime:     aws.Int64(sinceMs),
 		StartFromHead: aws.Bool(true), // read from the beginning
 	}
 
-	foundLogs := make([]string, 0)
+	var sinceMs int64
+	if since != nil {
+		sinceMs = since.UnixNano() / 1e6 // convert to millisecond timestamp
+		params.StartTime = aws.Int64(sinceMs)
```
khanhntd, line 110: Would be better to pass startTime and EndTime as requirement instead of using `sinceMs` and `untilMs` since there will be rarely a case there are logs within the same time. (force better than lenient here)

internal/awsservice/cloudwatchlogs.go
```diff
@@ -194,22 +152,15 @@ func ValidateLogsInOrder(t *testing.T, logGroup, logStream string, logLines []st
...
-	}
+	return foundLogs, nil
 }
 
-// isLogGroupExists confirms whether the logGroupName exists or not
-func IsLogGroupExists(t *testing.T, logGroupName string) bool {
-
+// IsLogGroupExists confirms whether the logGroupName exists or not
+func IsLogGroupExists(logGroupName string) bool {
 	cwlClient, clientContext, err := getCloudWatchLogsClient()
```
khanhntd, line 160: You can call the [cwlClient directly](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/internal/awsservice/constant.go#L38-L46) without using `getCloudWatchLogsClient`

test/metric_value_benchmark/container_insights_test.go
```diff
@@ -6,26 +6,37 @@
...
 	test_runner.BaseTestRunner
 }
 
+//go:embed agent_resources/container_insights_node_telemetry.json
+var emfContainerInsightsSchema string
+
 var _ IECSTestRunner = (*ContainerInsightsTestRunner)(nil)
 
-func (t *ContainerInsightsTestRunner) validate() status.TestGroupResult {
+func (t *ContainerInsightsTestRunner) validate(e *environment.MetaData) status.TestGroupResult {
```
khanhntd, line 31: Seems it quite become more complex than I expect. But its not worth to solve it right away.

internal/awsservice/cloudwatchlogs.go
```diff
@@ -6,16 +6,14 @@ package awsservice
 import (
 	"context"
 	"errors"
+	"github.com/qri-io/jsonschema"
```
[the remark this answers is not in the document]
SaxyPandaBear: Why? It's in the go mod

internal/awsservice/cloudwatchlogs.go
SaxyPandaBear: What does this mean? What do you mean rarely a case?

test/metric_value_benchmark/container_insights_test.go
SaxyPandaBear: What is complex about this? I am passing in the environment so I can get the cluster name and ARN

internal/awsservice/cloudwatchlogs.go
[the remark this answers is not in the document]
khanhntd: [Goimport](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) for formatting and grouping the similar dependencies together. VSCodes has `go-group-imports` that [groups by default](https://marketplace.visualstudio.com/items?itemName=aleksandra.go-group-imports) with 
```
Standard library imports,
Third party imports,
Own (belonging to the organization) imports.
```

test/metric_value_benchmark/emf_test.go
```diff
@@ -90,7 +100,47 @@ func (t *EMFTestRunner) validateEMFMetric(metricName string) status.TestResult {
...
+	rs := jsonschema.Must(emfMetricValueBenchmarkSchema)
+
+	validateLogContents := func(s string) bool {
+		return strings.Contains(s, "\"EMFCounter\":5")
+	}
+
+	now := time.Now()
+	ok, err := awsservice.ValidateLogs(group, stream, nil, &now, func(logs []string) bool {
+		if len(logs) < 1 {
+			return false
```
khanhntd, line 132: Would it be better to return error here for the validator? Since its not clear the reason its failed between when its does not have logs and when its does not match the emf?

SaxyPandaBear: No, I am not changing this to return an error.

internal/awsservice/cloudwatchlogs.go
SaxyPandaBear: There are scenarios where we don't know the start time or the end time. So no.

internal/awsservice/cloudwatchlogs.go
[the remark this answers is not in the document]
SaxyPandaBear: Is this part of our linter?

test/ecs/ecs_metadata/ecs_metadata_test.go
```diff
@@ -20,28 +24,63 @@ import (
...
+			}
+			return true
+		})
+		assert.NoError(t, err)
+		assert.True(t, ok)
+
+		break
+	}
+
+	if logGroupFound {
```
khanhntd, line 83: This would always be false so `DeleteLogGroupAndStream` won't be triggered?

internal/awsservice/cloudwatchlogs.go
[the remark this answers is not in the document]
SaxyPandaBear: The linter doesn't run because the Makefile is invalid. I fixed that and there are lots of linter errors. I'm not fixing those in this PR because it's already a month old. I moved the import manually.

khanhntd (approved):
One small nit pick. Other than that, LGTM.

khanhntd (approved):

internal/awsservice/cloudwatchlogs.go
```diff
@@ -107,26 +52,45 @@ func DeleteLogGroup(logGroupName string) {
...
 		LogGroupName:  aws.String(logGroup),
 		LogStreamName: aws.String(logStream),
-		StartTime:     aws.Int64(sinceMs),
 		StartFromHead: aws.Bool(true), // read from the beginning
 	}
 
-	foundLogs := make([]string, 0)
+	var sinceMs int64
+	if since != nil {
+		sinceMs = since.UnixNano() / 1e6 // convert to millisecond timestamp
```
sky333999, line 84: I see this is how it was even before your changes, but wonder if there is a specific reason to do it this way instead of just using `since.UnixMilli()`.

test/ecs/ecs_metadata/resources/emf_prometheus_redis_schema.json
```diff
@@ -0,0 +1,75 @@
...
+    "TaskClusterName",
+    "TaskDefinitionFamily",
+    "TaskGroup",
+    "TaskId",
+    "TaskRevision",
+    "Timestamp",
+    "Version",
+    "container_name",
+    "job",
+    "prom_metric_type"
```
sky333999, line 72: This property wont exist in public CWA so Id imagine this test fails there?

SaxyPandaBear: The test passes 🤷

SaxyPandaBear: Also, to clarify, I grabbed the schema directly from what got published to CloudWatch Logs from my forked integration tests.

sky333999 (approved):

test/ecs/ecs_metadata/resources/emf_prometheus_redis_schema.json
sky333999: Yeah my bad.. this does exist even in CWA.

PR #138, opened 2023-02-20
Add A Terraform Linter

sethAmazon:
# Description of the issue
Terraform is not formatted 

# Description of changes
Add a terraform linter. 

# License

# Tests
Actions run in gh

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -22,6 +22,9 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ~1.19.2
+      - name: Lint Terraform
+        if: runner.os == 'Linux'
+        run: cd terraform && make check-fmt
       - name: Compile tests
         run: |
           echo "Compile tests"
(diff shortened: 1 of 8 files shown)


.github/workflows/build-check.yml
```diff
@@ -22,6 +22,9 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ~1.19.2
+      - name: Lint Terraform
```
khanhntd, line 25: The action only uses [terraform fmt -check -recursive -diff](https://github.com/actionshub/terraform-lint/blob/main/entrypoint.sh#L1) but we already [had that](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/terraform/Makefile#L4-L5) so its better to not use third party and implement ourselves IMO.

sethAmazon: Changed to use bash directly instead of action.

.github/workflows/build-check.yml
```diff
@@ -22,6 +22,13 @@ jobs:
...
+        if: runner.os == 'Linux'
+        run: |
+          if [[ -n "$(terraform fmt -check -recursive -diff)" ]]; then
```
khanhntd, line 28:
```suggestion
          cd terraform && make check-fmt
```

khanhntd (approved):

taohungyang (approved):

PR #143, opened 2023-02-20
Add performance test to the validator

khanhntd:
# Description of the issue
Add the performance to the validator options and add use case `logs` to performance validator. To be more specific regarding the current situation, the [performance test will generate the logs](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/performancetest/performance_test.go#L54) dynamically according to the number of logs; afterwards, [instruct the agent to monitor the logs,](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/performancetest/performance_test.go#L92) [get all the performance metrics ](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/performancetest/performance_test.go#L101)and [sending it to the database](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/performancetest/performance_test.go#L114)

# Description of changes
* Add [use case logs for validator ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/143/files#diff-f661d2e9882fafba531d27c2151a8f4b80ed0d69b5a34f4ff2ad6c53d2bf9ee8)
* Add [performance validator](https://github.com/aws/amazon-cloudwatch-agent-test/pull/143/files#diff-f29ffa8f3f7fe6348e48c4e51184d29ffc6737dd00398b5f3bb99121c0820753), similar to how [performance testing](https://github.com/aws/amazon-cloudwatch-agent-test/tree/main/test/performancetest) works

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
_Describe what tests you have done._

PR #144, opened 2023-02-20
Generate SDK Config beforehand

khanhntd:
# Description of the issue
Generate SDK Config instead of looking up every time and delete the performance test since it would be added as a part of [validator](https://github.com/aws/amazon-cloudwatch-agent-test/pull/143)

# License

# Tests
By running the changes on my fork, here is the result of it: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4036724893


internal/awsservice/dynamodb.go
```diff
@@ -0,0 +1,61 @@
...
+package awsservice
+
+import (
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+
+func AddItemIntoDatabase(databaseName, conditionExpression string, packet interface{}, expressAttributesNames map[string]string) error {
```
khanhntd, line 13: The SDK is added based on the[ performance validator.](https://github.com/aws/amazon-cloudwatch-agent-test/pull/143/files#diff-06922a0a54f2d4ec5d262269aa687bb154b542cde78bc15b5af79c9c5e5e8458R16)

internal/awsservice/cloudwatchlogs.go
```diff
@@ -49,20 +39,18 @@ func ValidateLogs(t *testing.T, logGroup, logStream string, numExpectedLogs int,
...
 	attempts := 0
-	var output *cloudwatchlogs.GetLogEventsOutput
 	var nextToken *string
 
 	for {
 		if nextToken != nil {
 			params.NextToken = nextToken
 		}
-		output, err = cwlClient.GetLogEvents(*clientContext, params)
+		output, err := CwlClient.GetLogEvents(cxt, params)
```
adam-mateen, line 48: should `cxt` be renamed `ctx`, just for convention's sake?

internal/awsservice/ecs.go
```diff
@@ -90,41 +78,19 @@ func GetClusterName(clusterArn string) string {
 }
 
 func listContainerInstances(clusterArn string) (*ecs.ListContainerInstancesOutput, error) {
-	svc, ctx, err := getEcsClient()
-	if err != nil {
-		return nil, err
-	}
 
 	input := &ecs.ListContainerInstancesInput{
```
jefchien, line 82: I'm not really seeing the benefit here. Why not just use the client directly? How much more difficult is it for someone to write:
```go
client.ListContainerInstances(context.Background(), &ecs.ListContainerInstancesInput{Cluster: aws.String(clusterArn)})
```

khanhntd: I'm fine with write it too. Try to keep the code small as much as possible though.

internal/awsservice/cloudwatchlogs.go
khanhntd: I'm good with that too ! Thanks for the suggestion.

internal/awsservice/ecs.go
jefchien: I remember discussing something similar in a previous PR. I don't think it's necessary to have these functions if they're not either transforming the input or triggering some other change. We don't need to have this extra layer.

khanhntd: The previous PR has a broader change regarding using the `interface` with the SDK. However, the extra layer you mentioned are not in scope of the PR. But I'm fine with changing it.

internal/awsservice/dynamodb.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"errors"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+
+func ReplaceItemInDatabase(databaseName string, packet map[string]interface{}) error {
```
khanhntd, line 15: The SDK is added based on the[ performance validator.](https://github.com/aws/amazon-cloudwatch-agent-test/pull/143/files#diff-06922a0a54f2d4ec5d262269aa687bb154b542cde78bc15b5af79c9c5e5e8458R16)

SaxyPandaBear (approved):

sethAmazon (approved):

PR #151, opened 2023-02-28
Add performance validator to validator

khanhntd:
# Description of the issue
To start the validator, there are three stages:
* [GenerateLoad](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L10):  Sending the load to CloudWatchAgent
* [CheckData](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L11) : To get metrics defined [by the yaml and validate the metrics](https://github.com/aws/amazon-cloudwatch-agent-test/compare/main...khanhntd:amazon-cloudwatch-agent-test:add_performance?expand=1#diff-4182821602cec8db84c8b29427a65c5a63beff4c19bcf93e305966bed05a721eR20-R72)
* [Cleanup](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L12) to clean up all the resources created by the validator.

However, as a plan goal, we want to add performance test to know the performance of CWA for each commit. Therefore,
onboarding performance test to the validator. As of right now, the performance test only [generates load ](https://github.com/aws/amazon-cloudwatch-agent-test/compare/main...khanhntd:amazon-cloudwatch-agent-test:add_performance?expand=1#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R26-R42). Other stage will in the upcoming PR.

The reason why I used the generator would be:
* The go test cannot leverage all the environment details when coming to ECS/EKS Fargate or EC2 (container environment) if deploy as a service
* Increase the cognitive load to add test (needs to know the agent run cycle and also all the API calls) which reduce abstraction
* Increase the scope of the test since there are times we load multiple config into a single EC2 instances

# License

# Tests
I have run it with my fork: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4297854336
```
null_resource.integration_test: Still creating... [13m50s elapsed]
null_resource.integration_test: Still creating... [14m0s elapsed]
null_resource.integration_test (remote-exec): 2023/02/28 23:17:00 Start to sleep 60s for CloudWatch to process all the metrics
null_resource.integration_test: Still creating... [14m10s elapsed]
null_resource.integration_test: Still creating... [14m20s elapsed]
null_resource.integration_test: Still creating... [14m30s elapsed]
null_resource.integration_test: Still creating... [14m40s elapsed]
null_resource.integration_test: Still creating... [14m50s elapsed]
null_resource.integration_test: Still creating... [15m0s elapsed]
null_resource.integration_test (remote-exec): 2023/02/28 23:18:00 Finish validation in 11.377706357066666 minutes
null_resource.integration_test: Creation complete after 15m7s [id=4183436441955450755]
```

Here is a graph for it: performance metrics are getting uploaded to CloudWatch
```
![image](https://user-images.githubusercontent.com/91758108/222241317-a5e3201a-1050-459f-96b0-52d971042185.png)

```


generator/resources/ec2_performance_test_matrix.json
```diff
@@ -4,7 +4,6 @@
         "username": "ec2-user",
         "installAgentCommand": "go run ./install/install_agent.go rpm",
         "ami": "cloudwatch-agent-integration-test-al2*",
-        "caCertPath": "/etc/ssl/certs/ca-bundle.crt",
```
khanhntd, line 7: We are delete `caCertPath` since performance/stress does not require it as a input. Here is all the required input for the [terraform performance ](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/terraform/performance/variables.tf#L1-L84) and how are [things get generated by the generator](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/generator/test_case_generator.go#L16-L29)

terraform/performance/main.tf
```diff
@@ -111,7 +111,7 @@ resource "null_resource" "integration_test" {
       "export PATH=$PATH:/snap/bin:/usr/local/go/bin",
       "echo run integration test",
       "cd ~/amazon-cloudwatch-agent-test",
-      "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s -c file:/${local.instance_temp_directory}/${local.cloudwatch_agent_config}",
+      "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s -c file:${local.instance_temp_directory}/${local.cloudwatch_agent_config}",
```
khanhntd, line 114: `${local.instance_temp_directory}` is already has `/` since the value for that variable is `/tmp`. Therefore, removing it to avoid conflict.

validator/validators/validator.go
```diff
@@ -9,11 +9,14 @@ import (
 	"time"
 
 	"github.com/aws/amazon-cloudwatch-agent-test/validator/models"
+	"github.com/aws/amazon-cloudwatch-agent-test/validator/validators/performance"
 	"github.com/aws/amazon-cloudwatch-agent-test/validator/validators/stress"
 )
 
 func NewValidator(vConfig models.ValidateConfig) (validator models.ValidatorFactory, err error) {
 	switch vConfig.GetValidateType() {
+	case "performance":
```
khanhntd, line 18: The validate type are being passed as [part of the yaml](https://github.com/aws/amazon-cloudwatch-agent-test/pull/151/files#diff-4182821602cec8db84c8b29427a65c5a63beff4c19bcf93e305966bed05a721eR4). In this case, its performance.

validator/models/validation_config.go
```diff
@@ -53,6 +57,8 @@ type MetricDimension struct {
 	Value string `yaml:"value"`
 }
 
+var _ ValidateConfig = (*validatorConfig)(nil)
```
khanhntd, line 60: This will validate if the `validatorConfig` has all the declared function by the [interface](https://stackoverflow.com/a/27804417)

validator/validators/performance/performance_validator.go
```diff
@@ -0,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
khanhntd, line 1: Here is how the performance declare the validator through three stages at declared in the PR:
* [GenerateLoad](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L10): 
* [CheckData](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L11) 
* [Cleanup](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L12)

However, as stated in the PR, we will only generate load in this PR with [metrics and logs only](https://github.com/aws/amazon-cloudwatch-agent-test/pull/151/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R35-R39). The get and push validation will be in the upcoming PR.

test/performance/logs/parameters.yml
```diff
@@ -0,0 +1,72 @@
+receivers: "logs"
```
khanhntd, line 1: The `parameters.yml` is the one who has all the information and instruct the validator to generate the logs (e.g receivers specify we would generate logs and the number of monitor logs would be 1000 - `number_monitored_logs`). It also includes all the metrics the [performance wants to capture](https://github.com/aws/amazon-cloudwatch-agent-test/pull/151/files#diff-4182821602cec8db84c8b29427a65c5a63beff4c19bcf93e305966bed05a721eR20-R72)

terraform/setup/main.tf
```diff
@@ -11,27 +11,50 @@ resource "aws_dynamodb_table" "performance-dynamodb-table" {
...
   }
 
   attribute {
-    name = "CommitDate"
-    type = "N"
+    name = "UseCase"
+    type = "S"
+  }
+
+  global_secondary_index {
```
khanhntd, line 37: `global_secondary_index` is a second queries that we can query the database besides the primary key. For more information on it, please [following this document ](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html)

sethAmazon (approved):

test/performance/logs/agent_config.json
```diff
@@ -0,0 +1,59 @@
...
+				],
+				"measurement": [
+				  "bytes_sent",
+				  "packets_sent"
+				],
+				"metrics_collection_interval": 1
+			},
+			"procstat": [
+				{
+				  "exe": "cloudwatch-agent",
```
sethAmazon, line 30: Nit: We can change to pid file in case there is something else on the system running with a similar name. 
Idc if this is changed if you disagree

test/performance/logs/parameters.yml
```diff
@@ -0,0 +1,72 @@
...
+
+test_case: "logs_performance"
```
sethAmazon, line 3: I wonder if there is a way to have code parameters instead of using yaml. It is fine if you think is cleaner.

terraform/performance/main.tf
taohungyang: Agreed. So this never succeeded before, right?

internal/awsservice/dynamodb.go
```diff
@@ -68,7 +68,7 @@ func GetItemInDatabase(databaseName, indexName string, checkingAttribute, checki
 			"#second_attribute": checkingAttribute[1],
 		},
 		ExpressionAttributeValues: map[string]types.AttributeValue{
-			":first_attribute":  &types.AttributeValueMemberN{Value: checkingAttributeValue[0]},
+			":first_attribute":  &types.AttributeValueMemberS{Value: checkingAttributeValue[0]},
```
taohungyang, line 71: Just to confirm if this is because of the type change in attribute in `setup/main.tf`

validator/models/validation_config.go
```diff
@@ -41,6 +42,9 @@ type validatorConfig struct {
 
 	MetricNamespace  string             `yaml:"metric_namespace"`
 	MetricValidation []MetricValidation `yaml:"metric_validation"`
+
+	CommitHash string `yaml:"commit_hash"`
+	CommitDate string `yaml:"commit_date"`
```
taohungyang, line 47: Do we need to add these two fields to `test/stress/statsd/parameters.yml`?

taohungyang (commented):
Nice adding comments/descriptions. That helps me understand this faster.

terraform/performance/main.tf
khanhntd: No it has not though !!!

validator/models/validation_config.go
khanhntd: That's the fun thing regarding `unmarshall` . If you are nut suppling the variable (e.g `CommitHash`), it will give the null or "" depends on the yaml (e.g `mapstructure` return null while `yaml` return empty string).

internal/awsservice/dynamodb.go
khanhntd: Yes that's correct. Since the S is for String and N is for Number and for upcoming PR, I will search based on CommitHash and CWA Use Case which both are String.

validator/models/validation_config.go
taohungyang: Yeah... agreed with the default value but will this cause an issue if `GetCommitInformation()` is called and `v.CommitDate` is null or ""
(Since it's calling `commitDate, _ := strconv.ParseInt(v.CommitDate, 10, 64)`)

khanhntd: So there is no use case of integration test or stress test using `GetCommitInformation` and it seems to be overengineering in adding it.

taohungyang: okk.. thanks for explaining.

taohungyang (approved):

validator/validators/performance/performance_validator.go
adam-mateen: A performance benchmark doesn't really need to check metric values anyways. In my opinion it just generates load and measure the Agent's resource usage (CPU, MEM, DISK, NETIO).

generator/resources/ec2_stress_test_matrix.json
```diff
@@ -37,7 +34,6 @@
         "username": "ec2-user",
         "installAgentCommand": "go run ./install/install_agent.go rpm",
         "ami": "cloudwatch-agent-integration-test-al2*",
-        "caCertPath": "/etc/ssl/certs/ca-bundle.crt",
```
adam-mateen, line 40: I still see `caCertPath` in `type matrixRow struct`. Could it be removed? Should it?

terraform/performance/main.tf
adam-mateen: On linux a double forward slash works fine (at least in bash):
`ls /tmp/` same result as `ls //tmp//`.

Regardless, it makes sense to remove unnecessary slashes.

generator/resources/ec2_stress_test_matrix.json
taohungyang: I see some other tests still has `caCertPath` in the .json so I don't think we should remove it.
e.g.:
- https://github.com/aws/amazon-cloudwatch-agent-test/blob/e07fe7adb1b1d75244d8984507d3f83a7237c3d3/generator/resources/ec2_linux_test_matrix.json#L8
- https://github.com/aws/amazon-cloudwatch-agent-test/blob/e07fe7adb1b1d75244d8984507d3f83a7237c3d3/generator/resources/ec2_gpu_test_matrix.json#L8
-

validator/validators/performance/performance_validator.go
```diff
@@ -0,0 +1,59 @@
...
+	var (
+		agentCollectionPeriod = s.vConfig.GetAgentCollectionPeriod()
+		agentConfigFilePath   = s.vConfig.GetCloudWatchAgentConfigPath()
+		dataType              = s.vConfig.GetDataType()
+		dataRate              = s.vConfig.GetDataRate()
+		receiver              = s.vConfig.GetPluginsConfig()
+	)
+	switch dataType {
+	case "logs":
+		err = common.StartLogWrite(agentConfigFilePath, agentCollectionPeriod, dataRate)
```
adam-mateen, line 36: With a name like "StartLogWrite", I assume it means it JUST starts the generator, but does not WAIT for it.
In other words I expect it to start generating in goroutines and immediately return.

Is this the case? Or does this call block?

generator/resources/ec2_stress_test_matrix.json
khanhntd: It been used for local stack to generate CA Bundle for the [CA Bundle test when there is region](https://github.com/aws/amazon-cloudwatch-agent-test/tree/main/test/ca_bundle) which can not use the auto CA bundle. However, for your answer, we cannot remove it.

validator/validators/performance/performance_validator.go
khanhntd: The name are confusing. However, `StartLogWrite` will wait for the generator to completely [generate the logs with X minutes](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/internal/common/logs.go#L18-L43)

validator/validators/performance/performance_validator.go
khanhntd: Yes that's the point of performance. It does not need to check the statsd's metrics value but only monitor the agent performance.

PR #153, opened 2023-03-02
Send metrics to database after query performance metrics from CWA

khanhntd:
# Description of the issue
Continue as part of onboarding metrics load to [performance validator test](https://github.com/aws/amazon-cloudwatch-agent-test/pull/151). As discussed in the previous PR,  there are three stages when starting the validator:

* [GenerateLoad](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L10): Sending the load to CloudWatchAgent
* [CheckData](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L11) : To get metrics defined [by the yaml and validate the metrics](https://github.com/aws/amazon-cloudwatch-agent-test/compare/main...khanhntd:amazon-cloudwatch-agent-test:add_performance?expand=1#diff-4182821602cec8db84c8b29427a65c5a63beff4c19bcf93e305966bed05a721eR20-R72)
* [Cleanup](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L12) to clean up all the resources created by the validator.

As part of previous PR, we already have `GenerateLoad` and `CleanUp`. Therefore, we are  adding the remaining part with `CheckData` since the whole operations inside the CheckData for performance test would be:
* [Get the performance metrics of CWA](https://github.com/aws/amazon-cloudwatch-agent-test/pull/153/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R162-R193) after [generate loads to CWA ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/153/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R45-R60)based on the required metrics by the generator configuration: `parameters.yml` (e.g logs)
*[ Calculate the performance metrics and generate metrics with statistical value](https://github.com/aws/amazon-cloudwatch-agent-test/pull/153/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R151-R156) (e.g max, min, p99, stddev) - porting from intern project
* Send the [final performance metrics of CWA  with statistical values to database ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/153/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R94-R129)

# Description of changes
Sending the performance metrics to Database will help us in recording the performance metrics of CWA during each commit and will reflect in the [FrontEnd Page](https://aws.github.io/amazon-cloudwatch-agent/)

# License

# Tests
By running the fork, I have able to validate [the performance test with logs successfully](https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4310619379)

We can see the performance metrics results with before calculate statistics
```
null_resource.integration_test (remote-exec): 2023/03/02 06:02:03 Start calculate metric statictics for metric net_packets_sent [2.1 2.7 5.2 2.8 1107 1939.4 2.2 2.4 5.6 2.5 999.1 1974.7 2.4 2.6 5.3 2.7 952.4 1793.5 2.5 2.5 5.2 2.8 1140.7 2129.8 2.2 2.9 5.8 2.4 993.7 1788.2 2.1 2.9 5.3 2.9 1011.2 2019.7 2 2.7 5.5 2.8 1036.2 2129.3 2.1 2.5 7.1 4 1146.4 1955.6 2.8 101.3 1674.9 6445.9 10610.5 3926.1 2.4 4.1 5.2 2.6 513.5 998.5]
```

and after calculating it,
```
Finished calculate metric statictics for metric net_packets_sent: {440.21999999999997 1905 7416.3 2 10 1110.0022529706866}
```
which will have the following results in the database
```
{"100":{"M":{"net_bytes_sent":{"M":{"Average":{"N":"101782.97666666667"},"P99":{"N":"444492.2"},"Period":{"N":"10"},"Std":{"N":"262448.01973010856"},"Min":{"N":"204"},"Max":{"N":"1754209.5"}}},"procstat_memory_rss":{"M":{"Average":{"N":"583.3093119999997"},"P99":{"N":"664.543232"},"Period":{"N":"10"},"Std":{"N":"110.14129714232358"},"Min":{"N":"142.7677184"},"Max":{"N":"664.543232"}}},"procstat_memory_data":{"M":{"Average":{"N":"892.3917516799997"},"P99":{"N":"934.526976"},"Period":{"N":"10"},"Std":{"N":"121.28089734791472"},"Min":{"N":"364.0348672"},"Max":{"N":"934.526976"}}},"procstat_write_bytes":{"M":{"Average":{"N":"19.91503872"},"P99":{"N":"40.026112"},"Period":{"N":"10"},"Std":{"N":"12.497415181559536"},"Min":{"N":"0.606208"},"Max":{"N":"40.026112"}}},"procstat_num_fds":{"M":{"Average":{"N":"1018.4249999999996"},"Period":{"N":"10"},"P99":{"N":"1100"},"Std":{"N":"42.473586399863485"},"Min":{"N":"817.9"},"Max":{"N":"1182.9"}}},"procstat_memory_vms":{"M":{"Average":{"N":"1621.6943274666667"},"Period":{"N":"10"},"P99":{"N":"1663.004672"},"Std":{"N":"119.11964320756728"},"Min":{"N":"1103.4177536"},"Max":{"N":"1663.004672"}}},"net_packets_sent":{"M":{"Average":{"N":"440.21999999999997"},"Period":{"N":"10"},"P99":{"N":"1905"},"Std":{"N":"1110.0022529706866"},"Min":{"N":"2"},"Max":{"N":"7416.3"}}},"procstat_memory_swap":{"M":{"Average":{"N":"0"},"P99":{"N":"0"},"Period":{"N":"10"},"Std":{"N":"0"},"Min":{"N":"0"},"Max":{"N":"0"}}},"procstat_cpu_usage":{"M":{"Average":{"N":"38.337993510195396"},"P99":{"N":"70.25992208910591"},"Period":{"N":"10"},"Std":{"N":"18.821200930328175"},"Min":{"N":"26.799011825037034"},"Max":{"N":"143.73881954361354"}}},"mem_total":{"M":{"Average":{"N":"16629.624832"},"Period":{"N":"10"},"P99":{"N":"16629.624832"},"Std":{"N":"0"},"Min":{"N":"16629.624832"},"Max":{"N":"16629.624832"}}}}},"1000":{"M":{"net_bytes_sent":{"M":{"Average":{"N":"262898.7233333333"},"P99":{"N":"1777264.6"},"Period":{"N":"10"},"Std":{"N":"497871.2742885174"},"Min":{"N":"204"},"Max":{"N":"2745740.6"}}},"procstat_memory_rss":{"M":{"Average":{"N":"1011.03736832"},"P99":{"N":"1415.094272"},"Period":{"N":"10"},"Std":{"N":"318.7403520065071"},"Min":{"N":"155.6729856"},"Max":{"N":"1451.515904"}}},"procstat_memory_data":{"M":{"Average":{"N":"1518.9065864533325"},"Period":{"N":"10"},"P99":{"N":"1677.258752"},"Std":{"N":"344.9219847555371"},"Min":{"N":"408.28887039999995"},"Max":{"N":"1677.258752"}}},"procstat_write_bytes":{"M":{"Average":{"N":"35.32088661333332"},"Period":{"N":"10"},"P99":{"N":"71.811072"},"Std":{"N":"22.709789979639456"},"Min":{"N":"0.6168576"},"Max":{"N":"71.811072"}}},"procstat_num_fds":{"M":{"Average":{"N":"1022.5049999999997"},"Period":{"N":"10"},"P99":{"N":"1106.1"},"Std":{"N":"40.53155324353279"},"Min":{"N":"918.1"},"Max":{"N":"1234.9"}}},"procstat_memory_vms":{"M":{"Average":{"N":"2272.375343786666"},"Period":{"N":"10"},"P99":{"N":"2435.096576"},"Std":{"N":"346.1554123972458"},"Min":{"N":"1145.1551744"},"Max":{"N":"2435.096576"}}},"net_packets_sent":{"M":{"Average":{"N":"808.5233333333332"},"P99":{"N":"6445.9"},"Period":{"N":"10"},"Std":{"N":"1704.3037622801346"},"Min":{"N":"2"},"Max":{"N":"10610.5"}}},"procstat_memory_swap":{"M":{"Average":{"N":"0"},"Period":{"N":"10"},"P99":{"N":"0"},"Std":{"N":"0"},"Min":{"N":"0"},"Max":{"N":"0"}}},"procstat_cpu_usage":{"M":{"Average":{"N":"74.2988991248597"},"P99":{"N":"260.1542794940978"},"Period":{"N":"10"},"Std":{"N":"76.76523848282207"},"Min":{"N":"30.397399469761876"},"Max":{"N":"276.58494463653733"}}},"mem_total":{"M":{"Average":{"N":"16629.633023999988"},"P99":{"N":"16629.633024"},"Period":{"N":"10"},"Std":{"N":"0.000000000010913936421275139"},"Min":{"N":"16629.633024"},"Max":{"N":"16629.633024"}}}}},"5000":{"M":{"net_bytes_sent":{"M":{"Average":{"N":"721695.5458796296"},"P99":{"N":"2645431.7"},"Period":{"N":"10"},"Std":{"N":"786688.2384752069"},"Min":{"N":"204"},"Max":{"N":"2887132.4"}}},"procstat_memory_rss":{"M":{"Average":{"N":"1842.593798516364"},"P99":{"N":"2544.0067584000003"},"Period":{"N":"10"},"Std":{"N":"632.5301123247696"},"Min":{"N":"156.5376512"},"Max":{"N":"2718.820352"}}},"procstat_memory_data":{"M":{"Average":{"N":"2489.3376261688873"},"P99":{"N":"2948.247552"},"Period":{"N":"10"},"Std":{"N":"744.5015757734853"},"Min":{"N":"393.41096960000004"},"Max":{"N":"2948.247552"}}},"procstat_write_bytes":{"M":{"Average":{"N":"73.25300389494949"},"Period":{"N":"10"},"P99":{"N":"163.950592"},"Std":{"N":"52.61340434165167"},"Min":{"N":"0.6180864"},"Max":{"N":"164.708352"}}},"procstat_num_fds":{"M":{"Average":{"N":"1112.151022727273"},"Period":{"N":"10"},"P99":{"N":"1370.75"},"Std":{"N":"127.7780998979802"},"Min":{"N":"873.7"},"Max":{"N":"1471.5454545454545"}}},"procstat_memory_vms":{"M":{"Average":{"N":"3227.111095561484"},"P99":{"N":"3701.886976"},"Period":{"N":"10"},"Std":{"N":"746.683073865154"},"Min":{"N":"1124.405248"},"Max":{"N":"3701.886976"}}},"net_packets_sent":{"M":{"Average":{"N":"1792.2091919191919"},"P99":{"N":"9117"},"Period":{"N":"10"},"Std":{"N":"2150.6997429231947"},"Min":{"N":"2"},"Max":{"N":"11005.7"}}},"procstat_memory_swap":{"M":{"Average":{"N":"0"},"Period":{"N":"10"},"P99":{"N":"0"},"Std":{"N":"0"},"Min":{"N":"0"},"Max":{"N":"0"}}},"procstat_cpu_usage":{"M":{"Average":{"N":"185.19243401744816"},"Period":{"N":"10"},"P99":{"N":"398.0546356201349"},"Std":{"N":"136.71815787944305"},"Min":{"N":"24.401007856901565"},"Max":{"N":"398.23272558931444"}}},"mem_total":{"M":{"Average":{"N":"16629.633023999988"},"P99":{"N":"16629.633024"},"Period":{"N":"10"},"Std":{"N":"0.000000000010913936421275139"},"Min":{"N":"16629.633024"},"Max":{"N":"16629.633024"}}}}}}
```

image

### DIFF
--- a/generator/resources/ec2_performance_test_matrix.json
+++ b/generator/resources/ec2_performance_test_matrix.json
@@ -6,16 +6,6 @@
         "ami": "cloudwatch-agent-integration-test-al2*",
         "arc": "amd64",
         "binaryName": "amazon-cloudwatch-agent.rpm",
-        "valuesPerMinute": 10,
-        "family": "linux"
-      },
-      {
-        "os": "al2",
-        "username": "ec2-user",
-        "installAgentCommand": "go run ./install/install_agent.go rpm",
-        "ami": "cloudwatch-agent-integration-test-al2*",
-        "arc": "amd64",
-        "binaryName": "amazon-cloudwatch-agent.rpm",
         "valuesPerMinute": 100,
         "family": "linux"
       },
@@ -27,6 +17,16 @@
         "arc": "amd64",
         "binaryName": "amazon-cloudwatch-agent.rpm",
         "valuesPerMinute": 1000,
+        "family": "linux"
+      },
+      {
+        "os": "al2",
+        "username": "ec2-user",
+        "installAgentCommand": "go run ./install/install_agent.go rpm",
+        "ami": "cloudwatch-agent-integration-test-al2*",
+        "arc": "amd64",
+        "binaryName": "amazon-cloudwatch-agent.rpm",
+        "valuesPerMinute": 5000,
         "family": "linux"
       }
 ]
--- a/internal/awsservice/dynamodb.go
+++ b/internal/awsservice/dynamodb.go
@@ -45,7 +45,7 @@ func AddItemIntoDatabaseIfNotExist(databaseName string, checkingAttribute, check
 				"#second_attribute": checkingAttribute[1],
 			},
 			ExpressionAttributeValues: map[string]types.AttributeValue{
-				":first_attribute":  &types.AttributeValueMemberN{Value: checkingAttributeValue[0]},
+				":first_attribute":  &types.AttributeValueMemberS{Value: checkingAttributeValue[0]},
 				":second_attribute": &types.AttributeValueMemberS{Value: checkingAttributeValue[1]},
 			},
 		})
--- a/test/performance/logs/parameters.yml
+++ b/test/performance/logs/parameters.yml
@@ -4,7 +4,7 @@ test_case: "logs_performance"
 validate_type: "performance"
 data_type: "logs"
 # Number of logs being written
-number_monitored_logs: 1000
+number_monitored_logs: 100
 # Number of metrics to be sent or number of log lines being written  each minute
 values_per_minute: "<values_per_minute>"
 # Number of seconds the agent should run and collect the metrics. In this case, 10 minutes
--- a/validator/validators/performance/performance_models.go
+++ b/validator/validators/performance/performance_models.go
@@ -1,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package performance
+
+type PerformanceInformation map[string]interface{}
+
+/*
+	Contains the following:
+		// A service name we want to monitor (e.g CloudWatchAgent)
+		"Service":          ServiceName,
+		// A use case for generate metrics loads (e.g statsd, collectd)
+		"UseCase":          receiver,
+		// Commit Information
+		"CommitDate":       commitDate,
+		"CommitHash":       commitHash,
+		// Data type (e.g metrics/traces/logs)
+		"DataType":         dataType,
+		// Performance metrics of the monitored services.
+		"Results":          result,
+		// The duration of time when running the services
+		"CollectionPeriod": collectionPeriod,
+		"InstanceAMI":      instanceAMI,
+		"InstanceType":     instanceType,
+*/
--- a/validator/validators/performance/performance_stats.go
+++ b/validator/validators/performance/performance_stats.go
@@ -1,0 +1,63 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package performance
+
+import (
+	"log"
+	"math"
+	"sort"
+)
+
+type Stats struct {
+	Average float64
+	P99     float64 //99% percent process
+	Max     float64
+	Min     float64
+	Period  int //in seconds
+	Std     float64
+}
+
+/*
+CalculateMetricStatisticsBasedOnDataAndPeriod takes in an array of data and returns the average, min, max, p99, and stdev of the data.
+statistics are calculated this way instead of using GetMetricStatistics API because GetMetricStatistics would require multiple
+API calls as only one metric can be requested/processed at a time whereas all metrics can be requested in one GetMetricData request.
+*/
+func CalculateMetricStatisticsBasedOnDataAndPeriod(data []float64, dataPeriod float64) Stats {
+	length := len(data)
+	if length == 0 {
+		return Stats{}
+	}
+
+	sort.Float64s(data)
+
+	min := data[0]
+	max := data[length-1]
+
+	sum := 0.0
+	for _, value := range data {
+		sum += value
+	}
+
+	avg := sum / float64(length)
+
+	if length < 99 {
+		log.Println("Note: less than 99 values given, p99 value will be equal the max value")
+	}
+	p99Index := int(float64(length)*.99) - 1
+	p99Val := data[p99Index]
+
+	stdDevSum := 0.0
+	for _, value := range data {
+		stdDevSum += math.Pow(avg-value, 2)
+	}
+
+	return Stats{
+		Average: avg,
+		Max:     max,
+		Min:     min,
+		P99:     p99Val,
+		Std:     math.Sqrt(stdDevSum / float64(length)),
+		Period:  int(dataPeriod / float64(length)),
+	}
+}
--- a/validator/validators/performance/performance_validator.go
+++ b/validator/validators/performance/performance_validator.go
@@ -4,11 +4,30 @@
 package performance
 
 import (
+	"fmt"
+	"log"
+	"strings"
 	"time"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/cenkalti/backoff/v4"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 
 	"github.com/aws/amazon-cloudwatch-agent-test/internal/awsservice"
 	"github.com/aws/amazon-cloudwatch-agent-test/internal/common"
 	"github.com/aws/amazon-cloudwatch-agent-test/validator/models"
+)
+
+const (
+	ServiceName      = "AmazonCloudWatchAgent"
+	DynamoDBDataBase = "CWAPerformanceMetrics"
+)
+
+var (
+	// The default unit for these metrics is byte. However, we want to convert to MB for easier understanding
+	metricsConvertToMB = []string{"mem_total", "procstat_memory_rss", "procstat_memory_swap", "procstat_memory_data", "procstat_memory_vms", "procstat_write_bytes", "procstat_bytes_sent"}
 )
 
 type PerformanceValidator struct {
@@ -42,6 +61,20 @@ func (s *PerformanceValidator) GenerateLoad() (err error) {
 }
 
 func (s *PerformanceValidator) CheckData(startTime, endTime time.Time) error {
+	metrics, err := s.GetPerformanceMetrics(startTime, endTime)
+	if err != nil {
+		return err
+	}
+
+	perfInfo, err := s.CalculateMetricStatsAndPackMetrics(metrics)
+	if err != nil {
+		return err
+	}
+
+	err = s.SendPacketToDatabase(perfInfo)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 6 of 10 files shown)


validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+
+func (s *PerformanceValidator) GetPerformanceMetrics(startTime, endTime time.Time) ([]types.MetricDataResult, error) {
+	var (
+		metricNamespace              = s.vConfig.GetMetricNamespace()
+		validationMetric             = s.vConfig.GetMetricValidation()
+		ec2InstanceId                = awsservice.GetInstanceId()
+		performanceMetricDataQueries = []types.MetricDataQuery{}
+	)
+	log.Printf("Start getting performance metrics from CloudWatch")
+	for _, metric := range validationMetric {
```
khanhntd, line 170: The performance validator will get all the generator configuration (e.g [logs](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/performance/logs/parameters.yml#L19-L72) )

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		metricDimensions := []types.Dimension{
```
khanhntd, line 171: Getting Instance ID from the EC2 to ensure all metrics are being unique to each other and not duplicate across instances or across workloads

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		for _, dimension := range metric.MetricDimension {
+			metricDimensions = append(metricDimensions, types.Dimension{
+				Name:  aws.String(dimension.Name),
+				Value: aws.String(dimension.Value),
+			})
+		}
+		performanceMetricDataQueries = append(performanceMetricDataQueries, s.buildStressMetricQueries(metric.MetricName, metricNamespace, metricDimensions))
+	}
+
+	metrics, err := awsservice.GetMetricData(performanceMetricDataQueries, startTime, endTime)
```
khanhntd, line 186: `GetMetricData` will [perform GetMetricData](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html) to retrieve all values based on the metrics required by the generator configuration.

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		receiver               = s.vConfig.GetPluginsConfig()
+		commitHash, commitDate = s.vConfig.GetCommitInformation()
+		dataType               = s.vConfig.GetDataType()
+		dataRate               = fmt.Sprint(s.vConfig.GetDataRate())
+		agentCollectionPeriod  = s.vConfig.GetAgentCollectionPeriod().Seconds()
+	)
+	performanceMetricResults := make(map[string]Stats)
+
+	for _, metric := range metrics {
+		metricLabel := strings.Split(*metric.Label, " ")
```
khanhntd, line 141: The `metric.Label` will return [human-readable label associated with the data.](https://github.com/aws/aws-sdk-go-v2/blob/main/service/cloudwatch/types/types.go#L747-L748) (e.g `cloudwatch-agent amazon-cloudwatch-agent procstat_cpu_usage`). However, this can be clipped with aggregated dimensions 
```
"aggregation_dimensions": [
			[
				"InstanceId"
			]
]],
```
which will get unclearable when query metrics. However, for future case such as having another file to monitor but using procstat to monitor (e.g performance test will have addition host resources metrics use case which include procstat), this would get unclear and aggregate by CW for the two monitored files by procstat.

Your guys thoughts on this one?

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		agentCollectionPeriod  = fmt.Sprint(s.vConfig.GetAgentCollectionPeriod().Seconds())
+		// The secondary global index that is used for checking if there are item has already been exist in the table
+		// The performance validator will query based on the UseCaseHash to confirm if the current commit with the use case
+		// has been exist or not? If yes, merge it. If not, sending it to the database
+		// https://github.com/aws/amazon-cloudwatch-agent-test/blob/e07fe7adb1b1d75244d8984507d3f83a7237c3d3/terraform/setup/main.tf#L46-L53
+		kCheckingAttribute = []string{"CommitHash", "UseCase"}
+		vCheckingAttribute = []string{fmt.Sprint(commitHash), receiver}
+	)
+
+	err := backoff.Retry(func() error {
```
khanhntd, line 108: This is to retry for case with [optimism locks](https://stackoverflow.com/questions/129329/optimistic-vs-pessimistic-locking). DynamoDB will lock the database since there are another requests going on. Therefore, we will need to retry since we always want the latest data for the performance metrics  (e.g running performance on the same commit to do an A-A testing)

generator/resources/ec2_performance_test_matrix.json
```diff
@@ -6,7 +6,7 @@
         "ami": "cloudwatch-agent-integration-test-al2*",
         "arc": "amd64",
         "binaryName": "amazon-cloudwatch-agent.rpm",
-        "valuesPerMinute": 10,
+        "valuesPerMinute": 100,
```
khanhntd, line 9: Changing the metrics/logs/traces sending per minutes to reflect more performance metrics instead of 10/100/1000.

internal/awsservice/dynamodb.go
```diff
@@ -45,7 +45,7 @@ func AddItemIntoDatabaseIfNotExist(databaseName string, checkingAttribute, check
 				"#second_attribute": checkingAttribute[1],
 			},
 			ExpressionAttributeValues: map[string]types.AttributeValue{
-				":first_attribute":  &types.AttributeValueMemberN{Value: checkingAttributeValue[0]},
+				":first_attribute":  &types.AttributeValueMemberS{Value: checkingAttributeValue[0]},
```
khanhntd, line 48: Since we are using [UseCase  (a string) and Commit Hash (a string)](https://github.com/aws/amazon-cloudwatch-agent-test/pull/153/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R100-R105), therefore, we will change from `AttributeValueMemberN` to `AttributeValueMemberS` (number to string)

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		existingPerfInfo, err := awsservice.GetItemInDatabase(DynamoDBDataBase, "UseCaseHash", kCheckingAttribute, vCheckingAttribute, perfInfo)
+		if err != nil {
+			return err
+		}
+
+		// Get the latest performance information from the database and update by merging the existing one
+		// and finally replace the packet in the database
+		maps.Copy(existingPerfInfo["Results"].(map[string]interface{}), perfInfo["Results"].(map[string]interface{}))
```
khanhntd, line 116: The reason why we want to merge is that there are many limitations on number of queries that DynamoDB can provide (e.g DynamoDB only provides maximum of [2 key queries in the database](https://stackoverflow.com/questions/65390063/dynamodbexception-conditions-can-be-of-length-1-or-2-only)). So, there are no ways for us to provide more granularity (e.g query commit hash and number of values per metric). Therefore, we need to merge all the performance metrics sending with 100 metrics/minute , 1000 metrics/minute, 5000 metrics/minute together. Afterwards, replace the existing database to avoid having duplicate key in the database.

validator/validators/performance/performance_stats.go
```diff
@@ -0,0 +1,63 @@
...
+
+package performance
+
+import (
+	"log"
+	"math"
+	"sort"
+)
+
+type Stats struct {
```
khanhntd, line 12: The statistics metrics that customers will get when showing with the frontend page.[ Porting from intern projects](https://github.com/aws/amazon-cloudwatch-agent-test/blob/d5105cdc461c6fcb13049cf2d38c287674d94e21/test/performancetest/performance_query_utils.go#L49-L57)

validator/validators/performance/performance_validator.go
```diff
@@ -42,6 +61,20 @@ func (s *PerformanceValidator) GenerateLoad() (err error) {
 }
 
 func (s *PerformanceValidator) CheckData(startTime, endTime time.Time) error {
```
adam-mateen, line 63: nit: The name `CheckData()` does not match with what this function is doing. Fetching metrics, post-processing them and uploading a report to a database...

I realize the name is part of the interface.
Maybe add another optional function to the interface for `PublishResults()`.

Or maybe that causes a different problem - having big interfaces that are loosely implemented.
Don't want an "Animal" interface with functions like "Speak()" since not animals can speak ;)

validator/validators/performance/performance_validator.go
adam-mateen: Since we control the JSON configuration for the performance tests, we ultimately control aggregation_dimensions and the metric label. So as long as you can clearly document and avoid a future developer turning on aggregation_dimensions and screwing up the performance results... Then I am okay with it.

validator/validators/performance/performance_validator.go
khanhntd: Agree. However, an interface which has general method would be a better case here since not Animal can all `Fly()` but most of all the `Animals` can `Speak`.  The plan for me would be replace all our existing test with the validator and `CheckData` can be applied to all our existing tests !!

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+	performanceMetricResults := make(map[string]Stats)
+
+	for _, metric := range metrics {
+		metricLabel := strings.Split(*metric.Label, " ")
+		metricName := metricLabel[len(metricLabel)-1]
+		metricValues := metric.Values
+		//Convert every bytes to MB
+		if slices.Contains(metricsConvertToMB, metricName) {
+			for i, val := range metricValues {
+				metricValues[i] = val / (1000000)
```
adam-mateen, line 147: Should we be using base 10? Or use MiB (1024*1024). I prefer the latter.

khanhntd: I'm good with both. But for you only, I will change to 1024 * 1024 lol

validator/validators/performance/performance_validator.go
khanhntd: Then I'm cool with it too !!

validator/validators/performance/performance_models.go
```diff
@@ -0,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package performance
+
+type PerformanceInformation map[string]interface{}
```
adam-mateen, line 6: nit: rename TestResult or TestInfo. The package is already named performance, so you don't really need it in the name.

validator/validators/performance/performance_stats.go
adam-mateen: I think this should include more than just P99. Why not add P95, P90?
It is easy to add more stats in the future, or is changing it "hard"?

validator/validators/performance/performance_stats.go
```diff
@@ -0,0 +1,63 @@
...
+	Period  int //in seconds
+	Std     float64
+}
+
+/*
+CalculateMetricStatisticsBasedOnDataAndPeriod takes in an array of data and returns the average, min, max, p99, and stdev of the data.
+statistics are calculated this way instead of using GetMetricStatistics API because GetMetricStatistics would require multiple
+API calls as only one metric can be requested/processed at a time whereas all metrics can be requested in one GetMetricData request.
+*/
+func CalculateMetricStatisticsBasedOnDataAndPeriod(data []float64, dataPeriod float64) Stats {
```
adam-mateen, line 26: It seems a bit silly to have a function like this since CloudWatch will return avg, min, max, p99 for us. No need to change anything now, but maybe there is a way to just query this information from CloudWatch... For example if the performance benchmark runs for an hour (beginning at the 0th minute of a wall clock hour) then we just query CW for the metric(s) with an hourly period.

Better yet, we basically be pulling as much information out of CloudWatch about the performance benchmark and storing it in DynamoDB. That way we could overlay time series graphs from multiple test runs and see how P99 changes over time instead of just a single P99 value for the entire test run.

validator/validators/performance/performance_stats.go
khanhntd: Yes its easy to do so. That's the good thing about non-SQL Database. We don't need to define the structure for it beforehand but only some index. Therefore, easily adding with changing the calculations

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		Id: aws.String(strings.ToLower(metricName)),
+	}
+	return metricDataQuery
+}
+
+// packIntoPerformanceInformation will package all the information into the required format of MongoDb Database
+// https://github.com/aws/amazon-cloudwatch-agent-test/blob/e07fe7adb1b1d75244d8984507d3f83a7237c3d3/terraform/setup/main.tf#L8-L63
+func packIntoPerformanceInformation(receiver, dataType, collectionPeriod, commitHash string, commitDate int64, result interface{}) PerformanceInformation {
+	instanceAMI := awsservice.GetImageId()
+	instanceType := awsservice.GetInstanceType()
```
adam-mateen, line 217: Any chance GetImageId() or GetInstanceType could fail?
If so, what can we do to handle that type of error?
1. log error?
2. retry?
3. fail the test?

validator/validators/performance/performance_validator.go
```diff
@@ -57,3 +90,153 @@ func (s *PerformanceValidator) Cleanup() error {
...
+		"CommitHash":       commitHash,
+		"DataType":         dataType,
+		"Results":          result,
+		"CollectionPeriod": collectionPeriod,
+		"InstanceAMI":      instanceAMI,
+		"InstanceType":     instanceType,
+	}
+}
+
+func isAllValuesGreaterThanOrEqualToZero(values []float64) bool {
```
adam-mateen, line 232: This function exists in `metric_dimension_test.go`. Looks to be the same other than it taking in an additional `metricName string`.

I was going to suggest sharing the code.
But this is pretty darn simple I don't think it is fine to re implement here.

validator/validators/performance/performance_validator.go
adam-mateen: This assumes a new EC2 instance ID per test run, right? Might want to add this type of assumption to the README or Wiki. I could see myself trying to run this test on my permanent EC2 instance multiple times.

validator/validators/performance/performance_validator.go
khanhntd:
```
Any chance GetImageId() or GetInstanceType could fail?
```
It would be less likely only in the case of turning off IMDS. However, this could be prevented by the following options:
* Instead of calling IMDS here, we could pass it as an parameters in the [generator configuration](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validation_config.go#L29-L48) (e.g instance_id in parameters.yml)  since [we know this kind of information when we create the EC2 instances](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/terraform/performance/main.tf#L54-L69) and get it there

validator/validators/performance/performance_validator.go
```diff
@@ -4,13 +4,32 @@
...
 
+const (
+	ServiceName      = "AmazonCloudWatchAgent"
+	DynamoDBDataBase = "CWAPerformanceMetrics"
+)
+
+var (
+	// The default unit for these metrics is byte. However, we want to convert to MB for easier understanding
+	metricsConvertToMB = []string{"mem_total", "procstat_memory_rss", "procstat_memory_swap", "procstat_memory_data", "procstat_memory_vms", "procstat_write_bytes", "procstat_bytes_sent"}
+)
```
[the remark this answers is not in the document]
khanhntd: I agree with you regarding `hashmap` in terms of getting key (time complexity: O(1)) However, small collections of items will have the same (O(n) with small collections will ~ to O(1)) and I would avoid creating `empty-struct` since its only served as a vague placeholder as much as possible. Simplicity in exchange (performance won't be the case here though)

adam-mateen (approved):

validator/validators/performance/performance_validator.go
adam-mateen: I am fine leaving it as is.
The error seems very unlikely to happen as long as we control the instance the test runs on and properly turn on IMDS. I think logging a failure is good enough. Up to you though.

test/performance/logs/parameters.yml
```diff
@@ -8,7 +8,7 @@ number_monitored_logs: 1000
 # Number of metrics to be sent or number of log lines being written  each minute
 values_per_minute: "<values_per_minute>"
 # Number of seconds the agent should run and collect the metrics. In this case, 10 minutes
```
williazz, line 10: nit: update comment

validator/validators/performance/performance_models.go
williazz: Question: if we know the properties, why is PerformanceInformation unstructured?

validator/validators/performance/performance_validator.go
khanhntd: That's correct. However, you can also run `run this test on my permanent EC2 instance multiple times`. This does not stop you from doing that

test/performance/logs/parameters.yml
khanhntd: Thanks. I have no reason or recollection why I changed it back to 60 instead of 600.

validator/validators/performance/performance_models.go
khanhntd: The reason why it would be structured this way [is because](https://github.com/aws/amazon-cloudwatch-agent-test/pull/153/files#diff-334efd747575193d3505843e769add29bb2818c73a18ca77c9c4f97105ec3104R120)
```
func ReplaceItemInDatabase(databaseName string, packet map[string]interface{}) error {
	item, err := attributevalue.MarshalMap(packet)
	if err != nil {
		return err
	}

	_, err = DynamodbClient.PutItem(ctx,
		&dynamodb.PutItemInput{
			Item:      item,
			TableName: aws.String(databaseName),
		})

	return err
}
```
since you can argue it would be better to use struct for specific test but for general case, it would be better to use `map[string]interface{}`

validator/validators/performance/performance_validator.go
williazz: yeah i was going to say the same thing, this should probably be a shared resource

khanhntd: Its pretty simple and not commonly used so I have not thought of sharing the function between the two. I'm fine with create a common function between the two if you guys has a strong urge with it.

williazz (approved):

adam-mateen (approved):

PR #162, opened 2023-03-09
Add feature validator to the validator

khanhntd:
# Description of the issue
To start the validator, there are three stages:
* [GenerateLoad](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L10):  Sending the load to CloudWatchAgent
* [CheckData](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L11) : To get metrics defined [by the yaml and validate the metrics](https://github.com/aws/amazon-cloudwatch-agent-test/compare/main...khanhntd:amazon-cloudwatch-agent-test:add_performance?expand=1#diff-4182821602cec8db84c8b29427a65c5a63beff4c19bcf93e305966bed05a721eR20-R72)
* [Cleanup](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/models/validator_factory.go#L12) to clean up all the resources created by the validator.

For feature validation, we only confirm as a **sanity check if CWA able to receive metric/logs/trace without E2E testing.**

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
By using `terraform apply --auto-approve` and running it locally 
```
ueue_0_rx_cnt [{driver ena} {interface en0}]} {swap_free []} {swap_used_percent []} {processes_blocked []} {processes_running []} {netstat_tcp_close []} {netstat_udp_socket []} {mem_available_percent []} {mem_used_percent []}]  }
null_resource.integration_test (remote-exec): 2023/03/10 21:44:43 Start to sleep 16.592645 s for the metric to be available in the beginning of next minute
null_resource.integration_test: Still creating... [1m0s elapsed]
null_resource.integration_test: Still creating... [1m10s elapsed]
null_resource.integration_test (remote-exec): 2023/03/10 21:45:00 Start to generate load in 60.000000 s for the agent to collect and send all the metrics to CloudWatch within the datapoint period
null_resource.integration_test: Still creating... [1m20s elapsed]
null_resource.integration_test: Still creating... [1m30s elapsed]
null_resource.integration_test: Still creating... [1m40s elapsed]
null_resource.integration_test: Still creating... [1m50s elapsed]
null_resource.integration_test: Still creating... [2m0s elapsed]
null_resource.integration_test: Still creating... [2m10s elapsed]
[...]

PR #165, opened 2023-03-11
Add support for feature validation with MacOS and Windows OS

khanhntd:
# Description of the issue
Continue of #162 
Adding support for Windows and MacOs with features validation. Therefore, we can increase the coverage and increase the confidence, decrease regression with CloudWatch Agent for all the supported OS  before releasing to CloudWatch Agent's customers

# Description of changes
* [Add support sample count for validation and value validation for metrics](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-797a68b028fdf868b4b0cba0f575d9fad8abcb4f2a4fb6df143d4c6beadf256dR55-R60). Therefore, by using the following yaml, the validator will validate the metrics themselves if it within bound or having the same value
```
metric_validation:
-   metric_name: "statsd_0"
    metric_value: 0.0
    metric_sample_count: 1
```
* Add [support for logs validation](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-797a68b028fdf868b4b0cba0f575d9fad8abcb4f2a4fb6df143d4c6beadf256dR62-R65). Therefore, by using the following yaml, the validator will validate the logs themselves if it really have 2 of the `This is a log line.` in the log stream `test1.log`
```
log_validation:
  - log_value: "This is a log line."
    log_lines: 2
    log_stream: "test1.log"
```
* Fix sending initial metric/logs during the first minute (e.g [statsd](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-781bb3810ede78771b4deb63bb6bb797a2d06434f0d844429b5ad16a20304e0fR122-R128))
# License

# Tests
There are two ways I have run to confirm the validation:
* **MacOs with locally**
```
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Checking i-084c95fdee99d435b/test1.log
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Done paginating log events for i-084c95fdee99d435b/test1.log and found 2 logs
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Start to validate log '# 0 - This is a log line.' with number of logs lines 1 within log group i-084c95fdee99d435b, log stream test1.log, start time 2023-03-25 00:26:00 +0000 GMT and end time 2023-03-25 00:27:00 +0000 GMT
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Checking i-084c95fdee99d435b/test1.log
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Done paginating log events for i-084c95fdee99d435b/test1.log and found 2 logs
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Start to validate log '# 1 - This is a log line.' with number of logs lines 1 within log group i-084c95fdee99d435b, log stream test1.log, start time 2023-03-25 00:26:00 +0000 GMT and end time 2023-03-25 00:27:00 +0000 GMT
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Checking i-084c95fdee99d435b/test1.log
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Done paginating log events for i-084c95fdee99d435b/test1.log and found 2 logs
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Test case: macos_feature, validate type: feature has been successfully validated
null_resource.integration_test (remote-exec): 2023/03/25 00:29:00 Finish validation in 3.35446068055 minutes
```

* **Windows with locally**
```
null_resource.integration_test (remote-exec): 2023/03/25 00:01:00 Start to validate log '# 1 - This is a log line.' with number of logs lines 1 within log group i-0afb0bd960dd36af6, log stream test1.log, start time 2023-03-24 23:58:00 +0000 UTC and end time 2023-03-24 23:59:00 +0000 UTC
null_resource.integration_test (remote-exec): 2023/03/25 00:01:00 Checking i-0afb0bd960dd36af6/test1.log
null_resource.integration_test (remote-exec): 2023/03/25 00:01:00 Done paginating log events for i-0afb0bd960dd36af6/test1.log and found 2 logs
null_resource.integration_test (remote-exec): 2023/03/25 00:01:00 Test case: win_feature, validate type: feature has been successfully validated
null_resource.integration_test (remote-exec): 2023/03/25 00:01:00 Finish validation in 3.4197621966666665 minutes
null_resource.integration_test: Still creating... [3m50s elapsed]
#< CLIXML
<Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04"><Obj S="progress" RefId="0"><TN RefId="0"><T>System.Management.Automation.PSCustomObject</T><T>System.Object</T></TN><MS><I64 N="SourceId">1</I64><PR N="Record"><AV>Preparing modules for first use.</AV><AI>0</AI><Nil /><PI>-1</PI><PC>-1</PC><T>Completed</T><SR>-1</SR><SD> </SD></PR></MS></Obj><Obj S="progress" RefId="1"><TNRef RefId="0" /><MS><I64 N="SourceId">1</I64><PR N="Record"><AV>Preparing modules for first use.</AV><AI>0</AI><Nil /><PI>-1</PI><PC>-1</PC><T>Completed</T><SR>-1</SR><SD> </SD></PR></MS></Obj></Objs>#< CLIXML
<Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04"><Obj S="progress" RefId="0"><TN RefId="0"><T>System.Management.Automation.PSCustomObject</T><T>System.Object</T></TN><MS><I64 N="SourceId">1</I64><PR N="Record"><AV>Preparing modules for first use.</AV><AI>0</AI><Nil /><PI>-1</PI><PC>-1</PC><T>Completed</T><SR>-1</SR><SD> </SD></PR></MS></Obj></Objs>null_resource.integration_test: Creation complete after 3m53s [id=9149370514043737241]
```

* Windows when running with fork https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/4517440449/attempts/1
```
data.aws_ami.latest: Reading...
module.basic_components.data.aws_iam_role.cwagent_iam_role: Reading...
module.basic_components.data.aws_iam_role.cwagent_iam_role: Read complete after 0s [id=cwa-e2e-iam-role]
module.basic_components.data.aws_iam_instance_profile.cwagent_instance_profile: Read complete after 0s [id=AIPASN2NTSDK6VEHCSTX4]
module.basic_components.data.aws_security_group.security_group: Read complete after 0s [id=sg-088a9b6b4c1fe3bfb]
data.aws_ami.latest: Read complete after 1s [id=ami-0a4728590116240e9]
aws_instance.cwagent: Refreshing state... [id=i-0efd660064ecbbc65]
module.basic_components.data.aws_vpc.vpc: Read complete after 1s [id=vpc-04287887fb0aab3db]
module.basic_components.data.aws_subnets.public_subnet_ids: Reading...
module.basic_components.data.aws_subnets.public_subnet_ids: Read complete after 0s [id=us-west-2]
null_resource.integration_test: Refreshing state... [id=7183864745353502804]
```


internal/common/logs.go
```diff
@@ -50,15 +53,19 @@ func writeToLogs(filePath string, duration time.Duration, logLinesPerMinute int)
 	defer ticker.Stop()
 	endTimeout := time.After(duration)
 
-	//loop until the test duration is reached
+	// Sending the logs within the first minute before the ticker kicks in the next minute
+	for i := 0; i < logLinesPerMinute; i++ {
+		_, err := f.WriteString(fmt.Sprintf(logLine, i))
```
khanhntd, line 58: The first sending would be more prioritized in catching error. If it fails during the first time, then we would return error immediately  without worrying about [the second write](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-32ea2f52d0d817b42a874abb6e550c6c2040c34707954b8efd66e0c474660738R68). However, if the first write success, then it would guarantee to have no errors in the second write.

internal/awsservice/cloudwatchlogs.go
```diff
@@ -97,7 +96,7 @@ func getLogsSince(logGroup, logStream string, since, until *time.Time) ([]string
 		if err != nil {
 			if errors.As(err, &rnf) && attempts <= StandardRetries {
 				// The log group/stream hasn't been created yet, so wait and retry
-				time.Sleep(time.Minute)
+				time.Sleep(30 * time.Second)
```
khanhntd, line 99: Reduce the number of time for validating logs. Right now, for every time we call `ValidateLogs`, if we fail, we retry for 1 minute with a total of 5 retries ( 5 minutes); therefore, reducing to 30s with a total of 3 retries (1'30s)

internal/common/metrics.go
```diff
@@ -105,11 +119,18 @@ func sendStatsdMetrics(metricPerMinute int, duration time.Duration) error {
 	defer ticker.Stop()
 	endTimeout := time.After(duration)
 
+	// Sending the statsd metric within the first minute before the ticker kicks in the next minute
```
khanhntd, line 122: Whenever we call `sendStatsdMetrics`, we create a ticker for every [60 second](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-781bb3810ede78771b4deb63bb6bb797a2d06434f0d844429b5ad16a20304e0fR118), however; the first call (e.g from 0 to 1 minutes) won't be sent; therefore, start sending statsd metrics here during the first time for and after every 60s, we will send statsd metrics regualarly
```
case <-ticker.C:
			for t := 0; t < metricPerMinute; t++ {
				client.Inc(fmt.Sprint(t), int64(t), 1.0)
				client.Count(fmt.Sprint(t), int64(t), []string{}, 1.0)
			}
```

test/feature/mac/agent_config.json
```diff
@@ -36,43 +38,35 @@
...
-              "devfs"
+              "*"
             ],
             "measurement": [
               "free",
               "used_percent"
             ],
-            "drop_device": true
-          },
-          "ethtool": {
```
khanhntd, line 66: Ethtool is [not supported in MacOS](https://github.com/influxdata/telegraf/blob/6a3b27126a26dd43d2a287b4c9d8831dca6bf687/plugins/inputs/ethtool/ethtool_notlinux.go) and only supported in Linux since we are using [ethtool from telegraf](https://github.com/influxdata/telegraf/tree/6a3b27126a26dd43d2a287b4c9d8831dca6bf687/plugins/inputs/ethtool). Even though [CWA rules supports Darwin Rules](https://github.com/aws/amazon-cloudwatch-agent/blob/c85501042b088014ec40b636a8b6b2ccc9739738/translator/translate/metrics/metrics_collect/ethtool/ethtool.go#L59), it is not possible for collecting ethtool metrics in MacOS

test/feature/mac/parameters.yml
```diff
@@ -1,5 +1,5 @@
 # Receivers that agent needs to tests
-receivers: "host"
+receivers: ["host","statsd"]
```
khanhntd, line 2: `Host` is instance system resources such as (e.g cpu,mem). We will add `emf` and `collectd` in the following future.

validator/validators/stress/stress_validator.go
```diff
@@ -113,11 +113,11 @@ var (
 				"net_packets_sent":     float64(700),
 			},
 			"logs": {
-				"procstat_cpu_usage":   float64(225),
```
khanhntd, line 116: Sending logs during the [first minute changing the logs stress metrics ](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-32ea2f52d0d817b42a874abb6e550c6c2040c34707954b8efd66e0c474660738R57)

validator/validators/feature/feature_validator.go
```diff
@@ -31,25 +38,59 @@ func (s *FeatureValidator) GenerateLoad() error {
...
-	if err != nil {
+	if err := common.StartLogWrite(agentConfigFilePath, agentCollectionPeriod, dataRate); err != nil {
 		multiErr = multierr.Append(multiErr, err)
 	}
 
 	// Sending metrics based on the receivers; however, for scraping plugin  (e.g prometheus), we would need to scrape it instead of sending
-	err = common.StartSendingMetrics(receiver, agentCollectionPeriod, dataRate)
-	if err != nil {
-		multiErr = multierr.Append(multiErr, err)
+	for _, receiver := range receivers {
```
khanhntd, line 49: Feature validation is cover all of the use case for a given OS. Therefore, we don't want to use a single receiver here (e.g prometheus, emf) unlike [all other validator](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/validator/validators/performance/performance_validator.go#L53-L59)

Makefile
```diff
@@ -5,8 +5,8 @@ TOOLS_BIN_DIR := $(abspath ./build/tools)
 LINTER = $(TOOLS_BIN_DIR)/golangci-lint
 
 WIN_BUILD = GOOS=windows GOARCH=amd64 go build -trimpath -o $(BUILD_SPACE)/validator/windows/amd64
-LINUX_AMD64_BUILD = GOOS=linux GOARCH=amd64 go build -trimpath -o $(BUILD_SPACE)/validator/linux/amd64
-LINUX_ARM64_BUILD = GOOS=linux GOARCH=arm64 go build -trimpath -o $(BUILD_SPACE)/validator/linux/arm64
+LINUX_AMD64_BUILD = CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -trimpath -o $(BUILD_SPACE)/validator/linux/amd64
+LINUX_ARM64_BUILD = CGO_ENABLED=0  GOOS=linux GOARCH=arm64 go build -trimpath -o $(BUILD_SPACE)/validator/linux/arm64
```
adam-mateen, line 9: Why is it necessary to disable CGO here?
There are some benefits to using the default (enabled):
https://stackoverflow.com/questions/64531437/why-is-cgo-enabled-1-default

internal/common/logs.go
```diff
@@ -50,15 +53,19 @@ func writeToLogs(filePath string, duration time.Duration, logLinesPerMinute int)
...
+
 	for {
 		select {
 		case <-ticker.C:
 			for i := 0; i < logLinesPerMinute; i++ {
-				_, err := f.WriteString(fmt.Sprintf("# %d - This is a log line.", i))
-				if err != nil {
-					log.Printf("Error in writing a string to the file %s: %v ", filePath, err)
-				}
+				f.WriteString(fmt.Sprintf(logLine, i))
```
adam-mateen, line 68: It is not a guarantee that just because the first write succeeded every other write will also succeed.
That said, I think it is fine for "test code" to have a bit less error checking.
Another option is to use "assert" or "panic" on any unexpected failure (no need to gracefully handle/recover from errors, just fail loudly).

internal/common/metrics.go
```diff
@@ -53,6 +53,28 @@ func sendCollectDMetrics(metricPerMinute int, duration time.Duration) error {
...
+		err = client.Write(ctx, &api.ValueList{
+			Identifier: api.Identifier{
+				Host:   exec.Hostname(),
+				Plugin: fmt.Sprint(t),
+				Type:   "gauge",
+			},
+			Time:     time.Now(),
+			Interval: time.Minute,
+			Values:   []api.Value{api.Gauge(t)},
+		})
```
adam-mateen, line 67:
```
&api.ValueList{
			Identifier: api.Identifier{
				Host:   exec.Hostname(),
				Plugin: fmt.Sprint(t),
				Type:   "gauge",
			},
			Time:     time.Now(),
			Interval: time.Minute,
			Values:   []api.Value{api.Gauge(t)},
		}
```
looks duplicated in places.

I suggest a helper/constructor to return the `api.ValueList`.

```
func newValueList(val int) *api.ValueList {
...
}

File: internal/common/logs.go [hunk 11]
```diff
@@ -153,3 +162,12 @@ func GenerateLogConfig(numberMonitoredLogs int, filePath string) error {
 
 	return nil
 }
+
+// getTempFolder gets the temp folder for generate logs
+// depends on the operating system
+func getTempFolder() string {
+	if runtime.GOOS == "windows" {
+		return "C:"
```
adam-mateen, line 170: nit: Windows should have an equivalent temp folder. I don't recommend we use the root drive.
```
C:\Users\Administrator\AppData\Local\Temp
```
or whatever is in `%TEMP%` env var might be better.

Makefile
khanhntd: The problem with that would be: if turning on `CGO_ENABLED=1` and building in Debian and doing a cross compilation and putting up into an AL2, then the AL2 host require load the host OS's native libraries (in this case, its `glibc`). However, different OS has different GLIB version and Github Action; therefore, causing the different in cross compilation with the following error 
```
null_resource.performance (remote-exec): ./validator: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./validator)
null_resource.performance (remote-exec): ./validator: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./validator)
```

**Note**:  AL2 by default having a [glibc of 2.26 while Al2022 has GLIBC 2.32](https://repost.aws/questions/QUrXOioL46RcCnFGyELJWKLw/glibc-2-27-on-amazon-linux-2). So it would be better to have a build that can cross compile all OS instead of OS that has specific GLIBC version.

internal/common/logs.go
khanhntd: Interesting !! Initially I thought we only need to write 1 or two files to this `C:\` folder so `C:\` is fine. I'm good with changing it to `C:\Users\Administrator\AppData\Local\Temp`

internal/common/metrics.go
```diff
@@ -84,16 +106,8 @@ func sendCollectDMetrics(metricPerMinute int, duration time.Duration) error {
...
-		Address:     ":8125",
-		Prefix:      "statsd",
-		UseBuffered: true,
-		// interval to force flush buffer. full buffers will flush on their own,
-		// but for data not frequently sent, a max threshold is useful
-		FlushInterval: 300 * time.Millisecond,
-	}
-	client, err := statsd.NewClientWithConfig(statsdClientConfig)
+	// https://github.com/DataDog/datadog-go#metrics
+	client, err := statsd.New("127.0.0.1:8125", statsd.WithMaxMessagesPerPayload(100), statsd.WithNamespace("statsd"), statsd.WithoutTelemetry())
```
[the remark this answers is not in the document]
khanhntd: `Datadog` has more flexibility than I imagine:
* It can [send events and service checks](https://github.com/DataDog/datadog-go#usage)
* It also supports [distribution](https://docs.datadoghq.com/metrics/custom_metrics/dogstatsd_metrics_submission/#distribution)
* Supports [more options in control the packsize under high throughput environment - e.g control the packet size when sending through UDP ](https://github.com/DataDog/datadog-go/blob/7d2f91bc7a91003561b6657d5974d92584833f98/statsd/options.go)

Makefile
adam-mateen: There is no plan or reason to compile the test code on darwin and copy it to an AL2 host and run it there.

Doesn't the test code get copied to the host-under-test and run directly on it?

test/performance/collectd/parameters.yml
```diff
@@ -1,12 +1,12 @@
...
 
 test_case: "collectd_performance"
 validate_type: "performance"
 data_type: "metrics"
 # Number of metrics to be sent or number of log lines being written  each minute
 values_per_minute: "<values_per_minute>"
-# Number of seconds the agent should run and collect the metrics. In this case, 5 minutes
-agent_collection_period: 300 
+# Number of seconds the agent should run and collect the metrics. In this case, 3 minutes
+agent_collection_period: 180
```
adam-mateen, line 9: Might need more than 3 minutes to get an accurately measure steady state performance of CWA. Why do you feel confident to reduce it from 5?

validator/validators/feature/feature_validator.go
```diff
@@ -64,3 +105,87 @@ func (s *FeatureValidator) Cleanup() error {
...
+			return false
+		}
+		actualNumberOfLogLines := 0
+		for _, l := range logs {
+			if strings.Contains(l, logLine) {
+				actualNumberOfLogLines += 1
+			}
+		}
+
+		return numberOfLogLine <= actualNumberOfLogLines
```
adam-mateen, line 125: Should this be a more strict `==` check? If we know exactly how many log lines were written to the file, we should be able to expect that exact amount, right?

Makefile
khanhntd: The validator is not a `test package` but an [validator application](https://github.com/aws/amazon-cloudwatch-agent-test/tree/main/validator). So yes it is in need to compile the test code in X host and put it on Y host. Windows has problems with allocation memory when using `go test` so its only possible to use validator

validator/validators/stress/stress_validator.go
```diff
@@ -113,12 +113,12 @@ var (
...
-				"procstat_memory_rss":  float64(440000000),
+				"procstat_cpu_usage":   float64(400),
```
adam-mateen, line 116: is this 400% CPU. Out of how many cores?

validator/validators/performance/performance_validator.go
```diff
@@ -48,7 +48,7 @@ func (s *PerformanceValidator) GenerateLoad() (err error) {
 		agentConfigFilePath   = s.vConfig.GetCloudWatchAgentConfigPath()
 		dataType              = s.vConfig.GetDataType()
 		dataRate              = s.vConfig.GetDataRate()
-		receiver              = s.vConfig.GetPluginsConfig()
+		receiver              = s.vConfig.GetPluginsConfig()[0] //Assuming one plugin at a time
```
adam-mateen, line 51: If this only uses index 0, then when will the other plugins get validated? Why bother passing in plugins that the function does not need?

validator/validators/stress/stress_validator.go
khanhntd: Out of 4 cores based on the[t3a.xlarge](https://instances.vantage.sh/aws/ec2/t3a.xlarge) and yes, its depends on the instance size

test/metric_value_benchmark/statsd_test.go
```diff
@@ -167,13 +171,7 @@ func (t *StatsdTestRunner) validateStatsdMetric(metricName string) status.TestRe
...
 // It sends each metric with dimensions at a 1 second interval.
 func sendStatsd() error {
-	config := statsd.ClientConfig{
-		Address:       ":8125",
-		Prefix:        "",
-		UseBuffered:   true,
-		FlushInterval: 300 * time.Millisecond,
-	}
-	client, err := statsd.NewClientWithConfig(&config)
+	client, err := statsd.New("127.0.0.1:8125", statsd.WithMaxMessagesPerPayload(100))
```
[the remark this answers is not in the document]
khanhntd: Yes to with [WithBufferFlushInterval](https://github.com/DataDog/datadog-go/blob/7d2f91bc7a91003561b6657d5974d92584833f98/statsd/options.go#L155-L160)
```
client, err := statsd.New("127.0.0.1:8125", statsd.WithBufferFlushInterval(300 * time.Millisecond))
```

validator/validators/stress/stress_validator.go
```diff
@@ -49,14 +49,14 @@ var (
 				"net_packets_sent":     float64(105),
 			},
 			"logs": {
-				"procstat_cpu_usage":   float64(40),
-				"procstat_memory_rss":  float64(152000000),
+				"procstat_cpu_usage":   float64(200),
```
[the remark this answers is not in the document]
khanhntd: Yes that's correct. It is being defined in the generator configuration (e.g [collectd](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/performance/collectd/parameters.yml#L1)). The possible and easiest way I would say would be having a validation for the `ValidatorConfig` (e.g receivers only supports the following values ["logs", "statsd", "collectd"]). However, I agree with your thoughts
```
Somehow make it clear that the name "logs" here must match the name in the YML, or in some other go code.
```

test/performance/collectd/parameters.yml
khanhntd: First minute and second minute would be enough for the buffer to fill our metrics/traces/logs. However, I agree with you `accurately measure steady state performance of CWA` the more time it runs since the average for 5 minutes would mostly similar to 7 minutes within an error rate of 10%.

internal/common/logs.go
SaxyPandaBear: I don't understand. What second write are you referring to? The loop below only has one WriteString call.

validator/validators/basic/basic_validator.go
```diff
@@ -0,0 +1,192 @@
...
+			return false
+		}
+		actualNumberOfLogLines := 0
+		for _, l := range logs {
+			if strings.Contains(l, logLine) {
+				actualNumberOfLogLines += 1
+			}
+		}
+
+		return numberOfLogLine <= actualNumberOfLogLines
```
SaxyPandaBear, line 126: Why is this <=?

validator/validators/basic/basic_validator.go
```diff
@@ -0,0 +1,192 @@
...
+	)
+	switch dataType {
+	case "logs":
+		awsservice.DeleteLogGroup(ec2InstanceId)
+	}
+
+	return nil
+}
+
+func (s *BasicValidator) ValidateLogs(logStream string, logLine string, numberOfLogLine int, startTime, endTime time.Time) error {
```
SaxyPandaBear, line 110: Nit: startTime and endTime share a type, but logStream and logLine don't.
 ```go
func (s *BasicValidator) ValidateLogs(logStream, logLine string, numberOfLogLine int, startTime, endTime time.Time) error {
```

File: internal/common/logs.go (hunk 1 above)
Comment (khanhntd) [reply internal/common/logs.go:58]: So we have two writes:
* First write will be in the first minute
```
for i := 0; i < logLinesPerMinute; i++ {
		_, err := f.WriteString(fmt.Sprintf(logLine, i))
```
* Second write would be after 60 second
```
for {
		select {
		case <-ticker.C:
			for i := 0; i < logLinesPerMinute; i++ {
				_, err := f.WriteString(fmt.Sprintf("# %d - This is a log line.", i))
				if err != nil {
					log.Printf("Error in writing a string to the file %s: %v ", filePath, err)
				}
```
So we only need to catch the first error instead of catching the second write errors

File: validator/validators/stress/stress_validator.go (hunk 18 above)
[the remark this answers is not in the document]
Comment (khanhntd) [reply validator/validators/stress/stress_validator.go:52]: Sorry intend to make it in the next PR. However, I'm fine with changing it [within this PR](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165/files#diff-797a68b028fdf868b4b0cba0f575d9fad8abcb4f2a4fb6df143d4c6beadf256dR91-R96)

File: validator/validators/basic/basic_validator.go (hunk 20 above)
Comment (khanhntd) [reply validator/validators/basic/basic_validator.go]: Agree with you. Thanks for catching it, Andrew. Not sure the reason why I changed it for starttime and endtime but not logstream and logline.

File: validator/validators/basic/basic_validator.go (hunk 19 above)
Comment (khanhntd) [reply validator/validators/basic/basic_validator.go:126]: Yes [it has been discussed here](https://github.com/aws/amazon-cloudwatch-agent-test/pull/165#discussion_r1152239780).
```
It would be better to <= since there are logs we cannot control (e.g Security logs in Windows write every minute while System logs write every 5 minute). If in windows 2023 change the default interval, would == be true anymore? If we understand the changes sure; however, it would be better to <= to avoid flakiness and make it more deterministic
```

test/performance/collectd/parameters.yml
khanhntd: Yes agree. Changing to the original 5 minutes instead of 3 minutes. It should be fine for waiting 5 minutes for increasing accuracy.

PR #178, opened 2023-03-22
Sample count

khanhntd:
# Description of the issue
_Describe the problem or feature in addition to a link to the issues._

# Description of changes

# License

# Tests


test/metric_value_benchmark/agent_configs/statsd_config.json
```diff
@@ -1,13 +1,13 @@
...
-        "namespace": "statsd_test",
+        "namespace": "MetricValueBenchmarkTest",
         "append_dimensions": {
             "InstanceId": "${aws:InstanceId}"
         },
         "metrics_collected": {
             "statsd": {
                 "metrics_aggregation_interval": 30,
-                "metrics_collection_interval": 5,
+                "metrics_collection_interval": 1,
```
khanhntd, line 10: `metrics_collection_interval` will affect the value of `statsd` since `statsd` are using [cache](https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/inputs/statsd/statsd.go#L75-L81) to aggregate. Therefore, `counter` in this case will have the value of 5 ([since we are sending every second with statsd](https://github.com/aws/amazon-cloudwatch-agent-test/pull/178/files#diff-68bb589fd3a05343cba4bfe377a4be5a995fd4e48a9e8ff817f3c69b5b688dceR52)) if the `metrics_collection_interval` is 5

test/metric_value_benchmark/statsd_test.go
```diff
@@ -6,62 +6,25 @@
...
-	"github.com/cactus/go-statsd-client/v5/statsd"
 )
 
 var _ test_runner.ITestRunner = (*StatsdTestRunner)(nil)
+
 type StatsdTestRunner struct {
 	test_runner.BaseTestRunner
 }
 
-// stopChan is for stopping the goroutine generating statsd metrics.
```
khanhntd, line 26: Deleting this since the `metric_value_benchmark` are sharing the resources to each other. Therefore, the resources need to be unique within each file. Instead of generate statsd metrics here, use a common function [sendStatdMetrics](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/internal/common/metrics.go#L86-L118) to send the required metrics. The long term would replace the metric [value benchmark with the validator](https://github.com/aws/amazon-cloudwatch-agent-test/tree/main/validator)

ymtaye (approved):
Looks good to me!

internal/common/metrics.go
```diff
@@ -49,25 +49,36 @@ func sendCollectDMetrics(metricPerMinute int, duration time.Duration) error {
...
 	defer ticker.Stop()
 	endTimeout := time.After(duration)
 
 	for {
 		select {
 		case <-ticker.C:
-			for t := 0; t < metricPerMinute; t++ {
-				err = client.Write(ctx, &api.ValueList{
+			for t := 1; t <= metricPerMinute/2; t++ {
+				_ = client.Write(ctx, &api.ValueList{
```
SaxyPandaBear, line 60: why are we ignoring this err now

test/metric_value_benchmark/collectd_test.go
```diff
@@ -72,41 +59,58 @@ func (t *CollectDTestRunner) validateCollectDMetric(metricName string) status.Te
...
+		})
+	}
 
+	dims, failed := t.DimensionFactory.GetDimensions(instructions)
 	if len(failed) > 0 {
 		return testResult
 	}
-
 	fetcher := metric.MetricValueFetcher{}
+	// Namespace must match the JSON config.
```
SaxyPandaBear, line 88: This is a given. We don't need this comment. It should have been caught at review time before.

test/metric_value_benchmark/collectd_test.go
```diff
@@ -72,41 +59,58 @@ func (t *CollectDTestRunner) validateCollectDMetric(metricName string) status.Te
...
 
-	if !isAllValuesGreaterThanOrEqualToZero(metricName, values) {
+	runDuration := t.GetAgentRunDuration()
+	// aggregationInterval must match the JSON configuration.
+	aggregationInterval := 30 * time.Second
+	// If aggregation is not happening there could be a data point every 5 seconds.
+	// So validate the upper bound.
+	upperBound := int(runDuration/aggregationInterval) + 2
+	// Allow 2 missing data points in case CW-Metrics-Web-Service has a 1 minute
+	// delay to store.
```
SaxyPandaBear, line 102: ? This is not the SLA for metric put-to-get. We should not need this amount of variance.

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -120,17 +119,33 @@ func shouldRunEC2Test(env *environment.MetaData, t *test_runner.TestRunner) bool
...
+			log.Printf("Values are not all greater than or equal to zero for %s", metricName)
 			return false
 		}
+		totalSum += value
+	}
+	metricErrorBound := 0.1
+	metricAverageValue := totalSum / float64(len(values))
+	upperBoundValue := expectedValue * (1 + metricErrorBound)
+	lowerBoundValue := expectedValue * (1 - metricErrorBound)
+	if expectedValue > 0 && (metricAverageValue > upperBoundValue || metricAverageValue < lowerBoundValue) {
```
SaxyPandaBear, line 144: Why are we allowing for this variance?

SaxyPandaBear (approved):

SaxyPandaBear:
I don't like that we need to introduce CGO for this.

test/metric_value_benchmark/metrics_value_benchmark_test.go
khanhntd: Glad you asked !!! CWA has two kinds of inputs:
* Normal inputs for host resources metrics (e.g `cpu`, `mem`)
* Service inputs for (e.g `stasd`, `collectd`)

For the first case, if we ever want to validate the host resources which we would tied up to the instances and will be flaky to do so. Therefore, we cannot validate exactly the value that tied to the instance size

For the second case, in case such as `statsd` has incremented value when being a counter or a sampling (sampling of counter), if the collection interval is 5 second, the counter will have the value of 5. However, this won't be true all the time since `counter` can be delayed when dealing with cache (e.g the counter value in the first 5 second will be 4 and the counter for the next 5 second will fill up the lost amount which is 6). So it will flaky too. Therefore, it would be flaky if consider high resolution metrics

internal/common/metrics.go
khanhntd: Since the [second write](https://github.com/aws/amazon-cloudwatch-agent-test/pull/178/files/583f794c60c318695351896820d9ec205933630e#diff-781bb3810ede78771b4deb63bb6bb797a2d06434f0d844429b5ad16a20304e0fR71-R81) will handle the error and if its failed the second time, then it would failed the first time. In case such as the [buffer being filled up](https://github.com/aws/amazon-cloudwatch-agent-test/pull/178/files/583f794c60c318695351896820d9ec205933630e#diff-781bb3810ede78771b4deb63bb6bb797a2d06434f0d844429b5ad16a20304e0fR82-R84), the second one would be the culprit for filling the disk up so we don't need to validate the first write.

test/metric_value_benchmark/collectd_test.go
khanhntd: Got cha. Thanks for your comments on it !!!

test/metric_value_benchmark/collectd_test.go
khanhntd: In high resources consumption environment, I would disagree with you; however, if we only running to validate then I agree this should be acceptable

khanhntd:
> I don't like that we need to introduce CGO for this.

Sorry for this. The amount of changes are small and this could help improving the performance/stress so I thought I could add it.

test/metric_value_benchmark/collectd_test.go
SaxyPandaBear: No. You're talking about how long it takes to get a response from CloudWatch. What is the resource constrained environment?

test/metric_value_benchmark/metrics_value_benchmark_test.go
SaxyPandaBear: That is an implementation detail for the individual test runners. Changing the entire implementation across all of the test runners is unnecessary and changes our validation behavior.

khanhntd: Hmm ! I would prefer generalization  that applied to all plugins but you are right on the fact that 
```
That is an implementation detail for the individual test runners.
```

PR #104, opened 2023-01-27
First acceptance test

ChenaLee:
# Description of the issue
Need to test that agent config file can only be modified by root.

# Description of changes
Added a new security test dir and added the first test. 
This is used in a new workflow created : https://github.com/aws/amazon-cloudwatch-agent/pull/675

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
All tests pass: https://github.com/ChenaLee/amazon-cloudwatch-agent/actions/runs/4018230002

PR #179, opened 2023-03-22
First acceptance test (authored by @chenalee)

williazz:
# Description of the issue
* Finishing up #104 after ownership was transferred from @ChenaLee to me
* Depends on related [PR #724 in aws/amazon-cloudwatch-agent](https://github.com/aws/amazon-cloudwatch-agent/pull/724)

# Description of changes
Please read #104 for context

Only finishing touches I made were
* refactored filepermissions_test using functional paradigm
* removed rule/conditions interface
* remove use of testrunner so that `filepermissions_test.go` can be invoked separately
* fix build failures on windows
* replace deprecated `syscall` package with `sys/unix` and `sys/windows`
* resolve merge conflicts / nits

# License

# Tests
* build checks
* [Ran on fork](https://github.com/williazz/amazon-cloudwatch-agent/actions/runs/4601202126)
* `ca_bundle` and `ECSFargateIntegrationTest` are broken on my fork


williazz:
Acceptance test isn't running because it depends on unrelated ca_bundle and ECSFargateIntegrationTest are failing. Need to fix those

khanhntd (approved):
For the workflow, overall if we only concern the functionality, it would be LGTM from me. But if talking about scalability issue, then IMO there would be lots of thing can be improved (e.g for `nvidia_gpu` which extracts metrics from a binary , you cannot validate the metrics but only able to do basic acceptance testing; so you need to create two testing for `nvidia_gpu` - one for metrics and one for acceptance)

internal/rule/rule.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"log"
+)
+
+type Rule[T any] struct {
+	Conditions []ICondition[T]
+}
+
+func (r *Rule[T]) Evaluate(target T) (bool, error) {
```
sethAmazon, line 14: This looks like the validator pattern. I always felt the validator pattern is confusing and should only be used if there are many validations. If not, just have a couple of functions that execute one after the other.

sethAmazon:
Can you go test this on an ex2 instance?

test/acceptance/acceptance_test.go
```diff
@@ -0,0 +1,57 @@
...
+func init() {
+	environment.RegisterEnvironmentMetaDataFlags(envMetaDataStrings)
+}
+
+type AcceptanceTestSuite struct {
+	suite.Suite
+	test_runner.TestSuite
+}
+
+var _ test_runner.ITestSuite = (*AcceptanceTestSuite)(nil)
```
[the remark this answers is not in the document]
williazz: @ChenaLee created a test suite so we can add more security tests in the future, iirc. I think it makes sense

williazz:
@sethAmazon I wouldn't be opposed to running it in parallel. @ChenaLee provided [her reasoning here](https://github.com/aws/amazon-cloudwatch-agent/pull/724#discussion_r1149377548) and I think I'm okay with it, so I left it alone

filesystem/unix_permission.go
```diff
@@ -7,43 +7,99 @@ package filesystem
...
+	return fmt.Errorf("file's owner does not have enough permission at path %s", filePath)
 }
 
 // CheckFileOwnerRights check that the given owner is the same owner of the given filepath
 func CheckFileOwnerRights(filePath, requiredOwner string) error {
-	var stat syscall.Stat_t
-	if err := syscall.Stat(filePath, &stat); err != nil {
-		return fmt.Errorf("Cannot get file's stat %s: %v", filePath, err)
-	}
+	ownerUsername, err := GetFileOwnerUserName(filePath)
```
adam-mateen, line 97: good reuse

filesystem/unix_permission.go
```diff
@@ -7,43 +7,99 @@ package filesystem
...
 
-	if owner, err := user.LookupId(fmt.Sprintf("%d", stat.Uid)); err != nil {
-		return fmt.Errorf("Cannot look up file owner's name %s: %v", filePath, err)
-	} else if owner.Name != requiredOwner {
-		return fmt.Errorf("Agent does not have permission to protect file %s", filePath)
+	if err != nil {
+		return fmt.Errorf("cannot look up file owner's name %s: %v", filePath, err)
```
adam-mateen, line 100: I have a general question about bubbling up errors in Golang.
If the called function returns an error, should the caller just return that error, or returned a modified/appended error message?

For example line 100 could have just been `return err`

internal/rule/unix_rules.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"github.com/aws/amazon-cloudwatch-agent-test/filesystem"
+	"log"
+)
+
+type FilePermissionExpected struct {
+	PermissionCompared filesystem.FilePermission
+	ShouldExist        bool
+}
```
adam-mateen, line 16: To make it easier for developers to tell which interface(s) you intend to implement it helps to add a line like:
```
// Verify Foo implements Bar
var _ MyPkg.Bar = (*Foo)(nil)
```

internal/rule/unix_rules.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if e.ShouldExist {
+		return has, nil
+	} else {
+		return !has, nil
+	}
+}
+
+var _ ICondition[string] = (*FilePermissionExpected)(nil)
```
adam-mateen, line 35: I recommend putting this line just before or after the definition of `FilePermissionExpected`.

internal/rule/unix_rules.go
```diff
@@ -0,0 +1,70 @@
...
+	has, err := filesystem.FileHasPermission(target, e.PermissionCompared)
+	if err != nil {
+		return false, err
+	}
+
+	if e.ShouldExist {
+		return has, nil
+	} else {
+		return !has, nil
+	}
```
adam-mateen, line 32: nit:
```
	if e.ShouldExist {
		return has, nil
	} else {
		return !has, nil
	}
```
could just be:
```
return e.shouldExist == has, nil
```
right?

test/acceptance/acceptance_test.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, testRunner := range getEc2TestRunners() {
+		testRunner.Run(suite)
+	}
+
+	suite.Assert().Equal(status.SUCCESSFUL, suite.Result.GetStatus(), "Security Test Suite Failed")
+}
+
+func TestAcceptanceTestSuite(t *testing.T) {
+	suite.Run(t, new(AcceptanceTestSuite))
+}
```
adam-mateen, line 57: It would be a less confusing to follow the Golang defined mechanisms for testing.
As in just put each test case in a function named `TestXXX`, e.g. `TestFilePermission`, or `Test

1 big problem with the current "test suite" is that developers cannot easily run a single test case on their local ec2 instance. I very much want to just SSH into my EC2 instance and run:
`go test packageName -run NameOfTest`


Today, if I want to run the "cpu" related integration tests I must comment out code from the test suite.
It isn't a huge problem, but I would like to avoid the same problem again.


Can you think of any ways to simplify the implementation down to a set of test cases that can be individually run without modifying test code? e.g. `go test packageName -run NameOfTest`. 

If we make our test look and behave like other go tests it will be easier for future devs to understand.

test/acceptance/acceptance_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+type AcceptanceTestSuite struct {
+	suite.Suite
+	test_runner.TestSuite
+}
+
+var _ test_runner.ITestSuite = (*AcceptanceTestSuite)(nil)
+
+func (suite *AcceptanceTestSuite) GetSuiteName() string {
+	return "SecurityTestSuite"
```
adam-mateen, line 31: Should the name be `AcceptanceTestSuite`?
IIRC there was push back about using the word "security", as if it was _taboo_.

test/acceptance/acceptance_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+var (
+	ec2TestRunners []*test_runner.TestRunner
+)
+
+func getEc2TestRunners() []*test_runner.TestRunner {
+	if ec2TestRunners == nil {
```
adam-mateen, line 39: Why is the `== nil` check needed? What would happen if you just returned a new slice with the FilePermissionTestRunner?

generator/resources/ec2_acceptance_test_matrix.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "os": "ubuntu-20.04",
+    "username": "ubuntu",
+    "instanceType":"t3a.medium",
+    "installAgentCommand": "go run ./install/install_agent.go deb",
+    "ami": "cloudwatch-agent-integration-test-ubuntu*",
+    "caCertPath": "/etc/ssl/certs/ca-certificates.crt",
+    "arc": "amd64",
```
adam-mateen, line 9: Typically the abbreviation is "arch", I do not expect you to change it in this PR though.

test/acceptance/testrunners/filepermissiontestrunner.go
```diff
@@ -0,0 +1,121 @@
...
+func (m *FilePermissionTestRunner) GetAgentConfigFileName() string {
+	return "minimum_config.json"
+}
+
+func (m *FilePermissionTestRunner) GetMeasuredMetrics() []string {
+	return []string{}
+}
+
+func (m *FilePermissionTestRunner) GetAgentRunDuration() time.Duration {
+	return 1 * time.Minute
```
adam-mateen, line 76: You could probably shorten this down to 1 second.
The Agent does not need to run at all right?
This test is just verifying the file permissions after install.

test/acceptance/testrunners/filepermissiontestrunner.go
```diff
@@ -0,0 +1,121 @@
...
+
+var testCases = map[string]rule.Rule[string]{
+	agentConfigPath:      onlyRootCanWriteRule,
+	agentConfigCopiedDir: onlyCWAgentCanWriteRule,
+	translatedTomlPath:   onlyRootCanWriteRule,
+}
+
+var testGroupResult *status.TestGroupResult = nil
+
+func (m *FilePermissionTestRunner) Validate() status.TestGroupResult {
```
adam-mateen, line 55: Shouldn't the validation be done in the function called `Validate()`?
Why is `validatePermissions` called in `SetupAfterAgentRun()`?

test/acceptance/testrunners/filepermissiontestrunner.go
```diff
@@ -0,0 +1,121 @@
...
+func (m *FilePermissionTestRunner) SetupAfterAgentRun() error {
+	testResults := make([]status.TestResult, len(testCases))
+
+	count := 0
+	for k, v := range testCases {
+		testResults[count] = m.validatePermissions(k, v)
+		count++
+	}
+
+	testGroupResult = &status.TestGroupResult{
```
adam-mateen, line 88: The global variable `testGroupResult` should not be needed to pass info between `SetupAfterAgentRun()` and `Validate()`. I think the call to `validatePermissions()` could just be done in `Validate()`.

test/acceptance/testrunners/filepermissiontestrunner.go
```diff
@@ -0,0 +1,121 @@
...
+			&rule.PermittedEntityMatch{ExpectedOwner: aws.String("root"), ExpectedGroup: aws.String("root")},
+			&rule.FilePermissionExpected{PermissionCompared: filesystem.OwnerWrite, ShouldExist: true},
+			&rule.FilePermissionExpected{PermissionCompared: filesystem.AnyoneWrite, ShouldExist: false},
+		},
+	}
+	onlyCWAgentCanWriteRule = rule.Rule[string]{
+		Conditions: []rule.ICondition[string]{
+			&rule.PermittedEntityMatch{ExpectedOwner: aws.String("cwagent"), ExpectedGroup: aws.String("cwagent")},
+			&rule.FilePermissionExpected{PermissionCompared: filesystem.OwnerWrite, ShouldExist: true},
+			&rule.FilePermissionExpected{PermissionCompared: filesystem.AnyoneWrite, ShouldExist: false},
```
adam-mateen, line 42: Seth is right, this does just look like 3 function calls with different inputs.
Would it be simpler to remove the Rule type and Condition interface and just have 2 common functions `CheckPermittedEntityMatch()` and `CheckFilePermission()`?

filesystem/unix_permission.go
williazz: I had the same question actually. `Error handling and go` doc suggests to keep `error` separate from `message` and any other important information because `error` alone probably won't tell you enough. For example: 

```go
type appError struct {
    Error   error
    Message string
    Code    int
}
```

test/acceptance/acceptance_test.go
williazz: yes good catch, this is an old string that needs to be updated

test/acceptance/acceptance_test.go
williazz: Nothing, it just caches the test runners so it doesn't have to be created again. I guess the function isn't pure anymore, so just depends on what design pattern we have agreed on

test/acceptance/acceptance_test.go
williazz: thinking about how to address this

filesystem/unix_permission.go
williazz: I think that we should standardize how we format error messages in a separate PR. I agree with your callout

internal/rule/rule.go
ChenaLee: (I didn't even know there was a name to this but) What part is confusing, and how does that affect users of the interface? Why is having many validators make it better/different? Theoretically you can just keep adding separate functions that execute one after another no matter how many there are. Is it worth deleting interface that is already better prepped for more usecases and gives intuitive names to users?

test/acceptance/testrunners/filepermissiontestrunner.go
ChenaLee: Yeah what was initially not considered in the very beginning was that not all validations will happen via calling CW backend. iirc some of the files(?) we need to validate no longer existed after stopping agent.

williazz: yeah i just encountered this issue, i'm moving it back and leaving a comment

williazz: looks like this line is the culprit 

https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/test_runner/base_test_runner.go#L108

test/acceptance/testrunners/filepermissiontestrunner.go
williazz: validatePermissions must be done during SetupAfterAgentRun() before config.json gets deleted

filesystem/unix_permission.go
```diff
@@ -7,43 +7,100 @@ package filesystem
...
+	GroupWrite:  unix.S_IWGRP,
+	AnyoneWrite: unix.S_IWOTH,
+	OwnerRead:   unix.S_IRUSR,
+	AnyoneRead:  unix.S_IROTH,
+}
+
+func FileHasPermission(filePath string, permission FilePermission) (bool, error) {
+	fileStat, ok := GetFileStatPermission(filePath)
+	if ok != nil {
+		return false, ok
```
adam-mateen, line 35: ok is a confusing variable name. I would expect `ok` to be a bool and `true` to be the happy path.

I suggest:
```
fileStat, err := GetFileStatPermission(filePath)
if err != nil {
    return false, err
}
```

filesystem/unix_permission.go
```diff
@@ -7,43 +7,100 @@ package filesystem
...
+}
+
 // CheckFileRights check that the given file path has been protected by the owner.
 // If the owner is changed, they need at least the sudo permission to override the owner.
 func CheckFileRights(filePath string) error {
-	var stat syscall.Stat_t
-	if err := syscall.Stat(filePath, &stat); err != nil {
-		return fmt.Errorf("Cannot get file's stat %s: %v", filePath, err)
+	var stat unix.Stat_t
+	if err := unix.Stat(filePath, &stat); err != nil {
```
adam-mateen, line 79: hooray for removing a bit of syscall usage! 😎

test/acceptance/agent_configs/minimum_config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "agent": {
+    "run_as_user": "cwagent"
+  },
+  "metrics": {
+    "metrics_collected": {
+      "mem": {
+        "measurement": [
+          "mem_used_percent"
+        ]
```
adam-mateen, line 10: Beyond just configs, consider that installing the agent using the control script also also allows a user to download the configuration file from SSM.
It might be valuable to test that the control script puts the downloaded JSON config in the right location, with the right permissions and owner.
(If it is outside the scope of this test, consider opening another ticket to track that work.)

test/acceptance/filepermissions_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+func createGroupResult() status.TestGroupResult {
+	testResults := make([]status.TestResult, 3)
+	testGroupResult := status.TestGroupResult{
+		Name:        "FilePermissions",
+		TestResults: testResults,
+	}
+	testResults[0] = onlyUserCanWriteToFilepath("root", agentConfigPath)
```
adam-mateen, line 53: Users create the JSON configuration files and they can put them wherever they want. They specify the path to the JSON config when using the control script... So isn't it possible that `/opt/aws/amazon-cloudwatch-agent/bin/config.json` doesn't exist and instead the file is at `/tmp/adam.json`? Isn't it also acceptable that the owner is not `root`?

test/acceptance/filepermissions_test.go
```diff
@@ -0,0 +1,114 @@
...
+	testResults[1] = onlyUserCanWriteToFilepath("cwagent", agentConfigCopiedDir)
+	testResults[2] = onlyUserCanWriteToFilepath("root", translatedTomlPath)
```
williazz, line 55: This test is failing for some reason. Somehow, when I refactored this test, the `toml` file owner changed from `root` to `cwagent`. I went through line by line and couldn't understand why this behavior changed

adam-mateen (approved):

test/acceptance/filepermissions_test.go
williazz: Yeah that was my thinking as well, but that doesn't explain the behavior change. I'm ok changing it though

filesystem/unix_permission.go
williazz: good catch, i remember making the same change but i think i overrode it somehow

test/acceptance/filepermissions_test.go
williazz: anyways i fixed it

test/acceptance/filepermissions_test.go
```diff
@@ -0,0 +1,118 @@
...
+		log.Printf("FAILED fileHasOwnerAndGroup(); expected GROUP=%v; actual=%v", expectedGroup, group)
+		return false, nil
+	}
+	return true, nil
+}
+
+func fileHasPermission(filepath string, permission filesystem.FilePermission, shouldHavePermission bool) (bool, error) {
+	hasPermission, err := filesystem.FileHasPermission(filepath, permission)
+	if err != nil {
+		return false, fmt.Errorf("fileHasPermission(), could not invoke filesystem.FileHasPermission; err=%v", err)
```
jefchien, line 115: nit: Looks like we log the error in the other functions. Would it make sense to log it here as well?

jefchien (approved):

test/acceptance/filepermissions_test.go
```diff
@@ -0,0 +1,118 @@
...
+	testResult.Status = status.SUCCESSFUL
+	return testResult
+}
+
+func fileHasOwnerAndGroup(filepath, expectedOwner, expectedGroup string) (bool, error) {
+	owner, err := filesystem.GetFileOwnerUserName(filepath)
+	if err != nil {
+		err := fmt.Errorf("FAILED fileHasOwnerAndGroup(); error=%v", err)
+		log.Println(err)
+		return false, err
```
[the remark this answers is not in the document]
williazz: I think it's important to log an error if the GetFileOwnerUserName() or GetFileGroupName() fails with error, but yeah it's a good callout that we don't actually use that error anywhere. I'll add some logic for it

test/acceptance/filepermissions_test.go
williazz: good catch. I think I had some trouble thinking about when to log vs return an error

test/acceptance/filepermissions_test.go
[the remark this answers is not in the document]
williazz: After some thinking, `onlyUserCanWriteToFilepath()` should not care whether `fileHasOwnerAndGroup()` or `fileHasPermission()` fails due to internal error or incorrect values. However, I think we should still keep the current format because it logs the information needed for debugging.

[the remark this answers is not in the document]
jefchien: Sure, I'm just saying you probably don't need the boolean since an error can be treated as a failure.

PR #195, opened 2023-04-11
Canary repeat downloads

adam-mateen:
# Description of the issue
The issue is that the canary test is not testing frequently enough to detect issues before customers do.
We want to test downloading, installing, and starting the agent every 5 minutes instead of just once a day (and just once at release time).

# Description of changes
Assuming there is a canary test host which is re deployed via github action once a day and when a new agent version is released. The change is to make the test create a CRON job to rerun itself every 5 minutes. And to report some custom metrics which indicate the type of failure.

I considered spawning a long lived process which would rerun the test, but that would not work if the process crashes, or the host gets rebooted. So future updated will be needed to make this canary test run on Windows (use "scheduled tasks"or a concept similar to CRON).

# License

# Tests
1. SSH into a canary test host (ec2 instance) and run the new test code:
```
[ec2-user@ip-172-31-49-128 amazon-cloudwatch-agent-test]$ AWS_REGION=us-west-2 go test ./test/canary/ -v -p 1 -c-agent-integration-test -bucket=private-cloudwatch-
=== RUN   TestCanary
2023/04/11 14:53:29 Testing all EC2 plugins
2023/04/11 14:53:35 Copy File resources/canary_config.json to /opt/aws/amazon-cloudwatch-agent/bin/config.json
2023/04/11 14:53:35 File resources/canary_config.json abs path /home/ec2-user/amazon-cloudwatch-agent-test/test/canary/resources/canary_config.json
2023/04/11 14:53:35 File : resources/canary_config.json copied to : /opt/aws/amazon-cloudwatch-agent/bin/config.json
2023/04/11 14:53:35 Agent has started
2023/04/11 14:53:35 Testing all EC2 plugins
2023/04/11 14:53:35 Testing all EC2 plugins
2023/04/11 14:53:35 Copy File resources/canary_test_cron to /etc/cron.d/canary_test_cron
2023/04/11 14:53:35 File resources/canary_test_cron abs path /home/ec2-user/amazon-cloudwatch-agent-test/test/canary/resources/canary_test_cron
2023/04/11 14:53:35 File : resources/canary_test_cron copied to : /etc/cron.d/canary_test_cron
--- PASS: TestCanary (6.73s)
PASS
ok      github.com/aws/amazon-cloudwatch-agent-test/test/canary 6.738s
```
2. Verify the CRON job was created and loaded: `sudo tail /var/log/cron `
3. Verified new metrics show up in CloudWatch console:
Screen Shot 2023-04-11 at 10 16 58 AM

------------------

I also tested from my personal AWS account, which does not have the installer in the specific S3 bucket, and as expected the test fails and the `DownloadFail` metric is > 0 (depends on how many times I run the test per minute).
```
[ec2-user@ip-172-31-17-153 amazon-cloudwatch-agent-test]$ AWS_REGION=us-west-2 go test ./test/canary/ -v -p 1 -count=1 -computeType=EC2 -bucket=private-cloudwatch-agent-integration-test
=== RUN   TestCanary
2023/04/11 18:19:18 Testing all EC2 plugins
2023/04/11 18:19:19 error: downloading, operation error S3: GetObject, https response error StatusCode: 403, RequestID: KX3P2RWX2XCKD4Q8, HostID: BgARpqJJscr2tfZRZyMujonE4YWr8wKmMrd+q1OyXBeZxSv9xLeI8lmdAEZu8/fFdclGwzjKvtODdIUfmKRyAw==, api error AccessDenied: Access Denied
2023/04/11 18:19:19 error: name DownloadFail, err operation error S3: GetObject, https response error StatusCode: 403, RequestID: KX3P2RWX2XCKD4Q8, HostID: BgARpqJJscr2tfZRZyMujonE4YWr8wKmMrd+q1OyXBeZxSv9xLeI8lmdAEZu8/fFdclGwzjKvtODdIUfmKRyAw==, api error AccessDenied: Access Denied
    canary_test.go:69: 
        	Error Trace:	/home/ec2-user/go/src/github.com/aws/amazon-cloudwatch-agent-test/test/canary/canary_test.go:69
        	            				/home/ec2-user/go/src/github.com/aws/amazon-cloudwatch-agent-test/test/canary/canary_test.go:38
        	Error:      	Received unexpected error:
        	            	operation error S3: GetObject, https response error StatusCode: 403, RequestID: KX3P2RWX2XCKD4Q8, HostID: BgARpqJJscr2tfZRZyMujonE4YWr8wKmMrd+q1OyXBeZxSv9xLeI8lmdAEZu8/fFdclGwzjKvtODdIUfmKRyAw==, api error AccessDenied: Access Denied
        	Test:       	TestCanary
--- FAIL: TestCanary (0.56s)
FAIL
FAIL	github.com/aws/amazon-cloudwatch-agent-test/test/canary	0.568s
FAIL
```
Screen Shot 2023-04-11 at 1 24 22 PM

### DIFF
--- a/internal/common/agent_util_unix.go
+++ b/internal/common/agent_util_unix.go
@@ -84,7 +84,7 @@ func printOutputAndError(stdout []byte, err error) {
 	if ok {
 		stderr = string(ee.Stderr)
 	}
-	log.Printf("failed\n\tstdout:\n%v\n\tstderr:\n%v\n", string(stdout), stderr)
+	log.Printf("failed\n\tstdout:\n%s\n\tstderr:\n%s\n", string(stdout), stderr)
 }
 
 func UninstallAgent(pm PackageManager) error {
--- a/test/canary/canary_test.go
+++ b/test/canary/canary_test.go
@@ -60,7 +60,7 @@ func TestCanary(t *testing.T) {
 func reportMetric(t *testing.T, name string, err error) {
 	var v float64 = 0
 	if err != nil {
-		log.Printf("error: name %v, err %v", name, err)
+		log.Printf("error: name %s, err %s", name, err)
 		v += 1
 	}
 	require.NoError(t, awsservice.ReportMetric("CanaryTest", name, v,


test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
+	bucket := environment.GetEnvironmentMetaData(envMetaDataStrings).Bucket
+	// Need to create a temporary file at low privilege.
+	// Then use sudo to copy it to the CRON directory.
+	src := "resources/canary_test_cron"
+	updateCron(src, region, bucket)
+	dst := "/etc/cron.d/canary_test_cron"
+	common.CopyFile(src, dst)
+}
+
+func updateCron(filepath, region, bucket string) {
```
khanhntd, line 105: It would be better to avoid setup a `cron` or holding up an instance or directly involve the command on that specific instance. It would not be to generalize when coming to other platforms such as Windows OS  and it would be better to setup instance temporarily if you consider about patching too

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
 	environment.RegisterEnvironmentMetaDataFlags(envMetaDataStrings)
 }
 
-// This copies the canary config, starts the agent, and confirm version is correct.
-// Sanity test runs earlier in the terraform code execution
+// TestCanary verifies customers can download the agent, install it, and it "works".
+// Reports metrics on the number of download failures, install failures, and start failures.
 func TestCanary(t *testing.T) {
```
khanhntd, line 35: Why we would need to have the `downloaded_cwa.rpm` here if we can setup all the related things there and we only [need to provide the RPM](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/terraform/ec2/linux/main.tf#L66-L76)

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
-	// Canary set up
-	downloadVersionFile()
+	installerFilePath := "./downloaded_cwa.rpm"
+	err := downloadInstaller(installerFilePath)
+	reportMetric(t, "DownloadFail", err)
```
khanhntd, line 38: It would be better to leverage custom metrics we have (e.g statsd)

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
-	actualVersion, err := os.ReadFile(installAgentVersionPath)
-	if err != nil {
-		t.Fatalf("Failure reading installed version file %s err %v", installAgentVersionPath, err)
+	actualVersion, _ := os.ReadFile(installAgentVersionPath)
+	expectedVersion, _ := getVersionFromS3()
+	if expectedVersion != string(actualVersion) {
+		err = errors.New("agent version mismatch")
 	}
-	assert.Equal(t, string(expectedVersion), string(actualVersion))
+	reportMetric(t, "VersionFail", err)
```
khanhntd, line 55: It good we want to be robust but IMO, test should be generalize as much as possible (if these states failed, it would show the error in the github action console, then why we should look at these metrics?)

test/canary/canary_test.go
adam-mateen: I discussed this with the rest of QuickSilver. It is simpler and less cost to have 1 long lived instance that is continuously testing than to rely on GitHub actions to launch a new EC2 instance every 5 minutes.

test/canary/canary_test.go
adam-mateen: We want to repeatedly download from S3. Not just repeatedly install the already provided RPM.

test/canary/canary_test.go
khanhntd: ... I would disagree with that: that's what `on-demand` for and holding a instance is most cost than spinning up an instance

test/canary/canary_test.go
adam-mateen: I don't think that is better. If there is an issue with downloading, installing, or starting CloudWatch Agent then we cannot rely on sending statsd metrics to CloudWatch Agent as the mechanism to report such failures. We intentionally want to report metrics using AWS SDK, or AWS CLI, or something other than CWA itself.

internal/awsservice/s3.go
```diff
@@ -0,0 +1,33 @@
...
+	"log"
+	"os"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+
+func DownloadFile(bucket, key, outFilename string) error {
```
adam-mateen, line 16: DownloadFile is basically the same code that was already in the CanaryTest. I just made it generic and moved it to this `s3.go` file.

internal/common/agent_util_unix.go
```diff
@@ -66,6 +66,33 @@ func TouchFile(filePathAbsolute string) error {
...
 }
 
+func printOutputAndError(function string, stdout []byte, err error) {
+	if err == nil {
+		return
+	}
+	stderr := ""
+	ee, ok :=  err.(*exec.ExitError)
+	if ok {
+		stderr = string(ee.Stderr)
```
adam-mateen, line 76: This is kind of cool. `ExitError` is a subtype of `error`, and it has an extra field for the stderr.

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
+
+	setupCron()
```
adam-mateen, line 57: I am going to change this so the test fails if `setupCron()` fails.

adam-mateen, line 57: `CopyFile` calls `log.Fatal()` on error which would cause the test to fail.

test/canary/canary_test.go
adam-mateen: As far as I know the canary test is the only mechanism. We don't want to rely on customers contacting us via Slack or tickets.

SaxyPandaBear: What is the argument here? Launching an EC2 instance every 5 minutes has some amount of lag time, and really is just burdensome for no benefit. I think having a long-lived instance that pings is fine. What is this cron testing? Is it testing download, install and metric publication? I think that's acceptable.

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
+	common.CopyFile(src, dst)
+}
+
+func updateCron(filepath, region, bucket string) {
+	s := fmt.Sprintf("MAILTO=\"\"\n*/5 * * * * ec2-user cd /home/ec2-user/amazon-cloudwatch-agent-test && AWS_REGION=%s go test ./test/canary/ -v -p 1 -count=1 -computeType=EC2 -bucket=%s > ./cron_run.log\n", region, bucket)
+	b := []byte(s)
+	err := os.WriteFile(filepath, b, 0644)
 	if err != nil {
-		log.Fatalf("Can't get cloud agent version from s3 err %v", err)
+		log.Println("error: creating temp cron file")
```
jefchien, line 110: So, this failure won't notify us of an issue, but I guess the metrics that the canary publishes would be missing and we'd alarm on those.

test/canary/canary_test.go
jefchien: I think that's what we'd need to do. Right now the `DownloadFail = 1`, then none of the other metrics will get reported, so we'd need to count those as failures as well.

jefchien: We'd need to have some buffer for missing data during host replacements.

test/canary/canary_test.go
khanhntd: > What is the argument here? Launching an EC2 instance every 5 minutes has some amount of lag time, and really is just burdensome for no benefit.

Agree with you regarding the amount of lag time. However, spinning up a new instances and don't use cron would give us a general mechanism for handling all test case (e.g SSM validation, Windows, MacOs)

khanhntd: Then its better to decrease the amount of time (e.g from 1 days to 5 minute) for Github Action though?

test/canary/canary_test.go
adam-mateen: @SaxyPandaBear yes we will continue to alarm if infrastructure metrics are missing data. As @jefchien said, there is be an allowance for a few minutes of missing data due to the expected host replacement once a day.

The custom metrics like "DownloadFail" might not be necessary. I think they will make it easier to tell what exactly went wrong. And eventually we may want many canary hosts in different regions or running different operating systems. In which case a single alarm on the "DownloadFail" metric (no dimensions) would suffice. We would still want some way to alarm if the agent on 1 of the many canary hosts was missing data though...

internal/common/agent_util_unix.go
```diff
@@ -66,6 +66,33 @@ func TouchFile(filePathAbsolute string) error {
 	return nil
 }
 
+func printOutputAndError(function string, stdout []byte, err error) {
```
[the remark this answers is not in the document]
adam-mateen: My main goal was to avoid duplicating the code below:
```
	if err == nil {
		return
	}
	stderr := ""
	ee, ok :=  err.(*exec.ExitError)
	if ok {
		stderr = string(ee.Stderr)
```
in both `InstallAgent` and `UninstallAgent`.
Basically any function that calls `exec.Command()` followed by `c.Output()` could/should be logging both the stdout AND stderr.

I will rename the variable.

internal/common/agent_util_unix.go
```diff
@@ -66,6 +66,33 @@ func TouchFile(filePathAbsolute string) error {
...
+	if err == nil {
+		return
+	}
+	stderr := ""
+	ee, ok :=  err.(*exec.ExitError)
+	if ok {
+		stderr = string(ee.Stderr)
+	}
+	log.Printf("%v failed\n\toutput:\n%v\n\terror:\n%v\n",
```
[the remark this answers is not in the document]
adam-mateen: fixed

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
 }
 
-func downloadVersionFile() {
-	metadata := environment.GetEnvironmentMetaData(envMetaDataStrings)
-	ctx := context.Background()
-	c, err := config.LoadDefaultConfig(ctx)
+// reportMetric is just a helper to report a metric and conditionally fail
+// the test based on the error.
+func reportMetric(t *testing.T, name string, err error) {
+	var v float64 = 0
```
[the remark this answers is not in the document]
adam-mateen: I thought of doing that, but my general preference is simple code and explicit code. Sometimes those things contradict though. In this case I felt like it doesn't hurt to explicitly write the initial value, and it makes it more "readable" to devs, especially those not super familiar with Golang. Just my opinion though... If we have LINT rules that say otherwise I will happily conform. Or if anyone else +1 your comment :)

test/canary/canary_test.go
```diff
@@ -32,46 +30,83 @@ func init() {
...
+		v += 1
 	}
-	file, err := os.Create(downloadAgentVersionPath)
+	awsservice.ReportMetric("CanaryTest", name, v, types.StandardUnitCount)
+	require.NoError(t, err)
+}
+
+func downloadInstaller(filepath string) error {
+	bucket := environment.GetEnvironmentMetaData(envMetaDataStrings).Bucket
+	key := "release/amazon_linux/amd64/latest/amazon-cloudwatch-agent.rpm"
```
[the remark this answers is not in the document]
adam-mateen: fixed. Made the installer s3 key a cmd line option.

test/canary/canary_test.go
```diff
@@ -32,46 +30,82 @@ func init() {
...
+	defer setupCron()
+	installerFilePath := "./downloaded_cwa.rpm"
+	err := downloadInstaller(installerFilePath)
+	reportMetric(t, "DownloadFail", err)
+
+	// Don't care if uninstall fails. Agent might not be installed anyways.
+	_ = common.UninstallAgent()
+
+	err = common.InstallAgent(installerFilePath)
+	reportMetric(t, "InstallFail", err)
```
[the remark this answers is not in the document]
adam-mateen: `reportMetric` does not log any messages, just send a metric to cloudwatch. I fixed it so now the only logging is in, `printOutputAndError()`.

test/canary/canary_test.go
adam-mateen: WriteFile failing is very unlikely. And yes, if the CRON is not setup the metrics like `DownloadFail` would be missing data.

internal/common/agent_util_unix.go
```diff
@@ -66,6 +73,51 @@ func TouchFile(filePathAbsolute string) error {
...
+	out, err := c.Output()
+	printOutputAndError(out, err)
+	return err
+}
+
+// InstallAgent can determine the package manager based on the installer suffix.
+func InstallAgent(installerFilePath string) error {
+	log.Printf("Installing Agent...")
+	var c *exec.Cmd
+	// Assuming lower case
```
SaxyPandaBear, line 110: we could transform the path to be lowercase to enforce it, but I don't think it's super important to make that change

environment/metadata.go
```diff
@@ -47,6 +49,10 @@ func registerComputeType(dataString *MetaDataStrings) {
 func registerBucket(dataString *MetaDataStrings) {
 	flag.StringVar(&(dataString.Bucket), "bucket", "", "s3 bucket ex cloudwatch-agent-integration-bucket")
 }
+func registerS3Key(dataString *MetaDataStrings) {
+	flag.StringVar(&(dataString.S3Key), "s3key", "release/amazon_linux/amd64/latest/amazon-cloudwatch-agent.rpm",
```
adam-mateen, line 53: default value for the flag is the s3key for the RPM. DEB and MSI testing can override it in the GitHub Action code in the same way that the s3 bucket is passed in.

test/canary/canary_test.go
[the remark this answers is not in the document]
SaxyPandaBear: I'm indifferent. I will say that if I just saw `v := 0`, I will assume it is an int.

SaxyPandaBear (approved):

jefchien (approved):

test/canary/canary_test.go
jefchien: I think both will work, but khanh's convinced me that it makes more sense to just have the Github Actions run more often instead of using a cronjob.

PR #219, opened 2023-05-11
Add additional OS support and userdata test for integration test

zhihonl:
# Description of the issue
In order to fully migrate the base integration test from the internal pipeline, the following OS must be supported in the integration test

- Centos 7
- Centos 8 Stream
- AL2
- Debian 10
- Debian 11
- Oracle Linux 7
- Oracle Linux 8
- Oracle Linux 9
- SUSE 12
- SUSE 15
- Ubuntu 20
- Ubuntu 22
- Red Hat 7
- Red Hat 8
(Note: Internal integration tests still use some OS that are EOL therefore those OS are omitted from this change)

The public repo is also missing the integration test for installing and testing CWAgent through userdata.

# Description of changes
The following OS are now being supported for integration tests in addition to the existing ones:
- Centos 7
- Centos 8 Stream
- Oracle Linux 7
- Oracle Linux 8
- SUSE 12
- SUSE 15
- Red Hat 7

Userdata test is added for OL9 to follow the pattern of the same test within the internal integration test which only tests user data on oracle linux. A metric rename test has been added into the cpu_test within metric_value_benchmark by renaming "time_active" metric to "cpu_time_active_renamed".

# License

# Tests
A test run has been done on the [fork repo](https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/4941700970/jobs/8835908110). All tests are disabled except cloudwatchlogs,metric_value_benchmark and run_as_user to reduce test duration. The failed tests are caused by collectd failure which is out of scope for this PR. Note that SLES12 no longer supports collectd

**Revision 2**
[New test run](https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/4963307835/jobs/8882562742) to include install_agent.go as the installation method of CloudWatch agent. Also added command to install and run Cloudwatch Agent on launch instead of post-launch.


generator/resources/ec2_userdata_test_matrix.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "os": "ol9",
+        "username": "ec2-user",
+        "instanceType": "t3a.medium",
+        "installAgentCommand": "go run ./install/install_agent.go rpm",
```
SaxyPandaBear, line 6: How will this work? We're also installing the agent directly in user data

generator/resources/ec2_userdata_test_matrix.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "os": "ol9",
```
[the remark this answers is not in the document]
zhihonl: Since the this PR is specific to the migration of Linux test based on the internal integration test, supporting Windows userdata test is out of scope. However I do agree there should be more similar testing done on Windows as well.

generator/resources/ec2_userdata_test_matrix.json
zhihonl: Originally this line does not actually get executed since we are hard-coding the install command. However I made new change to run the install_agent.go script to install therefore it should address this comment.

terraform/ec2/userdata/install_agent.sh
```diff
@@ -0,0 +1,12 @@
...
+echo sha ${cwa_github_sha}
+echo clone and install agent
+cd /home/ec2-user/
+git clone --branch ${github_test_repo_branch} ${github_test_repo}
+cd amazon-cloudwatch-agent-test
+aws s3 cp s3://${binary_uri} .
+export HOME=/root
+export GOPATH=$HOME/go
+export PATH=$PATH:$GOPATH/bin
+rpm -U ./amazon-cloudwatch-agent.rpm
```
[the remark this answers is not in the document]
zhihonl: This is a good point. I added new changes to install and run agent using the specified config on launch. Then we will run the integration test without installing post-launch to imitate the customers.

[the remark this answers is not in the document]
zhihonl: > Also, not a problem for right now, but running `rpm` restricts which flavors of linux this works on.

I modified the script to run the install_agent.go script to install instead so we can handle all flavors of linux in the future.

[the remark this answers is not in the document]
SaxyPandaBear: That won't catch the issues that we want to catch with testing user data. Let's sync up next week to discuss what we actually want to test here.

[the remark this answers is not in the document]
zhihonl: This is the new test run that includes the above changes and they have passed. https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/4963307835/jobs/8882562742

test/metric_value_benchmark/agent_configs/collectd_config.json
```diff
@@ -1,10 +1,4 @@
 {
-    "agent": {
```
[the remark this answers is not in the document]
zhihonl: This was done by mistake. Just added it back in.

test/test_runner/base_test_runner.go
```diff
@@ -57,10 +57,29 @@ func (t *BaseTestRunner) GetAgentRunDuration() time.Duration {
...
+		Name: t.TestRunner.GetTestName(),
+		TestResults: []status.TestResult{
+			{
+				Name:   "Agent Started",
+				Status: status.SUCCESSFUL,
+			},
+		},
+	}
+	err := error(nil)
+	if testName == "Userdata" {
```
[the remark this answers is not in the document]
zhihonl: I think it will be redundant since we already assign test name to each test.

generator/test_case_generator.go
```diff
@@ -76,6 +76,13 @@ var testTypeToTestDirMap = map[string][]string{
 	"ec2_acceptance": {
 		"./test/acceptance",
 	},
+	"ec2_userdata": {
+		"./test/cloudwatchlogs",
+		"./test/metric_value_benchmark",
+		"./test/run_as_user",
+		"./test/userdata",
```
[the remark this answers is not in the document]
zhihonl: The only difference is userdata pushes metric to a namespace called UserDataTest. This is to distinguish userdata CWAgent test by verifying that CWAgent can in fact start on launch using userdata. If just using the metric_value_benchmark it can get mixed up with the post-launch behavior instead since they will be under same namespace.

test/test_runner/base_test_runner.go
```diff
@@ -57,10 +57,29 @@ func (t *BaseTestRunner) GetAgentRunDuration() time.Duration {
...
 		testGroupResult = t.TestRunner.Validate()
+	} else{
+		testGroupResult, err = t.runAgent()
```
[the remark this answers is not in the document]
zhihonl: Take netstat_test as an example, its GetTestName() is defined as following
```
func (t *NetStatTestRunner) GetTestName() string {
	return "NetStat"
}
```
And since `Run` method used in base_test_runner has this line `testName := t.TestRunner.GetTestName()`, the if/else statement will go to else because the testName is NetStat

test/userdata/agent_configs/cpu_config.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "metrics": {
+    "namespace": "UserDataTest",
+    "append_dimensions": {
+      "InstanceId": "${aws:InstanceId}"
+    },
+    "metrics_collected": {
+      "cpu": {
+        "measurement": [
+          {"name": "time_active", "rename":"cpu_time_active_renamed"},
```
SaxyPandaBear, line 15: We don't need to test the full set of CPU metrics. The purpose of this test is to be a smoke test to ensure that we can install and run the agent e2e when installed and configured via user data scripts. I think we can trim down what we validate here. I think the rename is a nice touch.

generator/resources/ec2_userdata_test_matrix.json
SaxyPandaBear: This command gets executed in the terraform after the instance is created but before the integration test. Take a look at `terraform/ec2/linux/main.tf`. What you have, you will launch the agent via user data, and then the test will go and replace it. This value needs to be changed. It should not run installation here.

terraform/ec2/userdata/install_agent.sh
```diff
@@ -0,0 +1,20 @@
...
+export GOPATH=$HOME/go
+export PATH=$PATH:$GOPATH/bin
+echo Running command ${install_agent}
+${install_agent}
+cd /home/ec2-user/amazon-cloudwatch-agent-test/test/userdata/agent_configs/
+echo Starting agent now
+/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s -c file:cpu_config.json
+echo Let agent run for 60 seconds
+sleep 60
+echo Agent finished run for 60 seconds
```
SaxyPandaBear, line 18: Don't think this is necessary. This just slows down the test. We aren't validating that the agent ran successfully or anything. Also, if the ctl script fails, that should fail the EC2 instance launch anyway.

test/test_runner/base_test_runner.go
```diff
@@ -57,10 +57,29 @@ func (t *BaseTestRunner) GetAgentRunDuration() time.Duration {
...
+	testGroupResult := status.TestGroupResult{
+		Name: t.TestRunner.GetTestName(),
+		TestResults: []status.TestResult{
+			{
+				Name:   "Agent Started",
+				Status: status.SUCCESSFUL,
+			},
+		},
+	}
+	err := error(nil)
```
SaxyPandaBear, line 73:
```suggestion
	var err error
```

test/test_runner/base_test_runner.go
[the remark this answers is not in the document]
SaxyPandaBear: We should not rely on test name here. There is no validation that we use unique test names when writing the test runners. What we should do is update the test runner interface to basically turn the `runAgent` function into a no-op. Then we can use a default value for all of the other tests, and then update it just in the user data test runner implementation. 

There should not need to be any changes in the `Run` function. I want us to try to push everything into the test runner implementation.

test/userdata/metrics_value_benchmark_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/aws/amazon-cloudwatch-agent-test/environment"
+	"github.com/aws/amazon-cloudwatch-agent-test/environment/computetype"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/metric/dimension"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/status"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/test_runner"
+)
+
+const namespace = "UserDataTest"
+
+type MetricBenchmarkTestSuite struct {
```
SaxyPandaBear, line 25: I really dislike all of the boilerplate code that we are copying throughout the project. This will come back and bite us eventually.

test/userdata/metrics_value_benchmark_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	_, ok := env.EC2PluginTests[strings.ToLower(t.TestRunner.GetTestName())]
+	return ok
+}
+
+// isAllValuesGreaterThanOrEqualToExpectedValue will compare if the given array is larger than 0
+// and check if the average value for the array is not la
+// TODO: Moving metric_value_benchmark to validator
+// https://github.com/aws/amazon-cloudwatch-agent-test/pull/162
+func isAllValuesGreaterThanOrEqualToExpectedValue(metricName string, values []float64, expectedValue float64) bool {
```
SaxyPandaBear, line 96: We should not be duplicating this from the other file. Please pull this function out into a common place so we can use it everywhere.

test/userdata/metrics_value_benchmark_test.go
```diff
@@ -0,0 +1,121 @@
...
+	suite.Run(t, new(MetricBenchmarkTestSuite))
+}
+
+func shouldRunEC2Test(env *environment.MetaData, t *test_runner.TestRunner) bool {
+	if env.EC2PluginTests == nil {
+		return true // default behavior is to run all tests
+	}
+	_, ok := env.EC2PluginTests[strings.ToLower(t.TestRunner.GetTestName())]
+	return ok
+}
```
SaxyPandaBear, line 90: Don't think you need this. I did this originally to try to help speed up test iteration for individual EC2 tests. But this is an entirely different suite that is running.

test/test_runner/base_test_runner.go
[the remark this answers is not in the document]
sky333999: Ok I understand now but I think for someone looking at this for the first time, its going to be extremely hard to correlate this if condition to the line [here](https://github.com/aws/amazon-cloudwatch-agent-test/pull/219/files#diff-21f80332df902cb88dd8d816f1fb3c758d8884f0a30b2054a44ec3c91044bb1bR39). Especially when these files are across folders (Ideally we should move this base_test_runner to some common place but restructuring this whole project probably needs to be a separate task).

I think what would be better is to add a method along the lines of `ShouldRunAgent()` in the base test runner that defaults to true and override this with a false just for the UserData implementation.

test/userdata/agent_configs/cpu_config.json
```diff
@@ -0,0 +1,26 @@
+{
```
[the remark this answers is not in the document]
sky333999: Cant we parameterize the namespace?

generator/resources/ec2_userdata_test_matrix.json
sky333999: I was very confused about this to but I think here, the same var is being passed to the shell script [here](https://github.com/aws/amazon-cloudwatch-agent-test/pull/219/files#diff-21f80332df902cb88dd8d816f1fb3c758d8884f0a30b2054a44ec3c91044bb1bR39) and the remote-exec part to install agent via terraform after init/before integ test is removed.
So should be fine *I think* - but confusing.

generator/test_case_generator.go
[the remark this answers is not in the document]
sky333999: But when the "./test/metric_value_benchmark" suite is being run, none of the tests in that suite match your `Userdata` if condition logic right?

test/userdata/metrics_value_benchmark_test.go
SaxyPandaBear: There are a couple PRs that do this so I won't block on it here

test/userdata/metrics_value_benchmark_test.go
zhihonl: Removed this in new PR

terraform/ec2/userdata/install_agent.sh
zhihonl: Original thought was that we need time for the agent to run but you are right, this just slows down the test. I removed these lines in new PR

test/userdata/agent_configs/cpu_config.json
zhihonl: I removed all other CPU tests and kept it to `time_active` only

test/test_runner/base_test_runner.go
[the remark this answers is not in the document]
zhihonl: I implemented the change that Andrew asked for. Now we have a default method for RunAgent() and each TestRunner can override if needed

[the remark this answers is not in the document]
zhihonl: https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/5070162706
This is the test run with the mentioned changes. Failed runs are due to collectd issues.

internal/common/metrics.go
```diff
@@ -211,3 +212,34 @@ func SendEMFMetrics(metricPerInterval int, metricLogGroup, metricNamespace strin
 		}
 	}
 }
+
+// isAllValuesGreaterThanOrEqualToExpectedValue will compare if the given array is larger than 0
+// and check if the average value for the array is not la
+// TODO: Moving metric_value_benchmark to validator
+// https://github.com/aws/amazon-cloudwatch-agent-test/pull/162
+func IsAllValuesGreaterThanOrEqualToExpectedValue(metricName string, values []float64, expectedValue float64) bool {
```
SaxyPandaBear, line 220: This exists in https://github.com/aws/amazon-cloudwatch-agent-test/blob/8189a84e2d3831d402650639f0ca74d36cb60048/test/metric/metric_validation_util.go#L8 which got merged in first

terraform/ec2/userdata/install_agent.sh
```diff
@@ -0,0 +1,16 @@
...
+cd amazon-cloudwatch-agent-test
+aws s3 cp s3://${binary_uri} .
+export HOME=/root
+export GOPATH=$HOME/go
+export PATH=$PATH:$GOPATH/bin
+echo Running command ${install_agent}
+${install_agent}
+cd /home/ec2-user/amazon-cloudwatch-agent-test/test/userdata/agent_configs/
+echo Starting agent now
+/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s -c file:cpu_config.json
```
SaxyPandaBear, line 15: nit - We don't need to `cd` to this directory. You could just specify the full path

test/test_runner/base_test_runner.go
```diff
@@ -54,22 +62,9 @@ func (t *BaseTestRunner) GetAgentRunDuration() time.Duration {
...
-		testGroupResult = t.TestRunner.Validate()
-	}
-	s.AddToSuiteResult(testGroupResult)
-	if testGroupResult.GetStatus() != status.SUCCESSFUL {
-		log.Printf("%v test group failed due to %v", testName, err)
-	}
-}
-
-func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
+func (t *BaseTestRunner) RunAgent(T *TestRunner) (status.TestGroupResult, error) {
```
SaxyPandaBear, line 65: nit - don't use capital letters for parameter names. 
```suggestion
func (t *BaseTestRunner) RunAgent(runner *TestRunner) (status.TestGroupResult, error) {
```

Also, I think it's confusing that the RunAgent function is defined at the interface level, but requires an implementation of the runner to be passed in.

internal/common/metrics.go
zhihonl: Merged from main in the newest commit to use this method from `metric` package

test/emf_ecs/ecs_resources/ec2_launch/daemon/ecs_taskdef.tpl
```diff
@@ -0,0 +1,80 @@
+[
```
SaxyPandaBear, line 1: This looks like a merge gone wrong. I don't think this `emf_ecs` folder should exist

SaxyPandaBear (dismissed):

test/emf_ecs/ecs_resources/ec2_launch/daemon/ecs_taskdef.tpl
zhihonl: Removed the emf_ecs folder

terraform/ec2/userdata/main.tf
```diff
@@ -0,0 +1,112 @@
...
+    name   = "name"
+    values = [var.ami]
+  }
+}
+
+#####################################################################
+# Generate template file for EC2 userdata script
+#####################################################################
+data "template_file" "init" {
+  template = file("install_agent.sh")
```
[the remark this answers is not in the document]
zhihonl: Changed the name in new commit.

generator/test_case_generator.go
```diff
@@ -87,6 +87,13 @@ var testTypeToTestConfig = map[string][]testConfig{
 	"ec2_acceptance": {
 		{"./test/acceptance", ""},
 	},
+	"ec2_userdata": {
+		{"./test/cloudwatchlogs", ""},
+		{"./test/metric_value_benchmark", ""},
```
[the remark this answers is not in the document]
zhihonl: Initially I was running all these tests to imitate the internal userdata tests which just uses base integration test to test logs and bunch of metrics. These do not install agent outside of userdata since the install behavior is determined in each test's main.tf file and userdata test has its own unique main.tf file, so the agent installed is the one from userdata.

However I agree that having these tests are redundant and for the purpose of testing userdata, just a simple CPU config test should suffice so I removed them.

terraform/ec2/userdata/main.tf
```diff
@@ -0,0 +1,112 @@
...
+      "echo Getting Cloud-init Logs",
+      "sudo cat /var/log/cloud-init-output.log",
+      "echo prepare environment",
+      "export LOCAL_STACK_HOST_NAME=${var.local_stack_host_name}",
+      "export AWS_REGION=${var.region}",
+      "export PATH=$PATH:/snap/bin:/usr/local/go/bin",
+      "echo run integration test",
+      "cd ~/amazon-cloudwatch-agent-test",
+      "sudo chmod 777 ~/amazon-cloudwatch-agent-test/test/sanity/resources/verifyUnixCtlScript.sh",
+      "echo run sanity test && go test ./test/sanity -p 1 -v",
```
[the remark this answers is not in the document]
zhihonl: We are running sanity test based on this [test run](https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/5135270756/jobs/9240520445#step:6:375) so do you mean we shouldn't run sanity test?

test/test_runner/base_test_runner.go
```diff
@@ -41,6 +42,13 @@ type BaseTestRunner struct {
 	DimensionFactory dimension.Factory
 }
 
+func (t *BaseTestRunner) GetTestName() string {
+	return "BaseTestRunner"
+}
+func (t *BaseTestRunner) GetAgentConfigFileName() string {
```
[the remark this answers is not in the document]
zhihonl: Good point. My original implementation required these to be implemented so I can run them in runAgent() method but the new approach no longer needs them. Removed in new commit.

SaxyPandaBear:
Looks like there's a lint failure for terraform

test/test_runner/base_test_runner.go
```diff
@@ -33,6 +33,7 @@ type ITestRunner interface {
 	GetMeasuredMetrics() []string
 	SetupBeforeAgentRun() error
 	SetupAfterAgentRun() error
+	RunAgent(runner *TestRunner) (status.TestGroupResult, error)
```
SaxyPandaBear, line 36: I do really think it's weird to make the TestRunner interface take an implementation of itself as an argument for a method. I'll leave this be for now, but this is due for an overhaul

SaxyPandaBear (dismissed):
There are linting failures, so once those are fixed LGTM

test/test_runner/base_test_runner.go
zhihonl: Lint failures fixed. I agree this is a weird approach but I think we would need to change the way we structure test_runners in order for this to be more straightforward.

terraform/ec2/userdata/main.tf
```diff
@@ -0,0 +1,112 @@
...
+    inline = [
+      "echo Getting Cloud-init Logs",
+      "sudo cat /var/log/cloud-init-output.log",
+      "echo prepare environment",
+      "export LOCAL_STACK_HOST_NAME=${var.local_stack_host_name}",
+      "export AWS_REGION=${var.region}",
+      "export PATH=$PATH:/snap/bin:/usr/local/go/bin",
+      "echo run integration test",
+      "cd ~/amazon-cloudwatch-agent-test",
+      "sudo chmod 777 ~/amazon-cloudwatch-agent-test/test/sanity/resources/verifyUnixCtlScript.sh",
```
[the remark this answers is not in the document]
sky333999: Disregard this... i confused myself

terraform/ec2/userdata/main.tf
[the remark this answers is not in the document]
sky333999: Yup I confused myself... you can disregard this comment.

terraform/ec2/userdata/main.tf
```diff
@@ -0,0 +1,112 @@
...
+      "echo run sanity test && go test ./test/sanity -p 1 -v",
+      "go test ${var.test_dir} -p 1 -timeout 1h -computeType=EC2 -bucket=${var.s3_bucket} -plugins='${var.plugin_tests}' -cwaCommitSha=${var.cwa_github_sha} -caCertPath=${var.ca_cert_path} -v"
```
sky333999, line 81: Dont think plugins for this test.
Not a blocker though.

PR #222, opened 2023-05-15
Implement Integration Testing for EKS Container Insights

JayPolanco:
# Description of the issue
The CCWA does not have complete integration testing for container insights on EKS.

# Description of changes
This CR implements logging tests for container insights on EKS. It creates 11 different json test schemas which are used to verify the logs for the different EKS Cluster types. The logs differ based on the cluster type.

# License

# Tests
Ran tests to verify success - https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4959853782


test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -66,19 +77,80 @@ func (e *EKSDaemonTestRunner) validateInstanceMetrics(name string) status.TestRe
...
+		Status: status.FAILED,
+	}
+
+	now := time.Now()
+	group := fmt.Sprintf("/aws/containerinsights/%s/performance", eks.EKSClusterName)
+
+	// need to get the instances used for the EKS cluster
+	EKSInstances, err := awsservice.GetEKSInstances(eks.EKSClusterName)
+	if err != nil {
+		log.Println("failed to get EKS instances")
```
SaxyPandaBear, line 92: nit - would be nice to bubble up the error in the log:
```suggestion
		log.Println("failed to get EKS instances", err)
```

test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -66,19 +77,80 @@ func (e *EKSDaemonTestRunner) validateInstanceMetrics(name string) status.TestRe
...
+	testResult := status.TestResult{
+		Name:   "emf-logs",
+		Status: status.FAILED,
+	}
+
+	now := time.Now()
+	group := fmt.Sprintf("/aws/containerinsights/%s/performance", eks.EKSClusterName)
+
+	// need to get the instances used for the EKS cluster
+	EKSInstances, err := awsservice.GetEKSInstances(eks.EKSClusterName)
```
SaxyPandaBear, line 90: nit - I don't think I usually see local variables capitalized
```suggestion
	eksInstances, err := awsservice.GetEKSInstances(eks.EKSClusterName)
```

test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -66,19 +77,80 @@ func (e *EKSDaemonTestRunner) validateInstanceMetrics(name string) status.TestRe
...
+	if err != nil {
+		log.Println("failed to get EKS instances")
+		return testResult
+	}
+
+	for _, instance := range EKSInstances {
+		validateLogContents := func(s string) bool {
+			return strings.Contains(s, fmt.Sprintf("\"ClusterName\":\"%s\"", eks.EKSClusterName))
+		}
```
SaxyPandaBear, line 99: nit - we could define this function outside of the for loop

test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -66,19 +77,80 @@ func (e *EKSDaemonTestRunner) validateInstanceMetrics(name string) status.TestRe
...
+
+		var ok bool
+		stream := *instance.InstanceName
+		ok, err = awsservice.ValidateLogs(group, stream, nil, &now, func(logs []string) bool {
```
SaxyPandaBear, line 103: I think that the param is in the wrong spot here. 
```
func ValidateLogs(logGroup, logStream string, since, until *time.Time, validator func(logs []string) bool) (bool, error) {
```

The third param should probably be `&now` and the fourth param should be `nil`

internal/awsservice/eks.go
```diff
@@ -0,0 +1,49 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package awsservice
+
+import (
+	_ "embed"
```
SaxyPandaBear, line 7: Do we need to import this? I don't see an embed in this file

terraform/eks/daemon/main.tf
```diff
@@ -100,6 +101,11 @@ resource "aws_iam_role_policy_attachment" "node_CloudWatchAgentServerPolicy" {
   role       = aws_iam_role.node_role.name
 }
 
+resource "aws_iam_role_policy_attachment" "node_AmazonEKSClusterPolicy" {
```
SaxyPandaBear, line 104: Question - do we need the cluster policy on this IAM role? The node role is attached to each node of the cluster. The purpose (as I know it) of the AmazonEKSClusterPolicy managed policy is it is for managing the cluster.

test/metric_value_benchmark/eks_daemonset_test.go
JayPolanco: Sure, but then we would need to pass the environment.Metadata to the function which would then make it so we have to change the MatchEMFLogWithSchema function as it only takes a func(string). Would like to avoid this as it's also used for the ECS tests.

SaxyPandaBear: Are you sure? I'm looking specifically at this `for _, instance := range EKSInstances`. The validation function here doesn't actually use anything from each instance. It's just validating the cluster name so we could rip it out from here to define once and reuse

JayPolanco: Aren't you talking about the `validateLogContents` function? If so, this function gets used in `if !awsservice.MatchEMFLogWithSchema(l, rs, validateLogContents) {` and requires the function to be a func(string)

internal/awsservice/eks.go
JayPolanco: No, will remove. Had the embed here originally but moved it out

test/metric_value_benchmark/eks_daemonset_test.go
SaxyPandaBear: Yes. I'm saying you could define it outside of the for loop, pretty sure. But it's not super important so I'm fine just leaving it. Don't want to spend a lot of time on it

test/metric_value_benchmark/eks_daemonset_test.go
SaxyPandaBear: 🤔 I guess it depends on when `now` is defined relative to the logs being generated... if that's what it is right now and the ECS CI tests are working, then I'll live with it

terraform/eks/daemon/main.tf
JayPolanco: Honestly, I added this to try and resolve some of the issues with the iam roles, but I don't think this was actually needed. Can't say for sure as all my testing was done with it present, but I think it would be fine to remove it and if need be, we can add it back

test/metric_value_benchmark/eks_daemonset_test.go
JayPolanco: done

terraform/eks/daemon/main.tf
JayPolanco: removed, will add back if it is determined to be necessary

test/metric_value_benchmark/eks_daemonset_test.go
JayPolanco: correct, addressed

movence (approved):
LGTM

test/metric_value_benchmark/eks_resources/test_schemas/cluster.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "json schema for the cloudwatch agent k8s structured log",
+  "type": "object",
+
+  "properties": {
+    "ClusterName": {},
+    "Type": {},
+    "Sources": {},
+    "Timestamp": {},
+    "Version": {},
+    "cluster_failed_node_count": {},
```
sky333999, line 13: Why are metrics like `cluster_failed_node_count` not `required`?
We've added `namespace_number_of_running_pods` to `required` but in most of the others, we seem to leave out the metrics.

test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -66,19 +76,80 @@ func (e *EKSDaemonTestRunner) validateInstanceMetrics(name string) status.TestRe
...
+			}
+
+			for _, l := range logs {
+				var eksClusterType awsservice.EKSClusterType
+				err := json.Unmarshal([]byte(l), &eksClusterType)
+				if err != nil {
+					log.Println("failed to unmarshal log file")
+				}
+
+				log.Println(fmt.Sprintf("eksClusterType is: %s", eksClusterType.Type))
```
sky333999, line 115: These arent really ClusterTypes... they are just LogEvent types

internal/awsservice/eks.go
SaxyPandaBear: ... good catch. I didn't even think about why I didn't see a new file to review the second time around

test/metric_value_benchmark/eks_daemonset_test.go
SaxyPandaBear: I can clean this up at some point. Will add it to my list of things

PR #226, opened 2023-05-18
Add SSM Rename Test To Match Internal Pipeline Test

sethAmazon:
# Description of the issue
Need to test all the functionality that was in the internal pipeline tests
Add ssm tests 

# Description of changes
Add ssm tests

# License

# Tests
Ran in the private repo gha

### DIFF
--- a/internal/awsservice/cloudformation.go
+++ b/internal/awsservice/cloudformation.go
@@ -42,12 +42,16 @@ func StartStack(ctx context.Context, stackName string, client *cloudformation.Cl
 }
 
 func DeleteStack(ctx context.Context, stackName string, client *cloudformation.Client) {
+	r := recover()
 	deleteStackInput := cloudformation.DeleteStackInput{
 		StackName: aws.String(stackName),
 	}
 	_, err := client.DeleteStack(ctx, &deleteStackInput)
 	if err != nil {
-		log.Fatalf("Could not delete stack %v", stackName)
+		log.Fatalf("Could not delete stack %s", stackName)
+	}
+	if r != nil {
+		log.Fatalf("Delete stack %s is called on panic thus delete stack then continue panic", stackName)
 	}
 }
 
@@ -59,12 +63,12 @@ func FindStackInstanceId(ctx context.Context, stackName string, client *cloudfor
 		stacks, err := client.DescribeStacks(ctx, &cfStackInput)
 		if err != nil || len(stacks.Stacks) != 1 || stacks.Stacks[0].StackStatus != types.StackStatusCreateComplete {
 			log.Printf("Stack %s not ready in minute %d continue to next minute", stackName, i)
-			continue
-		}
-		for output := range stacks.Stacks[0].Outputs {
-			if *stacks.Stacks[0].Outputs[output].OutputKey == instanceIdKey {
-				log.Printf("Found instance id %s from stack %s", *stacks.Stacks[0].Outputs[output].OutputValue, stackName)
-				return *stacks.Stacks[0].Outputs[output].OutputValue
+		} else {
+			for output := range stacks.Stacks[0].Outputs {
+				if *stacks.Stacks[0].Outputs[output].OutputKey == instanceIdKey {
+					log.Printf("Found instance id %s from stack %s", *stacks.Stacks[0].Outputs[output].OutputValue, stackName)
+					return *stacks.Stacks[0].Outputs[output].OutputValue
+				}
 			}
 		}
 		log.Printf("Sleep for one minute to wait for stack to start")
--- a/internal/common/agent_util_windows.go
+++ b/internal/common/agent_util_windows.go
@@ -45,7 +45,9 @@ func CopyFile(pathIn string, pathOut string) error {
 
 }
 
-func StartAgent(configOutputPath string, fatalOnFailure bool) error {
+func StartAgent(configOutputPath string, fatalOnFailure bool, ssm bool) error {
+	// @TODO add ssm functionality
+
 	ps, err := exec.LookPath("powershell.exe")
 
 	if err != nil {
@@ -111,7 +113,7 @@ func printOutputAndError(stdout []byte, err error) {
 		return
 	}
 	stderr := ""
-	ee, ok :=  err.(*exec.ExitError)
+	ee, ok := err.(*exec.ExitError)
 	if ok {
 		stderr = string(ee.Stderr)
 	}
--- a/terraform/ec2/win/main.tf
+++ b/terraform/ec2/win/main.tf
@@ -30,6 +30,7 @@ resource "aws_key_pair" "aws_ssh_key" {
 locals {
   ssh_key_name        = var.ssh_key_name != "" ? var.ssh_key_name : aws_key_pair.aws_ssh_key[0].key_name
   private_key_content = var.ssh_key_name != "" ? var.ssh_key_value : tls_private_key.ssh_key[0].private_key_pem
+  ssm_parameter_name  = "WindowsAgentConfigSSMTest"
 }
 
 #####################################################################
@@ -72,8 +73,21 @@ resource "aws_instance" "cwagent" {
   }
 }
 
+data "local_file" "input" {
+  filename = module.validator.agent_config
+}
+
+# Size of windows json is too large thus can't use standard tier
+resource "aws_ssm_parameter" "upload_ssm" {
+  count = var.use_ssm == true ? 1 : 0
+  name  = local.ssm_parameter_name
+  type  = "String"
+  tier  = "Advanced"
+  value = data.local_file.input.content
+}
+
 resource "null_resource" "integration_test" {
-  depends_on = [aws_instance.cwagent, module.validator]
+  depends_on = [aws_instance.cwagent, module.validator, aws_ssm_parameter.upload_ssm]
 
   # Install software
   connection {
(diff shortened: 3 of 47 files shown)


internal/awsservice/cloudformation.go
```diff
@@ -42,12 +42,16 @@ func StartStack(ctx context.Context, stackName string, client *cloudformation.Cl
 }
 
 func DeleteStack(ctx context.Context, stackName string, client *cloudformation.Client) {
+	r := recover()
```
sethAmazon, line 45: Was not deleting stack on panic.

internal/awsservice/cloudformation.go
```diff
@@ -59,12 +63,12 @@ func FindStackInstanceId(ctx context.Context, stackName string, client *cloudfor
 		stacks, err := client.DescribeStacks(ctx, &cfStackInput)
 		if err != nil || len(stacks.Stacks) != 1 || stacks.Stacks[0].StackStatus != types.StackStatusCreateComplete {
 			log.Printf("Stack %s not ready in minute %d continue to next minute", stackName, i)
-			continue
```
sethAmazon, line 62: The short circuit was not allowing the sleep command to execute.

test/cloudformation/cloudformation_test.go
```diff
@@ -1,7 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT
 
-// TODO: we need to revisit this
```
sethAmazon, line 4: This does not need to run on windows. No need for todo

test/metric/stat.go
```diff
@@ -11,5 +11,5 @@ const (
 	SAMPLE_COUNT             Statistics = "SampleCount"
 	MINIMUM                  Statistics = "Minimum"
 	MAXUMUM                  Statistics = "Maxmimum"
-	HighResolutionStatPeriod            = 30
+	HighResolutionStatPeriod            = 10
```
sethAmazon, line 14: Reduce run time of metric tests

test/metric_value_benchmark/container_insights_test.go
```diff
@@ -54,8 +54,8 @@ func (t *ContainerInsightsTestRunner) GetAgentConfigFileName() string {
 	return "./agent_configs/container_insights.json"
 }
 
-func (t *ContainerInsightsTestRunner) GetAgentRunDuration() time.Duration {
-	return test_runner.MinimumAgentRuntime
+func (t *ContainerInsightsTestRunner) getAgentRunDuration() time.Duration {
```
sethAmazon, line 57: Because I reduced agent runtime for ec2 tests. Let ecs stay as 1 minute run time.

test/metric_value_benchmark/rename_ssm_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/aws/amazon-cloudwatch-agent-test/test/test_runner"
+)
+
+type RenameSSMTestRunner struct {
+	test_runner.BaseTestRunner
+}
+
+var _ test_runner.ITestRunner = (*RenameSSMTestRunner)(nil)
+
+func (m *RenameSSMTestRunner) Validate() status.TestGroupResult {
```
sethAmazon, line 21: This test matches the functionality of the internal ssm test

internal/common/agent_util_windows.go
```diff
@@ -45,7 +45,9 @@ func CopyFile(pathIn string, pathOut string) error {
 
 }
 
-func StartAgent(configOutputPath string, fatalOnFailure bool) error {
+func StartAgent(configOutputPath string, fatalOnFailure bool, ssm bool) error {
+	// @TODO add ssm functionality
```
sethAmazon, line 49: Will add a windows ssm test in another pr.

internal/awsservice/cloudformation.go
SaxyPandaBear, line 45: Should we push stuff out to the terraform level, and use terraform to create the stack, and then have a step after that terraform apply to run the CloudFormation test?

test/test_runner/base_test_runner.go
```diff
@@ -17,10 +20,10 @@ import (
...
 
 const (
-	configOutputPath     = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
-	agentConfigDirectory = "agent_configs"
-	extraConfigDirectory = "extra_configs"
-	MinimumAgentRuntime  = 1 * time.Minute
+	configOutputPath         = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	agentConfigDirectory     = "agent_configs"
+	extraConfigDirectory     = "extra_configs"
+	ThirtySecondAgentRuntime = 30 * time.Second
```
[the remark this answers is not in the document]
SaxyPandaBear: I don't really see the point of explicitly calling it `ThirtySecond`

test/test_runner/base_test_runner.go
```diff
@@ -31,6 +34,8 @@ type ITestRunner interface {
 	GetMeasuredMetrics() []string
 	SetupBeforeAgentRun() error
 	SetupAfterAgentRun() error
+	SSM() bool
```
SaxyPandaBear, line 37: nit - I can't tell what this means. 
```suggestion
	UseSSM() bool
```

test/test_runner/base_test_runner.go
```diff
@@ -84,9 +97,27 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
...
+		if err != nil {
+			return testGroupResult, errors.New("failed while reading config file")
+		}
+
+		agentConfig := string(agentConfigByteArray)
+
+		err = awsservice.PutStringParameter(t.TestRunner.SSMParameterName(), agentConfig)
+		if err != nil {
+			return testGroupResult, errors.New("failed to upload ssm parameter")
+		}
```
SaxyPandaBear, line 113: Why not just put this in the specific test runner's `SetupBeforeAgentRun()` function

internal/awsservice/cloudformation.go
sethAmazon: I like not needing terraform for our cf test. I think this is simpler for cf but I may be wrong.

test/test_runner/base_test_runner.go
```diff
@@ -29,8 +31,11 @@ type ITestRunner interface {
 	GetAgentConfigFileName() string
 	GetAgentRunDuration() time.Duration
 	GetMeasuredMetrics() []string
-	SetupBeforeAgentRun() error
+	SetupBeforeAgentRun(configFileName string, sSMParameterName string, useSSM bool) error
```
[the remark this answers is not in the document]
sethAmazon: I did this because of a comment to move agent config to the before agent start method

terraform/ec2/win/main.tf
```diff
@@ -106,7 +123,7 @@ resource "null_resource" "integration_test" {
     inline = [
       "set AWS_REGION=${var.region}",
       "validator.exe --validator-config=${module.validator.instance_validator_config} --preparation-mode=true",
-      "powershell \"& 'C:\\Program Files\\Amazon\\AmazonCloudWatchAgent\\amazon-cloudwatch-agent-ctl.ps1' -a fetch-config -m ec2 -s -c file:${module.validator.instance_agent_config}\"",
+      var.use_ssm == "powershell \"& 'C:\\Program Files\\Amazon\\AmazonCloudWatchAgent\\amazon-cloudwatch-agent-ctl.ps1' -a fetch-config -m ec2 -s -c ssm:${local.ssm_parameter_name}\"" ? 1 : "powershell \"& 'C:\\Program Files\\Amazon\\AmazonCloudWatchAgent\\amazon-cloudwatch-agent-ctl.ps1' -a fetch-config -m ec2 -s -c file:${module.validator.instance_agent_config}\"",
```
movence, line 126: should this be `var.use_ssm == 1 ? ...` instead?

test/test_runner/base_test_runner.go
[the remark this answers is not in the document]
SaxyPandaBear: Could the values here be just fields of the runner so that we don't expose this implementation detail?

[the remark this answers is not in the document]
sethAmazon: Yes, but then we would need to insert them when we create the struct not the test case. Which is a pain. 

https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/metric_value_benchmark/metrics_value_benchmark_test.go#L86-L107

test/metric/stat.go
adam-mateen: Can you please check if this breaks any existing tests?
I think the statsd test might assume 30 seconds.
I know the `metric_aggregation_interval` in the JSON config is 30 seconds.
In general for any test that gets changed you should include screenshots or text output showing the test passes.

internal/awsservice/cloudformation.go
adam-mateen: +1 for being able to run `go test ./abc/def -run xyz` from an EC2 instance and having it test CWA on the local EC2 instance, or in this case execute CFN to launch another instance.

sethAmazon:
Failure for metric value benchmark for rhel8 and ol9 only same as the current private repo. It passes that test on other os same as private repo. 

Screenshot 2023-05-23 at 3 31 22 PM
Screenshot 2023-05-23 at 3 31 32 PM

test/metric/stat.go
adam-mateen: I'd prefer to get input from other engineers about whether we should be reducing run time.
Is testing each feature for a minute or 2 really enough?

Was the reason to reduce test time due to GitHub limits or what?

sethAmazon: Added screen shots now.

sethAmazon:
Screenshot 2023-05-23 at 3 33 36 PM

test/metric/stat.go
adam-mateen: Why not just increase how long the credentials are valid?

The action:
https://github.com/aws-actions/configure-aws-credentials/tree/v2/
has a `role-duration-seconds` option.

Can we change the IAM role in AWS console as well?

adam-mateen: Or what if instead of a single GitHub "step" in a single GitHub "job" that runs a bunch of tests, we spread the tests across many GitHub jobs. If there are 4 jobs that each run 1/4 of the tests, and each job gets credentials at the start of the job, would that solve the problem?

sethAmazon: Good idea. I'll put in a pr to the gha. I still like reducing the test time. Furthermore, I dislike how long it takes to run tests. Not only that, but I feel like the shorter. The better.

sethAmazon: I still would like to keep this change if possible. Going from 30 minutes for the metric agent benchmark test per run to 20 is a win.

test/test_runner/base_test_runner.go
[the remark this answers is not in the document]
sethAmazon: Resolved by using an inner struct

test/metric/stat.go
adam-mateen: Shorter test time makes developer cycle time for build/test/fix faster.
At some point we want longer running tests.
We do have a few "soak tests", but those don't cover every feature.

So if you feel that each of the features in the metric_value_benchmark suite are "sufficiently" tests by these 1 or 2 minute tests than I am fine keeping it that short.

Approving.

SaxyPandaBear (approved):

PR #231, opened 2023-05-23
Add test specific metric namespace to Statsd tests

movence:
# Description of the issue
All Statsd metric going to the same name space `StatsD` that might cause clashed data

# Description of changes
Add Statsd metric namespace that follows the format of `[feature]/[infra]`

# License


test/statsd/default_test.go
```diff
@@ -27,13 +27,13 @@ func (t *StatsDRunner) Validate() status.TestGroupResult {
 
 	// ECS taskdef with portMappings has some delays before getting metrics from statsd container
 	if strings.Contains(t.testName, "ECS") {
-		time.Sleep(30 * time.Second)
+		time.Sleep(60 * time.Second)
```
SaxyPandaBear, line 30: if we tweaked the aggregation interval to lower, could we also shave off time on this sleep?

SaxyPandaBear (approved):

test/statsd/default_test.go
movence: It could be. My hunch is that it has more to do with some delays from ECS container link, and I didn't want to have a flaky test by having a shorter sleep.

JayPolanco (approved):

PR #236, opened 2023-06-05
Add fluentd & fluentbit integration tests

movence:
# Description of changes
- Add `fluentd/fluentbit` integration tests with EKS
- Add sets of test target to test generator

# License

# Tests
on personal fork (**EMF tests are failing for some reason, but it's not related to this PR**)
https://github.com/movence/amazon-cloudwatch-agent/actions/runs/5176849905


generator/test_case_generator.go
```diff
@@ -36,69 +36,85 @@ type testConfig struct {
 	// e.g. statsd can have a multiple terraform module sets for difference test scenarios (ecs, eks or ec2)
 	testDir      string
 	terraformDir string
+	// define target matrix field as set(s)
+	// empty map means a testConfig will be created with a test entry for each entry from *_test_matrix.json
+	targets map[string]map[string]struct{}
```
sky333999, line 41: So these really seem like filters to decide whether or not an entry in the json should be retained in the matrix.

Why does need to be a map of map? Can this just be a map where the key value pairs match the key value pairs in the json?
So for example, a map with entry <"arc": "arm64">

generator/test_case_generator.go
```diff
@@ -36,69 +36,85 @@ type testConfig struct {
...
+		},
+		{
+			testDir: "./test/emf", terraformDir: "terraform/eks/daemon/emf",
+			targets: map[string]map[string]struct{}{"arc": {"amd64": {}}},
+		},
+		{
+			testDir: "./test/fluent", terraformDir: "terraform/eks/daemon/fluent/d",
+			targets: map[string]map[string]struct{}{"arc": {"amd64": {}}},
+		},
+		{testDir: "./test/fluent", terraformDir: "terraform/eks/daemon/fluent/bit"},
```
sky333999, line 114: Any reason we run fluentbit on both arm & amd but fluentd only on amd?

generator/resources/eks_daemon_test_matrix.json
```diff
@@ -1,5 +1,14 @@
...
-    "k8s_version": "1.24"
+    "k8s_version": "1.24",
+    "ami": "AL2_x86_64",
+    "instanceType":"t3a.medium",
+    "arc": "amd64"
+  },
+  {
+    "k8s_version": "1.24",
+    "ami": "AL2_ARM_64",
+    "instanceType":"m6g.large",
```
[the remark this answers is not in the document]
movence: yeah I was running into no resources available with `m6g.medium` when running the same pods from amd64 test using `t3a.medium`

generator/test_case_generator.go
movence: it's map of sets for fast lookup eg
```
{
   "arc": ["amd", "arm"],
   "os": ["al2", "ol9"],
}
```

generator/test_case_generator.go
movence: I don't know the history of the original tests, but only fluent-bit has both ARM and AMD covered.

sky333999: Hmm feels like we should be consistent for all of these infact.. but not a blocker if thats existing behavior.

sky333999 (approved):

internal/awsservice/cloudwatchlogs.go
```diff
@@ -135,6 +135,22 @@ func IsLogGroupExists(logGroupName string) bool {
...
+func GetLogStreams(logGroupName string) []types.LogStream {
+	describeLogStreamsOutput, err := CwlClient.DescribeLogStreams(ctx, &cloudwatchlogs.DescribeLogStreamsInput{
+		LogGroupName: aws.String(logGroupName),
+		OrderBy:      types.OrderByLastEventTime,
+		Descending:   aws.Bool(true),
+		Limit:        aws.Int32(10),
+	})
+
+	if err != nil {
+		log.Printf("failed to get log streams for log group: %s - err: %s", logGroupName, err)
```
SaxyPandaBear, line 147: nit. `%s` is for strings but you're passing an error for the second arg. I always use `%v`, but am open to whatever.

SaxyPandaBear (dismissed):

generator/test_case_generator.go
```diff
@@ -36,69 +36,85 @@ type testConfig struct {
...
-
+		{testDir: "./test/userdata"},
+	},
 	"eks_daemon": {
-		{"./test/metric_value_benchmark", ""},
-		{"./test/statsd", "terraform/eks/daemon/statsd"},
-		{"./test/emf", "terraform/eks/daemon/emf"},
+		{
+			testDir: "./test/metric_value_benchmark",
+			targets: map[string]map[string]struct{}{"arc": {"amd64": {}}},
```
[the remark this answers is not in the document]
movence: Yeah, I totally agree that there must be a cleaner and simpler way to do this. Let me think about it and address it in a different PR.

generator/test_case_generator.go
```diff
@@ -134,11 +150,35 @@ func genMatrix(testType string, testConfigs []testConfig) []matrixRow {
...
+func shouldAddTest(row *matrixRow, targets map[string]map[string]struct{}) bool {
+	for key, set := range targets {
+		var rowVal string
+		if key == "arc" {
+			rowVal = row.Arc
+		}
+
+		if rowVal == "" {
+			return true
+		}
```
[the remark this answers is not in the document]
movence: you are right on continue, totally missed it.
This could grow to include other configs like `os` `ami` and etc.

internal/awsservice/cloudwatchlogs.go
movence: updated

test/fluent/fluent_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+		ok, err := awsservice.ValidateLogs(group, *(streams[0].LogStreamName), nil, &now, func(logs []string) bool {
+			if len(logs) < 1 {
+				return false
+			}
+
+			// only 1 log message gets validate
+			// log message must include expected fields, and there could be more than 1 set of expected fields  per log group
+			var found = false
+			for _, fields := range filedsArr {
```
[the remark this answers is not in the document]
movence: fixed, good catch!

test/fluent/fluent_test.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	t.Log("finishing EKS fluent log validation...")
+}
+
+func getLogStreams(logGroupName string) []types.LogStream {
+	logStreams := make([]types.LogStream, 0)
+	for i := 0; i < logStreamRetry; i++ {
+		logStreams = awsservice.GetLogStreams(logGroupName)
```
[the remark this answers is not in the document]
movence: good call. moved to `GetLogStreams`

internal/awsservice/cloudwatchlogs.go
```diff
@@ -135,6 +137,30 @@ func IsLogGroupExists(logGroupName string) bool {
...
+			continue
+		}
+		
+		if len(describeLogStreamsOutput.LogStreams) > 0 {
+			return describeLogStreamsOutput.LogStreams 
+		}
+
+		time.Sleep(10 * time.Second)
+	}
+
```
SaxyPandaBear, line 160: nit - it would be nice to add a log here that we exhausted retries so we returned nothing

SaxyPandaBear (approved):

SaxyPandaBear (approved):

sky333999 (approved):

PR #238, opened 2023-06-05
Default soak time

adam-mateen:
# Description of the issue
The soak test is a short test that just starts the agent and some background generator programs.
The generator programs run indefinitely.
The intention was that the Terraform code would run a bash sleep command followed by a shutdown which would cause the EC2 instance to terminate. While that is still possible, I also think it is a good idea to allow users to control how long the generators run.

# Description of changes
Considered using a channel and having all goroutines check the channel.
Simpler than that is just sleeping the desired amount of time and then exiting.
There is no clean up to worry about.

# License

# Tests
- Manual testing:
    - `go run ./cmd/statsd-generator -runTime=15s -clientNum=7 -tps=100 -metricNum=10`
    - `go run ./cmd/log-generator -runTime=1m`
    - `go run ./cmd/log-generator -runTime=10s`
    - `go run ./cmd/emf-generator -runTime=10s -fileNum=100 -eventsPerSecond=100`

### DIFF
--- a/cmd/log-generator/log-generator.go
+++ b/cmd/log-generator/log-generator.go
@@ -3,13 +3,11 @@ package main
 import (
 	"flag"
 	"fmt"
-	"math"
 	"math/rand"
 	"os"
 	"path"
 	"runtime"
 	"strconv"
-	"sync"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/log-generator/log-generator.go
```diff
@@ -25,7 +23,7 @@ var (
 	eventSize       = flag.Int("eventSize", 120, "Identify the single log line size, in Byte.")
 	outPutPath      = flag.String("path", "", "Identify the path where log files will generate.")
 	filePrefix      = flag.String("filePrefix", "tmp", "Identify the log file prefix")
-	loopNum         = flag.Int64("loopNum", math.MaxInt64, "How many loops to write teh logentry.")
```
adam-mateen, line 28: For context these programs were just copied from an internal test repository. However `loopNum` is not needed. It indirectly controls the runtime, so I replaced it.

JayPolanco: Nice, this is much cleaner

JayPolanco (approved):

jefchien (approved):

PR #240, opened 2023-06-05
Fix async process

adam-mateen:
# Description of the issue
When the terraform code runs the test, the generators will exit when the terraform "remote-exec" exits.
We want the soak test to start generators which outlast the Terraform code.

I ran the soak test and then SSH'd into the ec2 instance and noticed the generators were not running as expected.

Related issues:
- https://github.com/hashicorp/terraform/issues/6229
- https://github.com/hashicorp/terraform/issues/5113
- http://stackoverflow.com/questions/36207752/how-can-i-start-a-remote-service-using-terraform-provisioning

# Description of changes
- Use `nohup` when starting background (aka async) processes on Linux. (Not sure if there is an issue with Windows yet. We can deal with that later).

# License

# Tests
- Manually ran the soak test workflow, then SSH'd into the EC2 instance and verified generators were left running.


ymtaye (approved):
LGTM!

bryce-carey (approved):

PR #242, opened 2023-06-08
migrate assume role credentials test

movence:
# Description of changes
Migrate assume role credentials test to GHA. The new terraform `ec2/creds/main` will create a temporary role that `cwa-e2e-iam-role` can assume.

# License

# Tests
Tested with a forked repo.
```
null_resource.integration_test_run (remote-exec): === RUN   TestAssumeRole
null_resource.integration_test_run (remote-exec): 2023/06/08 21:21:09 Running AssumeRoleTest
...
null_resource.integration_test_run (remote-exec): --- PASS: TestAssumeRole (31.93s)
null_resource.integration_test_run (remote-exec): PASS
null_resource.integration_test_run (remote-exec): ok  	github.com/aws/amazon-cloudwatch-agent-test/test/assume_role	31.932s
```


test/assume_role/agent_configs/config.json
```diff
@@ -0,0 +1,26 @@
+{
+  "agent": {
```
sethAmazon, line 2: This is not the role_arn for sending metrics to another account?

sethAmazon:
I thought our tests already assume the ec2 role. How does this differ?

movence:
Existing integ tests already use assume role attached as IAM role to an instance (instance profile by using `cwa-e2e-iam-role`). This test is specifically testing STS credentials written on a host, then CWA uses that temporary creds to emit data to CW. The reason this test creates another temporary IAM role is simply to get accessKey/secret/sessionToken by calling `sts assume-role` (temporary IAM role gets `cwa-e2e-iam-role` in its trust policy).

test/assume_role/agent_configs/config.json
movence: Are you tagging to a right line? or are you saying there should be another config in agent's config instead having creds file path in `/opt/aws/amazon-cloudwatch-agent/etc/common-config.toml`?

sethAmazon (approved):

sethAmazon:
Since this tests the common config to set the shared file and profile I do not think we will need a test for perm creds.

test/assume_role/agent_configs/config.json
SaxyPandaBear: There is a way to assume role directly for sending metrics or logs to a different account. If this is a lift and shift from the internal test, I don't really care which way we do it.

PR #245, opened 2023-06-12
Migrate multi config, runasuser mac tests

ymtaye:
# Description of changes
- Migrate `multi_config` test to GHA
-  add mac support for exisitng runasuser test 

# License

# Tests
Run tests in private fork


test/multi_config/multi_config_linux_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+func TestMultipleConfig(t *testing.T) {
+
+	agentConfigurations := []string{"resources/LinuxLogOnlyConfig.json", "resources/LinuxMemoryOnlyConfig.json"}
+
+	for index, agentConfig := range agentConfigurations {
+		common.CopyFile(agentConfig, configOutputPath)
+		log.Printf(configOutputPath)
+		if index == 0 {
+			common.StartAgent(configOutputPath, true, false)
```
adam-mateen, line 43: Would/could it make a difference if we slept some time between loops? Is that worth testing?

test/multi_config/multi_config_linux_test.go
```diff
@@ -0,0 +1,66 @@
...
+		types.DimensionFilter{
+			Name:  aws.String("InstanceId"),
+			Value: aws.String(ec2InstanceId),
+		},
+	}
+
+	expectedMetrics := []string{"mem_used_percent"}
+	for _, expectedMetric := range expectedMetrics {
+		awsservice.ValidateMetrics(t, expectedMetric, namespace, expectedDimensions)
+	}
```
adam-mateen, line 65: This test has a log config and a metric config, but we only validate the 1 memory metric.
Instead I think it would be better to also validate some log data.
Or if that is not possible, consider adding a few more configs for cpu, disk, net, etc... And after each added config check for the presence of the new metric.

1. Start with config for just mem
2. Verify mem metrics
3. Add config for cpu
4. Verify cpu metrics
5. Add config for ...
6. Verify ... metrics

I realize this might be more than what we tested in the past, but that is a good thing :)

test/multi_config/multi_config_windows_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/environment"
+)
+
+const (
+	configOutputPath = "C:\\ProgramData\\Amazon\\AmazonCloudWatchAgent\\config.json"
```
adam-mateen, line 17: This code is very similar to the linux test.
I suggest you create:
```
test/multi_config/multi_config_linux.go
test/multi_config/multi_config_windows.go
test/multi_config/multi_config_test.go
```

In the OS specific files you can define:
```
namespace
configOutputPath
agentConfigurations
```
and whatever else is unique to the 2 OSes, such that the test code can be common.

Assuming there is a set of metrics that can be collected from both hosts - cpu, mem, disk, net, and the only difference would be the namespace (might not even need a different namespace since the instance id dimension is used)

util/common/agent_util_windows.go
```diff
@@ -46,6 +46,29 @@ func CopyFile(pathIn string, pathOut string) error {
...
 
+func StartAgentWithMultiConfig(configOutputPath string, fatalOnFailure bool, ssm bool) error {
+
+	ps, err := exec.LookPath("powershell.exe")
+
+	if err != nil {
+		return err
+	}
+
+	bashArgs := append([]string{"-NoProfile", "-NonInteractive", "-NoExit", "& \"C:\\Program Files\\Amazon\\AmazonCloudWatchAgent\\amazon-cloudwatch-agent-ctl.ps1\" -a append-config -m ec2 -s -c file:" + configOutputPath})
```
adam-mateen, line 57: nit: `bashArgs` isn't the best name. Maybe just `params`, `args`, `arguments`.

adam-mateen (approved):
Approving, since this definitely adds test coverage. If merging, then I suggest opening another task to increase test coverage.

test/multi_config/resources/WindowsLogOnlyConfig.json
```diff
@@ -0,0 +1,132 @@
...
+      },
+      "LogicalDisk": {
+        "measurement": [
+          "% Free Space"
+        ],
+        "resources": [
+          "*"
+        ]
+      },
+      "PhysicalDisk": {
```
adam-mateen, line 28: This JSON file has a misleading name. It is called `WindowsLogOnlyConfig.json`, but I see a lot of metrics in it.

adam-mateen (approved):

test/multi_config/multi_config_test.go
```diff
@@ -0,0 +1,35 @@
...
+	environment.RegisterEnvironmentMetaDataFlags()
+}
+
+func AppendConfigs(config []string, configOutputPath string) {
+	for index, agentConfig := range config {
+		common.CopyFile(agentConfig, configOutputPath)
+
+		log.Printf(configOutputPath)
+		if index == 0 {
+			common.StartAgent(configOutputPath, true, false)
```
sethAmazon, line 22: So we always start the agent normally on the first config sent in?

test/multi_config/multi_config_test.go
```diff
@@ -0,0 +1,35 @@
...
+		} else {
+			common.StartAgentWithMultiConfig(configOutputPath, true, false)
+		}
+		time.Sleep(30 * time.Second)
```
sethAmazon, line 26: Why do we need to wait 30 seconds can this be less?

test/multi_config/multi_config_unix_test.go
```diff
@@ -0,0 +1,54 @@
...
+	configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	namespace        = "MultiConfigTest"
+)
+
+// Let the agent run for 2 minutes. This will give agent enough time to call server
+const agentRuntime = 2 * time.Minute
+
+func Validate() error {
+
+	agentConfigurations := []string{"resources/LinuxCpuOnlyConfig.json", "resources/LinuxMemoryOnlyConfig.json", "resources/LinuxDiskOnlyConfig.json"}
```
sethAmazon, line 28: Nit: How about a dir for each os. ex resources/linux/CpuOnlyConfig.json

util/common/agent_util_unix.go
```diff
@@ -121,6 +121,26 @@ func InstallAgent(installerFilePath string) error {
 	return err
 }
 
+func StartAgentWithMultiConfig(configOutputPath string, fatalOnFailure bool, ssm bool) error {
+	path := "file:"
+	if ssm {
+		path = "ssm:"
+	}
+	out, err := exec.
+		Command("bash", "-c", "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a append-config -m ec2 -s -c "+path+configOutputPath).
```
sethAmazon, line 130: I don't know how append config works. I assume if the agent is already running with a config then you run this command it add this config to already running config?

sethAmazon (approved):

util/common/agent_util_windows.go
```diff
@@ -33,8 +33,8 @@ func CopyFile(pathIn string, pathOut string) error {
 	}
 
 	log.Printf("File %s abs path %s", pathIn, pathInAbs)
-	bashArgs := append([]string{"-NoProfile", "-NonInteractive", "-NoExit", "cp " + pathInAbs + " " + pathOut})
-	out, err := exec.Command(ps, bashArgs...).Output()
+	params := append([]string{"-NoProfile", "-NonInteractive", "-NoExit", "cp " + pathInAbs + " " + pathOut})
+	out, err := exec.Command(ps, params...).Output()
```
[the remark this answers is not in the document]
ymtaye: Yeah, it was to fix naming

test/multi_config/multi_config_windows_test.go
```diff
@@ -0,0 +1,56 @@
...
+	// test for cloud watch metrics
+	ec2InstanceId := awsservice.GetInstanceId()
+	expectedDimensions := []types.DimensionFilter{
+		types.DimensionFilter{
+			Name:  aws.String("InstanceId"),
+			Value: aws.String(ec2InstanceId),
+		},
+	}
+
+	expectedMetrics := []string{"% Committed Bytes In Use", "% InterruptTime"}
```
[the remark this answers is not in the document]
ymtaye: I can remove the memory metric from the other config

sethAmazon (dismissed):

adam-mateen (dismissed):

adam-mateen (approved):

PR #252, opened 2023-06-15
Performance test bug fixes

sky333999:
# Description of the issue
Bugs were introduced over past commits that break the performance tests.

# Description of changes
* Update test dir paths for win, mac & validator. The validator should not use contextual paths since it can be imported at any level by diff plugnins. For example, mac & win are nested 3 levels deep but performance is nested 2 levels deep. And given they both source in the validator, it cant have relative paths.
* Remove unsuppported `metrics_collection_interval` for ethtool.

# License

# Tests
* Run with passing performance/system tests: https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/5280181528
* WIP to test win, mac and other performance tests

### DIFF
--- a/terraform/ec2/win/variable.tf
+++ b/terraform/ec2/win/variable.tf
@@ -53,7 +53,7 @@ variable "test_name" {
 
 variable "test_dir" {
   type    = string
-  default = "./test/feature/windows"
+  default = ""
 }
 
 variable "use_ssm" {
--- a/generator/test_case_generator.go
+++ b/generator/test_case_generator.go
@@ -92,7 +92,7 @@ var testTypeToTestConfig = map[string][]testConfig{
 		and Mac under the hood share similar plugins with Linux
 	*/
 	"ec2_mac": {
-		{testDir: "./test/feature/mac"},
+		{testDir: "../../../test/feature/mac"},
 	},
 	"ec2_windows": {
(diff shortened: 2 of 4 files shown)


chadpatel (approved):

movence (dismissed):

movence (approved):

sky333999:
This doesnt fully fix all the win & mac test failures, but puts us in a better state and would require additional fixes on top of this.

ChenaLee (approved):

terraform/ec2/win/variable.tf
```diff
@@ -53,7 +53,7 @@ variable "test_name" {
 
 variable "test_dir" {
   type    = string
-  default = "./test/feature/windows"
+  default = ""
```
zhihonl, line 56: Can we change this to `../../../test/feature/windows` as a temporary fix for now?

sky333999: Lemme do that for both win & mac in a follow up so that we can revert easily once we know the root cause.

zhihonl (approved):

terraform/ec2/win/variable.tf
sky333999: Actually this will only fix the win feature test and not the win restat test... since the default can only be one of those two.
Mac already has the correct path it seems.

PR #266, opened 2023-06-22
Fix Nvidia GPU for Windows and failing linux tests

movence:
# Description of changes
- Run Nvidia GPU and non-feature tests with the validator (This is because we can't use SSH with Windows instance due to security issue, and `winrm` throws OOM error while running `go test` (during compile)
- Reduce metric collection interval for some linux tests due to missing metrics during validations
- Update structure of test when it is run by both `go test` and `validator*`:
```
*_test.go (basically executes test in *_unix.go
*_unix.go (simple validation or extends `base_test_runner`)
*_windows.go (has Validate(.*) func)
```
- Replace shell execution in `restart` test since `validator` binary doesn't package static files (eg. agent config or script files)
- Update Windows terraform script to add agent config file to a temp directory. A test that gets executed by `validator` will pick up that agent config then put it in the final cwa directory. 

# License

# Tests
https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/5339516492/jobs/9678547439


terraform/ec2/win/main.tf
```diff
@@ -170,7 +173,7 @@ resource "null_resource" "integration_test_run_validator" {
   count = length(regexall("/feature/windows", var.test_dir)) > 0 ? 1 : 0
   depends_on = [
     null_resource.integration_test_setup,
-    null_resource.integration_test_reboot,
+    null_resource.integration_test_wait,
```
nathalapooja, line 176: null_resource.integration_test_wait does this step add any value for feature/windows test?

movence: no, but `_wait` resource doesn't even run with feature test. It is there just for more clear dependency chain.

nathalapooja (dismissed):

internal/common/agent_util_windows.go
```diff
@@ -102,6 +102,8 @@ func RunShellScript(path string, args ...string) (string, error) {
 	shellArgs = append(shellArgs, path)
 	shellArgs = append(shellArgs, args...)
 
+	log.Printf("running %v", shellArgs)
```
SaxyPandaBear, line 105: super mild nit - it might make more sense to join the args into a string than try to print the slice. 
```suggestion
	log.Printf("running %s", strings.Join(" ", shellArgs))
```

SaxyPandaBear (commented):
I skipped over the nvidia changes

SaxyPandaBear (approved):
was waiting for the nvidia gpu change before approving. lgtm

PR #267, opened 2023-06-22
Prometheus ec2 windows integration test

dchappa:
# Description of the issue
Adds prometheus integration test

# Tests
Integration testing


terraform/ec2/win/main.tf
```diff
@@ -190,12 +190,24 @@ resource "null_resource" "integration_test_run_validator" {
     destination = module.validator.instance_validator_config
   }
 
+  //runs validator and sets up prometheus java agent
   provisioner "remote-exec" {
     inline = [
+      "mkdir C:\\jmx_workload",
+      "powershell.exe -Command \"'---', 'rules:', '- pattern: \\\".*\\\"' | Set-Content -Path \\\"C:\\jmx_workload\\exporter_config.yaml\\\"\"",
+      "powershell.exe -Command \"'global:', '  scrape_interval: 1m', '  scrape_timeout: 10s', 'scrape_configs:', '  - job_name: jmx-exporter', '    sample_limit: 10000', '    file_sd_configs:', '      - files: [ \\\"C:\\\\jmx_workload\\\\prometheus_file_sd.yaml\\\" ]' | Set-Content -Path \\\"C:\\jmx_workload\\prometheus.yaml\\\"\"",
```
SaxyPandaBear, line 198: Personally I'd rather write the YAML file into this test package, and then use powershell to copy the file over to where we want it.

terraform/ec2/win/main.tf
```diff
@@ -190,12 +190,24 @@ resource "null_resource" "integration_test_run_validator" {
...
+      "powershell.exe -Command \"'- targets:', '  - 127.0.0.1:9404', '  labels:', '    application: test-app', '    InstanceId: ${var.ec2_instance_type}' | Set-Content -Path \\\"C:\\jmx_workload\\prometheus_file_sd.yaml\\\"\"",
```
SaxyPandaBear, line 199: `InstanceId: ${var.ec2_instance_type}'` will this work?

SaxyPandaBear (approved):

terraform/ec2/win/main.tf
sky333999: +1 . I would add a backlog item to track updates needed for this. This is adding a new pattern on top of the many ones we already have.

test/feature/windows/agent_config.json
```diff
@@ -217,7 +217,32 @@
           }
       },
       "metrics_collected": {
-        "emf": { }
+        "emf": { },
```
sky333999, line 220: This is going to be odd.. we have a prometheus integ test already that now runs on linux but now this sits with the windows integ tests.
New developers might struggle to understand what sits where.

terraform/ec2/win/main.tf
sky333999, line 199: nit: This doesnt have to be in a separate sd file... can just inline as static targets.

terraform/ec2/win/main.tf
movence: totally agree. I would just have a powersehll script file (or other types) with this many commands that TF can write to a host then it's used by the test.

terraform/ec2/win/main.tf
```diff
@@ -190,12 +190,24 @@ resource "null_resource" "integration_test_run_validator" {
...
+      "powershell.exe -Command \"[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12\"",
+      "powershell.exe -Command \"(New-Object Net.WebClient).DownloadFile('https://cwagent-prometheus-test.s3-us-west-2.amazonaws.com/jmx_prometheus_javaagent-0.12.0.jar', 'C:\\\\jmx_workload\\\\jmx_prometheus_javaagent-0.12.0.jar')\"",
+      "powershell.exe -Command \"(New-Object Net.WebClient).DownloadFile('https://cwagent-prometheus-test.s3-us-west-2.amazonaws.com/SampleJavaApplication-1.0-SNAPSHOT.jar', 'C:\\\\jmx_workload\\\\SampleJavaApplication-1.0-SNAPSHOT.jar')\"",
```
sky333999, line 202: Id add to the backlog task to research where the source code for these mystery jars is, how we manage its publication to the s3 bucket, who owns it etc.
We dont want to scramble if there is some security incident in the future.

I personally prefer the linux setup that publishes the prometheus metrics on a port as part of the test setup itself and I dont need to go look in other places. For tests that we need to go look at, the only other main place I would look is for components in our image builder. But now this is a new one which abstracts away everything in a jar.

terraform/ec2/win/main.tf
movence: also curious how this works

sky333999 (approved):
Approving assuming we create a backlog task for the followups.

sky333999:
Mind posting a successful integ test run from GHAs?

terraform/ec2/win/main.tf
dchappa: Will create a task with this information, as well as the content from above.

dchappa:
> Mind posting a successful integ test run from GHAs?

Messaged privately

PR #271, opened 2023-06-22
Add terraform for onprem and refactor for reuse

ChenaLee:
# Description of the issue
Agent is currently not being tested on on-premise environment. 
(ECS and EKS with on-prem instances will be addressed in a separate PR)

# Description of changes
Created a new terraform module for onprem usecases, and since most of the file is redundant with the existing ec2 terraform module, refactored them out.

The agent treats a host as onprem instead of EC2 if IMDS is not reachable AND at the same time RUN_AS_AWS environment variable is not set to true. 

This spins up an EC2 instance with IMDS on on purpose even though there's an option to launch with IMDS off. The test run currently needs to ssh into the instance, and if the instance was launched without IMDS connection, for some reason ssh will never be possible. (Behavior not officially documented)
This also allows us to pull down a credential credential via assume-role, so that the agent can use the credentials and not rely on instance IAM role which requires IMDS connection. This particular IAM role has both policies that tests require AND the ones the agent requires, although it could be separated. 

# License

# Tests
1) Verified that the agent can start in on-prem environment. Verified that agent start when the ec2(ec2 mode), and after IMDS connection is turned off, agent cannot start anymore (onprem mode). After storing credentials, agent could start wihtout IMDS connection (onprem mode). Verified that metrics were queriable by the test.
Screenshot 2023-06-22 at 14 49 11

2) Verified that after the refactoring, existing EC2 tests are not broken.

Screenshot 2023-06-23 at 11 51 37

### DIFF
--- a/terraform/ec2/linux_onprem/main.tf
+++ b/terraform/ec2/linux_onprem/main.tf
@@ -67,11 +67,11 @@ resource "null_resource" "integration_test_setup" {
       "sudo mkdir -p /.aws",
       "echo creating credentials file that the agent uses by default for onprem",
       "printf '\n[profile AmazonCloudWatchAgent]\nregion = us-west-2' | sudo tee -a /.aws/config",
-      "printf '\n[AmazonCloudWatchAgent]\naws_access_key_id=%s\naws_secret_access_key=%s\naws_session_token=%s' $(aws sts assume-role --role-arn arn:aws:iam::506463145083:role/CloudWatchAgentServerRoleOnPrem --role-session-name onpremtest --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' --output text) | sudo tee -a /.aws/credentials>/dev/null",
+      "printf '\n[AmazonCloudWatchAgent]\naws_access_key_id=%s\naws_secret_access_key=%s\naws_session_token=%s' $(aws sts assume-role --role-arn ${module.common.cwa_onprem_assumed_iam_role_arm} --role-session-name onpremtest --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' --output text) | sudo tee -a /.aws/credentials>/dev/null",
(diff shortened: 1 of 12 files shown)


terraform/ec2/linux_onprem/main.tf
```diff
@@ -0,0 +1,111 @@
...
+      "printf '\n[AmazonCloudWatchAgent]\naws_access_key_id=%s\naws_secret_access_key=%s\naws_session_token=%s' $(aws sts assume-role --role-arn arn:aws:iam::506463145083:role/CloudWatchAgentServerRoleOnPrem --role-session-name onpremtest --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' --output text) | sudo tee -a /.aws/credentials>/dev/null",
+      "printf '[credentials]\n  shared_credential_profile = \"AmazonCloudWatchAgent\"\n  shared_credential_file = \"/.aws/credentials\"' | sudo tee /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml>/dev/null",
+      "echo write the same credentials as default profile as well. AWS SDK clients used for testing looks for default. Without this, would have needed to specify profile name in the test code",
+      "printf '\n[default]\nregion = us-west-2' | sudo tee -a ~/.aws/config",
+      "printf '\n[default]\naws_access_key_id=%s\naws_secret_access_key=%s\naws_session_token=%s' $(aws sts assume-role --role-arn arn:aws:iam::506463145083:role/CloudWatchAgentServerRoleOnPrem --role-session-name test --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' --output text) | sudo tee -a ~/.aws/credentials>/dev/null",
```
movence, line 74: This arn should be configurable?

movence (commented):
👍 for breaking resources into modules.

chadpatel (approved):

terraform/ec2/linux_onprem/main.tf
ChenaLee: I just made a common variable because doesn't sound very likely that we will need a different onprem iam role per test or something. If that happens somehow it will be super easy add a variable. (add to the variables file, add to the common module's variable file with default value as the current one, pass the onprem tf var into the common module. Not more work to do it now vs anytime in the future)

chadpatel (approved):

chadpatel (approved):

movence (approved):

PR #277, opened 2023-06-23
Add option to exclude test from specific OS

zhihonl:
# Description of the issue
Some plugins are not compatible for certain OS, therefore it would be valuable to not run those tests and prevent errors that can be confusing for people without context.

# Description of changes
Add feature to exclude tests by modifying the matrix and using `excludedTests` parameter

# License

# Tests
[Test run](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/5382141072/jobs/9767388314#step:7:2972) that removes `collectd` test from SLES12.


SaxyPandaBear (approved):

SaxyPandaBear (approved):

environment/metadata.go
```diff
@@ -166,6 +172,25 @@ func fillEC2PluginTests(e *MetaData, data *MetaDataStrings) {
...
+	if e.ComputeType != computetype.EC2 {
+		return
+	}
+
+	if len(data.ExcludedTests) == 0 {
+		log.Println("Testing all EC2 plugins")
+		return
+	}
+
+	plugins := strings.Split(strings.ReplaceAll(data.ExcludedTests, " ", ""), ",")
```
sethAmazon, line 185: is the space replace needed are we expecting 
a,b or a, b

sethAmazon (approved):

environment/metadata.go
zhihonl: We are expecting a,b. I folllowed the same pattern as `fillEC2PluginTests` method and it's probably to prevent human errors.

PR #278, opened 2023-06-24
Implement fix for sles linux integ-test

JayPolanco:
# Description of the issue
EC2 Linux SLES integ-tests were failing due to the test checking for the `docker0` interface dimension for the net plugin on all EC2 instances. 

SLES versions of Linux do not publish net metrics to `docker0`. They only publish to the `eth0` interface dimension.

# Description of changes
This change passes the OS using terraform, allowing us to check for the `eth0` dimension when it's an SLES version of Linux

```
 {
  "test_dir": "./test/metric_value_benchmark",
  "os": "sles-15",
  "family": "linux",
  "testType": "ec2_linux",
  "arc": "amd64",
  "instanceType": "t3a.medium",
  "ami": "cloudwatch-agent-integration-test-sles-15*",
  "binaryName": "amazon-cloudwatch-agent.rpm",
  "username": "ec2-user",
  "installAgentCommand": "go run ./install/install_agent.go rpm",
  "caCertPath": "/etc/ssl/certs/ca-bundle.crt",
  "values_per_minute": 0,
  "k8s_version": "",
  "terraform_dir": "",
  "useSSM": false
 },
```

# License

# Tests
Successfully tested by running the workflow

### DIFF
--- a/test/metric_value_benchmark/net_test.go
+++ b/test/metric_value_benchmark/net_test.go
@@ -6,16 +6,22 @@
 package metric_value_benchmark
 
 import (
-	"github.com/aws/aws-sdk-go-v2/aws"
+	"fmt"
+	"log"
+	"strings"
 
+	"github.com/aws/amazon-cloudwatch-agent-test/environment"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/metric"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/metric/dimension"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/status"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/test_runner"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
 )
 
 type NetTestRunner struct {
 	test_runner.BaseTestRunner
+	env *environment.MetaData
 }
 
 var _ test_runner.ITestRunner = (*NetTestRunner)(nil)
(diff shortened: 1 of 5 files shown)


test/metric_value_benchmark/net_test.go
```diff
@@ -53,16 +59,34 @@ func (m *NetTestRunner) validateNetMetric(metricName string) status.TestResult {
...
-		{
-			Key:   "InstanceId",
-			Value: dimension.UnknownDimensionValue(),
-		},
-	})
+	var (
+		dims   []types.Dimension
+		failed []dimension.Instruction
+	)
+	log.Println(fmt.Sprintf("OS VERSION IS %s", m.env.OS))
```
SaxyPandaBear, line 66: Why not just 
```suggestion
	log.Printf("OS VERSION IS %s\n", m.env.OS)
```

test/metric_value_benchmark/net_test.go
```diff
@@ -53,16 +59,34 @@ func (m *NetTestRunner) validateNetMetric(metricName string) status.TestResult {
...
+			{
+				Key:   "interface",
+				Value: dimension.ExpectedDimensionValue{aws.String("docker0")},
+			},
+			{
+				Key:   "InstanceId",
+				Value: dimension.UnknownDimensionValue(),
+			},
+		})
+	}
```
SaxyPandaBear, line 89: Frankly we could consolidate this since the only thing that differs is the name of the expected dimension. 
```go
var dimName string
if strings.HasPrefix(m.env.OS, "sles") {
  dimName = "eth0"
} else {
  dimName = "docker0"
}
dims, failed = m.DimensionFactory.GetDimensions([]dimension.Instruction{
	{
		Key:   "interface",
		Value: dimension.ExpectedDimensionValue{&dimName},
	},
	{
		Key:   "InstanceId",
		Value: dimension.UnknownDimensionValue(),
	},
})

SaxyPandaBear (approved):

sky333999:
I was curious how the other tests are working with the docker0 interface since that isnt something available typically by default. And just noticed we have a docker component baked into our AMIs - which is missing from the SLES ones. So we should update those and not need these OS level overrides in code.

PR #290, opened 2023-06-29
Onprem one test

ChenaLee:
# Description of the issue
Testing setup for onprem had been added, but had a bug introduced during conflict resolution process.

# Description of changes
Fixed the bug

# License

# Tests
Screenshot 2023-06-29 at 16 32 09


terraform/ec2/common/linux/output.tf
```diff
@@ -17,3 +17,6 @@ output "proxy_instance_proxy_ip" {
   value = module.proxy_instance.proxy_ip
 }
 
+output "cwa_onprem_assumed_iam_role_arm" {
```
SaxyPandaBear, line 20: nitpick - I think there's a typo in `arm`. I'm guessing this is supposed to be `arn`

terraform/ec2/creds/variables.tf
```diff
@@ -106,3 +106,9 @@ variable "plugin_tests" {
   type    = string
   default = ""
 }
+
+variable "agent_start" {
```
SaxyPandaBear, line 110: Does this `agent_start` variable need to be propagated to the Windows tf as well? I also don't see where these variables are applied or referenced. Is that supposed to be here, or in the main agent repo?

SaxyPandaBear (approved):

terraform/ec2/creds/variables.tf
SaxyPandaBear: nvm found it https://github.com/aws/amazon-cloudwatch-agent/pull/782

terraform/ec2/common/linux/output.tf
jefchien: Would need to change it in common https://github.com/aws/amazon-cloudwatch-agent-test/blob/1eb7bb1de72d9b765891906f08d23ffafe8ff927/terraform/common/output.tf#L12 as well.

jefchien (approved):

PR #296, opened 2023-07-07
Windows performance test setup

zhihonl:
# Description of the issue
Performance tests are set up for Linux system, however they do not support Windows yet. We need to evaluate CloudWatch Agent performance on Windows to give us a better idea of agent's utilization of resources.

# Description of changes
Add terraform and validator support for Windows performance test. The scope of this change is to only include `logs` and `system`

# License

# Tests
[Test run](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/5489557502/jobs/10004110995) showing all performance tests passing.

Log lines showing the calculation of performance statistics:
```
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric cpu_usage
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric cpu_usage: {Average:33.53327243921445 P99:76.9905054203211 Max:96.7837845092262 Min:16.673308504371974 Period:1 Std:12.177531176865374}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric memory_rss
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric memory_rss: {Average:59.69087239583333 P99:69.5078125 Max:69.70703125 Min:51.25 Period:1 Std:4.092167746198867}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric memory_vms
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric memory_vms: {Average:50.82248697916667 P99:60.28125 Max:61.0546875 Min:43.11328125 Period:1 Std:3.98359989165005}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric write_bytes
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric write_bytes: {Average:4.826704781850179 P99:8.626749038696289 Max:8.691282272338867 Min:0.9679603576660156 Period:1 Std:2.245980088326311}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric Bytes_Sent_Per_Sec
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric Bytes_Sent_Per_Sec: {Average:0.00375930160423[4982](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/5489557502/jobs/10004690002#step:6:4983) P99:0.08802559226751328 Max:0.1363559514284134 Min:0 Period:1 Std:0.01781574466805772}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric Packets_Sent_Per_Sec
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric Packets_Sent_Per_Sec: {Average:4.157455588579178 P99:40.18694305419922 Max:87.32308959960938 Min:0 Period:1 Std:10.174524843675071}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Start calculate metric statictics for metric Available_Bytes
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:00 Finished calculate metric statictics for metric Available_Bytes: {Average:14819.618736979166 P99:14892.24609375 Max:14893.15625 Min:14763.609375 Period:1 Std:34.70832982689886}
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:01 Test case: system_performance_windows, validate type: performance has been successfully validated
null_resource.validator_windows[0] (remote-exec): 2023/07/07 19:45:01 Finish validation in 7.626697118333333 minutes
```

Manually verified the data being published to the DynamoDB


validator/validators/performance/performance_validator.go
```diff
@@ -131,6 +147,42 @@ func (s *PerformanceValidator) CalculateMetricStatsAndPackMetrics(metrics []type
 	return packIntoPerformanceInformation(uniqueID, receiver, dataType, fmt.Sprint(agentCollectionPeriod), commitHash, commitDate, map[string]interface{}{dataRate: performanceMetricResults}), nil
 }
 
+func (s *PerformanceValidator) CalculateWindowsMetricStatsAndPackMetrics(statistic []*cloudwatch.GetMetricStatisticsOutput) (PerformanceInformation, error) {
+	var (
+		receiver               = s.vConfig.GetPluginsConfig()[0] //Assuming one plugin at a time
```
bryce-carey, line 152: Could this array ever be empty? I presume no.

validator/validators/performance/performance_validator.go
```diff
@@ -213,3 +321,15 @@ func isAllValuesGreaterThanOrEqualToZero(values []float64) bool {
 	}
 	return true
 }
+
+func isAllStatisticsGreaterThanOrEqualToZero(datapoints []types.Datapoint) bool {
+	if len(datapoints) == 0 {
+		return false
```
bryce-carey, line 327: If `datapoints` is empty then it's actually true that all statistics are non-negative since there are no data points that are negative. Is there a reason it needs to return `false` in this case?

validator/validators/performance/performance_stats.go
```diff
@@ -61,3 +62,45 @@ func CalculateMetricStatisticsBasedOnDataAndPeriod(data []float64, dataPeriod fl
 		Period:  int(dataPeriod / float64(length)),
 	}
 }
+
+func CalculateMetricStatisticsWindows(datapoints []types.Datapoint, dataPeriod float64) Stats {
```
sky333999, line 66: This is more or less the same as the func above - which is what Id expect given there is no windows specific logic here.
Can we just create the `data[]` in the caller of this and invoke the func above? That way we dont need this here.

validator/validators/performance/performance_validator.go
```diff
@@ -45,19 +46,34 @@ func NewPerformanceValidator(vConfig models.ValidateConfig) models.ValidatorFact
 }
 
 func (s *PerformanceValidator) CheckData(startTime, endTime time.Time) error {
-	metrics, err := s.GetPerformanceMetrics(startTime, endTime)
-	if err != nil {
-		return err
-	}
+	if strings.Contains(s.vConfig.GetTestCase(), "windows") {
+		stat, err := s.GetWindowsPerformanceMetrics(startTime, endTime)
```
sky333999, line 50: nit: Instead of having specific OS specific methods, might be cleaner to make these interface methods and have linux & windows implement each.

validator/validators/performance/performance_validator.go
```diff
@@ -45,19 +46,34 @@ func NewPerformanceValidator(vConfig models.ValidateConfig) models.ValidatorFact
 }
 
 func (s *PerformanceValidator) CheckData(startTime, endTime time.Time) error {
-	metrics, err := s.GetPerformanceMetrics(startTime, endTime)
-	if err != nil {
-		return err
-	}
+	if strings.Contains(s.vConfig.GetTestCase(), "windows") {
```
sky333999, line 49: I would pass in the os family using the same << >> approach so we dont just rely on the names here.

validator/validators/performance/performance_validator.go
```diff
@@ -45,19 +46,34 @@ func NewPerformanceValidator(vConfig models.ValidateConfig) models.ValidatorFact
...
-	}
+		perfInfo, err := s.CalculateMetricStatsAndPackMetrics(metrics)
+		if err != nil {
+			return err
+		}
 
-	err = s.SendPacketToDatabase(perfInfo)
-	if err != nil {
-		return err
+		err = s.SendPacketToDatabase(perfInfo)
```
sky333999, line 73: If this is common for both, can sit outside the if/else logic.

validator/validators/performance/performance_validator.go
```diff
@@ -131,6 +147,42 @@ func (s *PerformanceValidator) CalculateMetricStatsAndPackMetrics(metrics []type
...
+		if slices.Contains(metricsConvertToMB, metricName) {
+			for i, val := range metricValues {
+				*metricValues[i].Average = *val.Average / (1024 * 1024)
+			}
+		}
+		log.Printf("Start calculate metric statictics for metric %s \n", metricName)
+		if !isAllStatisticsGreaterThanOrEqualToZero(metricValues) {
+			return nil, fmt.Errorf("\n values are not all greater than or equal to zero for metric %s with values: %v", metricName, metricValues)
+		}
+		// GetMetricStatistics provides these statistics, however this will require maintaining multiple data arrays
```
sky333999, line 175: Does metric rename not work for these?

validator/validators/performance/performance_validator.go
bryce-carey: Can you explain what you mean by "<< >> approach"? I'm interested in doing this for the stress tests.

sky333999: Sure yes. Our parameters.yml files have token replacement logic ([example](https://github.com/aws/amazon-cloudwatch-agent-test/blob/ee898e8bc44f4be0a16495c752dbf7e3478526ba/test/stress/system/parameters.yml#L15)) that enables specifying placeholders in there which will be substituted with variables that can be passed in from terraform.
Hence we could use a similar pattern to pass in the os family which is already available as a terraform var when applying these stress tests.

validator/validators/performance/performance_validator.go
sky333999: Double checked offline that metric decoration apparently is not supported for procstat so this comment can be disregarded :(

validator/validators/performance/performance_validator.go
zhihonl: `GetPluginsConfig()` will get the `receivers` variable from the `parameters.yml` file, so it is possbile for the array to be empty if there was a human error to not input any value for the variable. But realistically no, the array should never be empty.

validator/validators/performance/performance_validator.go
zhihonl: If datapoints is empty then that means we did not get any metric which should throw an error.

generator/test_case_generator.go
```diff
@@ -116,11 +116,36 @@ var testTypeToTestConfig = map[string][]testConfig{
...
+		{
+			testDir: "../../test/performance/statsd",
+			targets: map[string]map[string]struct{}{"os": {"al2": {}}},
+		},
+		{
+			testDir: "../../test/performance/collectd",
+			targets: map[string]map[string]struct{}{"os": {"al2": {}}},
+		},
+		{
+			testDir: "../../test/performance_windows/logs",
```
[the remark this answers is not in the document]
zhihonl: I did your suggestion for Windows. I left Linux since that should be a separate PR by itself considering the amount of refactor it requires.

terraform/performance/main.tf
```diff
@@ -88,6 +88,7 @@ resource "null_resource" "install_binaries" {
   provisioner "remote-exec" {
     inline = [
       local.ami_family["wait_cloud_init"],
+      "aws s3 cp s3://${var.s3_bucket}/integration-test/packaging/${var.cwa_github_sha}/amazon-cloudwatch-agent.msi .",
```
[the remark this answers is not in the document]
zhihonl: Line 92 only installs the Linux binary. If you follow the same pattern it won't get the Windows binary from S3 which is why I added this line here. Terraform does not allow conditional statement within provisioner script so this is the simplest/cleanest solution I can think of.

test/performance_windows/logs/agent_config.json
```diff
@@ -0,0 +1,64 @@
...
+		"logs_collected": {
+		  "files": {
+			"collect_list": [
+			  {
+				"file_path": "/tmp/test1.log",
+				"log_group_name": "{instance_id}",
+				"log_stream_name": "{instance_id}/tmp1",
+				"timezone": "UTC"
+			  }
+			]
```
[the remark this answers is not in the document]
zhihonl: Formatted in new revision

validator/validators/performance/performance_stats.go
zhihonl: Removed this function. Now we generate data[] for `types.Datapoint` variable then just use the `CalculateMetricStatisticsBasedOnDataAndPeriod` function

validator/validators/performance/performance_validator.go
zhihonl: This may require refactor for Linux tests. We should do this in a separate PR.

validator/validators/performance/performance_validator.go
zhihonl: Implemented this in the new revision

validator/validators/performance/performance_validator.go
zhihonl: Implemented this in the new revision

validator/validators/performance/performance_validator.go
zhihonl: Not sure why this is the case. May be worth to investigate for the future since I can see this being a potential issue.

ymtaye (dismissed):
LGTM!

generator/test_case_generator.go
```diff
@@ -122,6 +122,10 @@ var testTypeToTestConfig = map[string][]testConfig{
 		{testDir: "../../test/performance/statsd"},
 		{testDir: "../../test/performance/collectd"},
 	},
+	"ec2_windows_performance": {
```
bryce-carey, line 125: Nice! I like that this is specific to the platform similar to other integration tests.

util/awsservice/cloudwatchmetrics.go
```diff
@@ -113,6 +113,40 @@ func GetMetricData(metricDataQueries []types.MetricDataQuery, startTime, endTime
 	return data, nil
 }
 
+func GetMetricStatistics(metricName, namespace string, dimensions []types.Dimension,
```
bryce-carey, line 116: nit: this could be simplified to
```
metricStatsInput := cloudwatch.GetMetricStatisticsInput{
	MetricName: aws.String(metricName),
	Namespace:  aws.String(namespace),
	StartTime:  aws.Time(startTime),
	EndTime:    aws.Time(endTime),
	Period:     aws.Int32(periodInSeconds),
	Dimensions: dimensions,
}
if extendedStatType == nil {
    metricStatsInput.Statistics = statType
} else {
    metricStatsInput.ExtendedStatistics = extendedStatType
}

return CwmClient.GetMetricStatistics(ctx, &metricStatsInput)
```

validator/validators/performance/performance_validator.go
bryce-carey: Considering these are integration tests and we would want to know that this is empty (panic) then this seems appropriate.

bryce-carey (approved):

bryce-carey (approved):

PR #302, opened 2023-07-11
Win2012 test fix

dchappa:
# Description of the issue
Resolves issue with windows 2012 acceptance and restart testing not passing

# Description of changes
The `-NoExit` parameter causes the command to not exit after finishing, ultimately timing out. I've removed the calls from `NoExit` from the functions that are called from these tests

# License

# Tests
Integration testing

### DIFF
--- a/terraform/ec2/win/main.tf
+++ b/terraform/ec2/win/main.tf
@@ -153,7 +153,7 @@ resource "null_resource" "integration_test_run_restart" {
     user     = "Administrator"
     password = rsadecrypt(aws_instance.cwagent.password_data, local.private_key_content)
     host     = aws_instance.cwagent.public_dns
-    timeout = "2m"
+    timeout = "5m"
   }
 
   provisioner "file" {
--- a/terraform/ec2/win/main.tf
+++ b/terraform/ec2/win/main.tf
@@ -140,12 +140,40 @@ resource "null_resource" "integration_test_wait" {
   ]
 }
 
-resource "null_resource" "integration_test_run" {
+resource "null_resource" "integration_test_run_restart" {
   # run go test when it's not feature test
-  count = length(regexall("/feature/windows", var.test_dir)) < 1 ? 1 : 0
+  count = length(regexall("restart", var.test_dir)) > 0 ? 1 : 0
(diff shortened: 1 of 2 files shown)


terraform/ec2/win/main.tf
```diff
@@ -153,6 +153,7 @@ resource "null_resource" "integration_test_run" {
     user     = "Administrator"
     password = rsadecrypt(aws_instance.cwagent.password_data, local.private_key_content)
     host     = aws_instance.cwagent.public_dns
+    timeout = "5m"
```
movence, line 156: nit: we can remove this since 5m is the default?

movence (approved):

terraform/ec2/win/main.tf
dchappa: Good point, was tinkering with the timeouts before. Recent commit addresses that.

terraform/ec2/win/main.tf
```diff
@@ -168,6 +168,34 @@ resource "null_resource" "integration_test_run" {
...
+
+  provisioner "file" {
+    source      = module.validator.agent_config
+    destination = module.validator.instance_agent_config
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "set AWS_REGION=${var.region}",
+      var.use_ssm ? "powershell \"& 'C:\\Program Files\\Amazon\\AmazonCloudWatchAgent\\amazon-cloudwatch-agent-ctl.ps1' -a fetch-config -m ec2 -s -c ssm:${local.ssm_parameter_name}\"" : "powershell \"& 'C:\\Program Files\\Amazon\\AmazonCloudWatchAgent\\amazon-cloudwatch-agent-ctl.ps1' -a fetch-config -m ec2 -s -c file:${module.validator.instance_agent_config}\"",
```
sky333999, line 193: I see we already start the agent here.. so why is this needed again?

dchappa: that may have been a holdover from me trying to root cause the issue. Running tests again after removing it

movence (approved):

sky333999 (approved):

PR #310, opened 2023-07-21
Revert "Migrate multi config tests (#245)"

ymtaye:
This reverts commit 37c3dc5f526c0d2e0a612d7213a8cf82d2125b6c.

# Description of the issue
Tests are failing for this change. Reverting commit while working resolving test failures

# Description of changes
Removed Failing tests (reverted commit)

# License

# Tests
Run tests in private repo gha


ZhenyuTan-amz (approved):

sethAmazon:
Should we fix instead of trying to revert?

zhihonl (commented):
Can you share the link with the test run that runs all tests so we can make sure this fixes the tests

ymtaye:
> Should we fix instead of trying to revert?

Yes, but I meant to remove this as a immediate resolution since it constantly contaminating integ tests

ymtaye:
> Can you share the link with the test run that runs all tests so we can make sure this fixes the tests
yes, this won't fix the Darwin tests however since this was failing for the run_as_user as cwagent testcase

sky333999 (approved):

PR #311, opened 2023-07-21
Add Base Name Tag To ECS ASG For Searching

sethAmazon:
# Description of the issue
Search asg by tag filter does not work with wildcards

# Description of changes
Add a base tag that is not generated

# License

# Tests
N/A

### DIFF
--- a/terraform/ecs_ec2/daemon/main.tf
+++ b/terraform/ecs_ec2/daemon/main.tf
@@ -11,7 +11,7 @@ module "basic_components" {
 }
 
 resource "aws_ecs_cluster" "cluster" {
-  name = "cwagent-integ-test-cluster-${module.common.testing_id}"
+  name = "${local.basename}-${module.common.testing_id}"
 }
 
 ######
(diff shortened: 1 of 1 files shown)


ymtaye (approved):

terraform/ecs_ec2/daemon/main.tf
```diff
@@ -11,7 +11,7 @@ module "basic_components" {
 }
 
 resource "aws_ecs_cluster" "cluster" {
-  name = "cwagent-integ-test-cluster-${module.common.testing_id}"
+  name = "${local.basename}-${module.common.testing_id}"
```
adam-mateen, line 14: I have no issue adding more tags.
Just curious if we could have just searched for existing `ClusterName` tags containing (or starting with)  `cwagent-integ-test-cluster`?

adam-mateen (approved):

terraform/ecs_ec2/daemon/main.tf
```diff
@@ -63,6 +63,12 @@ resource "aws_autoscaling_group" "cluster" {
     value               = ""
     propagate_at_launch = true
   }
+
+  tag {
+    key                 = "BaseClusterName"
+    value               = local.basename
+    propagate_at_launch = true
+  }
```
adam-mateen, line 71: nit - since the purpose of this new tag is to identify resources that can be "cleaned" I suggest the key name be "AutomaticCleanerGroup" or "AutomaticCleanerTag" (or anything that makes it clear to a random person that the tag is related to automatically cleaning up resources.

And then we could just look for the presence of the tag name. Setting the value might be useful if we want to just clean up resources for certain tests or test suites...

PR #314, opened 2023-07-27
Change basic validator to check the exact line count.

jefchien:
# Description of the issue
The basic validator is used by the windows feature test to validate metrics and logs. Currently, the log validation just checks that the number of actual log lines is greater than or equal to the number of expected log lines. This will not catch duplicate log lines.

# Description of changes
Makes it an equals check instead.

# License

### DIFF
--- a/validator/validators/basic/basic_validator.go
+++ b/validator/validators/basic/basic_validator.go
@@ -105,13 +105,13 @@ func (s *BasicValidator) Cleanup() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


bryce-carey (approved):

SaxyPandaBear (approved):

zhihonl (approved):

adam-mateen (approved):
The change makes sense. but I'd like to see a screenshot (or some proof) that

PR #318, opened 2023-07-28
Multi-System Trace Integration Test

okankoAMZ:
# Description of the issue
We want to test in multiple versions of linux in order to catch any system specific bugs that could rise. 

# Description of changes
I have added a trace testing structure to test both otlp and xray.  

# License

# Tests
Ran integration test to make sure all test passes except for test with sles-12 ami, which also fails on LinuxIntegrationTest.


sethAmazon (approved):

util/common/traces.go
```diff
@@ -0,0 +1,98 @@
...
+	GetContext() context.Context
+	GetAgentRuntime() time.Duration
+	GetName() string
+}
+
+func TraceTest(t *testing.T, traceTest TracesTestInterface) error {
+	startTime := time.Now()
+	t.Logf("AgentConfigpath %s", traceTest.GetAgentConfigPath())
+	CopyFile(traceTest.GetAgentConfigPath(), ConfigOutputPath)
+	require.NoError(t, StartAgent(ConfigOutputPath, true, false), "Couldn't copy file")
```
jefchien, line 40: nit: This isn't the copy file function, so the log would be misleading.

util/common/traces.go
```diff
@@ -0,0 +1,98 @@
...
+	t.Logf("Successfully copied file.")
```
jefchien, line 41: nit: Do we need all of these logs?

util/common/traces.go
```diff
@@ -0,0 +1,98 @@
...
+	go func() {
+		require.NoError(t, traceTest.StartSendingTraces(context.Background()), "load generator exited with error")
+	}()
+	time.Sleep(traceTest.GetAgentRuntime())
+	traceTest.StopSendingTraces()
+	t.Logf("Stopped Traces")
+	StopAgent()
+	t.Logf("Stopped Agent")
+	testsGenerated, testsEnded := traceTest.GetTestCount()
+	t.Logf("For %s , Test Cases Generated %d | Test Cases Ended: %d", traceTest.GetName(), testsGenerated, testsEnded)
```
jefchien, line 52: These aren't test cases. These are spans/segments generated.

test/otlp/generator.go
```diff
@@ -0,0 +1,139 @@
...
+
+const (
+	serviceName                    = "load-generator"
+	attributeKeyAwsXrayAnnotations = "aws.xray.annotations"
+)
+
+type OtlpTracesGenerator struct {
+	common.TracesTestInterface
+	cfg                     *common.TraceConfig
+	testCasesGeneratedCount int
```
jefchien, line 30: nit: This isn't a test case count. It's a trace generated count.

test/otlp/trace_test.go
```diff
@@ -0,0 +1,68 @@
...
+				},
+			},
+		},
+	}
+	t.Logf("Sanity check: number of test cases:%d", len(testCases))
+	for name, testCase := range testCases {
+
+		t.Run(name, func(t *testing.T) {
+
+			OtlpTest := newLoadGenerator(testCase.generatorConfig)
```
jefchien, line 58: nit: Can we keep the generator and test case parameters separate?

util/common/traces.go
```diff
@@ -0,0 +1,98 @@
...
+	validateSegments(t *testing.T, segments []types.Segment, config *TraceConfig)
+	GetTestCount() (int, int)
+	GetAgentConfigPath() string
+	GetGeneratorConfig() *TraceConfig
+	GetContext() context.Context
+	GetAgentRuntime() time.Duration
+	GetName() string
+}
+
+func TraceTest(t *testing.T, traceTest TracesTestInterface) error {
```
jefchien, line 36: Should call
```go
t.Helper()
```
at the start of this helper function so it'll map back to the calling function when printing out the failures.

jefchien (commented):
Do you have a link to the passing test run that's mentioned in the description?

generator/test_case_generator.go
```diff
@@ -137,6 +137,10 @@ var testTypeToTestConfig = map[string][]testConfig{
 		{testDir: "../../test/stress/windows/logs"},
 		{testDir: "../../test/stress/windows/system"},
 	},
+	"ec2_trace_integration": {
+		{testDir: "./test/xray"},
```
jefchien, line 141: You'll need to remove this as well.

generator/test_case_generator.go
```diff
@@ -137,6 +137,9 @@ var testTypeToTestConfig = map[string][]testConfig{
 		{testDir: "../../test/stress/windows/logs"},
 		{testDir: "../../test/stress/windows/system"},
 	},
+	"ec2_trace_integration": {
```
jefchien, line 140: Could the test just be added to the ec2 linux block? Is there anything specific to the trace matrix that is preventing us from doing that?

util/common/agent_util_unix.go
```diff
@@ -41,10 +43,14 @@ func CopyFile(pathIn string, pathOut string) {
 	}
 
 	log.Printf("File %s abs path %s", pathIn, pathInAbs)
-	out, err := exec.Command("bash", "-c", "sudo cp "+pathInAbs+" "+pathOut).Output()
-
+	cmd := exec.Command("bash", "-c", "sudo cp "+pathInAbs+" "+pathOut)
```
jefchien, line 46: Why are we changing this?

okankoAMZ: It gives us more information about the execution. So instead of saying the error is "exit status 1" it will actually give an error like "config file doesn't exist". I think it is a valid improvement

test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -156,6 +156,7 @@ func (e *EKSDaemonTestRunner) validateLogs(env *environment.MetaData) status.Tes
 
 				if !awsservice.MatchEMFLogWithSchema(l, rs, validateLogContents) {
 					log.Println("failed to match log with schema")
+					log.Printf("log entry %s json schema %s", l, jsonSchema)
```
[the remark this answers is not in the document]
okankoAMZ: yeah i did is that okay?

test/xray/trace_test.go
```diff
@@ -0,0 +1,67 @@
...
+				},
+			},
+		},
+	}
+	for name, testCase := range testCases {
+
+		t.Run(name, func(t *testing.T) {
+
+			XrayTest := newLoadGenerator(testCase.generatorConfig)
+			XrayTest.AgentConfigPath = testCase.agentConfigPath
```
jefchien, line 58: nit: Would still prefer to have the test parameters separate from the load generator.

jefchien (approved):

jefchien (approved):

sethAmazon (approved):

PR #320, opened 2023-08-01
Add duplicate logs check.

jefchien:
# Description of the issue
Need to be able to detect duplicate logs in the integration tests.

# Description of changes
Adds a duplicate log check to the basic validator. Checks that there aren't any logs for the same timestamp and message. Refactored the log validation a bit to evaluate over LogEvents instead of the log messages to retain the timestamp.

# License

# Tests
Ran the integration tests and failed on the duplicates.

### DIFF
--- a/test/ecs/ecs_metadata/ecs_metadata_test.go
+++ b/test/ecs/ecs_metadata/ecs_metadata_test.go
@@ -7,12 +7,12 @@ import (
 	_ "embed"
 	"flag"
 	"fmt"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs/types"
 	"log"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/qri-io/jsonschema"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/aws/amazon-cloudwatch-agent-test/util/awsservice"
@@ -36,8 +36,6 @@ var clusterName = flag.String("clusterName", "", "Please provide the os preferen
 var schema string
 
 func TestValidatingCloudWatchLogs(t *testing.T) {
-	rs := jsonschema.Must(schema)
-
 	start := time.Now()
 
 	logGroupName := fmt.Sprintf(ECSLogGroupNameFormat, *clusterName)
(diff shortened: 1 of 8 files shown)


test/ecs/ecs_metadata/ecs_metadata_test.go
```diff
@@ -57,25 +55,25 @@ func TestValidatingCloudWatchLogs(t *testing.T) {
 
 		end := time.Now()
 
-		ok, err := awsservice.ValidateLogs(logGroupName, LogStreamName, &start, &end, func(logs []string) bool {
```
adam-mateen, line 60: Having an integration test for ECS metadata which relies on the agent's own log getting uploaded to CWL is not very robust. Log messages can change at a developer's whim...
But if there are no better options then okay.

util/awsservice/cloudwatchlogs.go
```diff
@@ -161,15 +168,97 @@ func GetLogStreams(logGroupName string) []types.LogStream {
...
+			if !ok {
+				messages = map[string]struct{}{}
+				byTimestamp[timestamp] = messages
+			}
+			_, ok = messages[message]
+			if !ok {
+				messages[message] = struct{}{}
+				continue
+			}
+			return fmt.Errorf("duplicate message found at %v | message: %s", time.UnixMilli(timestamp), message)
```
adam-mateen, line 260: nit: we could avoid the continue by doing:
```
_, ok = messages[message]
if ok {
	return fmt.Errorf("duplicate message found at %v | message: %s", time.UnixMilli(timestamp), message)
}
messages[message] = struct{}{}
```

util/awsservice/cloudwatchlogs.go
```diff
@@ -161,15 +168,97 @@ func GetLogStreams(logGroupName string) []types.LogStream {
...
+		if len(events) != count {
+			return fmt.Errorf("actual log events count (%v) does not match expected (%v)", len(events), count)
+		}
+		return nil
+	}
+}
+
+func AssertNoDuplicateLogs() LogEventsValidator {
+	return func(events []types.OutputLogEvent) error {
+		byTimestamp := make(map[int64]map[string]struct{})
```
adam-mateen, line 246: I guess there could be an agent bug reporting the same event multiple times with different timestamps... Like rereading a file or rereading an event log. But it is hard (or impossible) to distinguish this scenario from an event that legitimately repeats at different times.

util/awsservice/cloudwatchlogs.go
```diff
@@ -161,15 +168,97 @@ func GetLogStreams(logGroupName string) []types.LogStream {
...
+		if err != nil {
+			return fmt.Errorf("failed to execute schema validator: %w", err)
+		} else if len(keyErrors) > 0 {
+			return fmt.Errorf("failed schema validation: %v | schema: %s | log: %s", keyErrors, schema, message)
+		}
+		return nil
+	}
+}
+
+func AssertLogSubstring(substr string) LogEventValidator {
```
adam-mateen, line 203: nit: The name did not make it clear if this was asserting the log contains the substring or does not.
We could call it `AssertLogContainsSubstring`

util/awsservice/cloudwatchlogs.go
jefchien: Yeah we repeat the same logs quite often for tests, so it'd be harder to catch that.

adam-mateen (approved):

sky333999 (approved):

okankoAMZ (approved):

PR #322, opened 2023-08-01
ADD ECS EMF test without EC2 metadata on

zhihonl:
# Description of the issue
We want to to test if EMF still works with EC2 metadata turned off. 

# Description of changes
Add `metadataEnabled` variable to have test case where EC2 metadata is turned off for ECS integration test.

# License

# Tests
Test run: https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/5731828918/job/15533934402
Test failed when metadata is disabled. We need to investigate this and have a separate PR to address the issue.


terraform/ecs_ec2/daemon/main.tf
```diff
@@ -37,7 +37,7 @@ resource "aws_launch_configuration" "cluster" {
 
   user_data = "#!/bin/bash\necho ECS_CLUSTER=${aws_ecs_cluster.cluster.name} >> /etc/ecs/ecs.config"
   metadata_options {
-    http_endpoint               = "enabled"
+    http_endpoint               =  var.metadataEnabled
```
sethAmazon, line 40: How do we tell ecs metadata is up? Have we tested with each version of ecs metadata?

generator/resources/ecs_ec2_daemon_test_matrix.json
```diff
@@ -2,6 +2,13 @@
   {
     "instanceType":"t3a.xlarge",
     "ami": "cloudwatch-agent-integration-test-al2*",
-    "family": "linux"
+    "family": "linux",
+    "metadataEnabled": "enabled"
```
sethAmazon, line 6: I'm assuming targets means if it contains this then this can be a test case.

sethAmazon (dismissed):

generator/test_case_generator.go
```diff
@@ -31,6 +31,7 @@ type matrixRow struct {
 	TerraformDir        string `json:"terraform_dir"`
 	UseSSM              bool   `json:"useSSM"`
 	ExcludedTests       string `json:"excludedTests"`
+	MetadataEnabled     string `json:"metadataEnabled"`
```
bryce-carey, line 34: Will this impact other tests where the parameter is not specified? I presume we want to have this enabled by default nearly everywhere to preserve the current tests.

terraform/ecs_ec2/daemon/main.tf
```diff
@@ -225,6 +226,7 @@ resource "aws_ecs_service" "extra_apps_service" {
   task_definition = aws_ecs_task_definition.extra_apps_task_definition.arn
   desired_count   = 1
   launch_type     = "FARGATE"
+  enable_execute_command = true
```
zhihonl, line 229: Added `enable_execute_command` for debug purpose in case we ever need to run command  on the container itself.

terraform/ecs_ec2/daemon/main.tf
```diff
@@ -243,5 +245,20 @@ resource "null_resource" "validator" {
...
   }
-  depends_on = [aws_ecs_service.cwagent_service, aws_ecs_service.extra_apps_service]
+  depends_on = [aws_ecs_service.cwagent_service, aws_ecs_service.extra_apps_service, null_resource.disable_metadata]
 }
+
+resource "null_resource" "disable_metadata" {
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      echo Sleep for 30 seconds to allow instance to be attached to the cluster
```
zhihonl, line 255: When metadata is turned off immediately, ECS fails to attach the instance to the task so we will always fail.

generator/test_case_generator.go
zhihonl: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/5863584354/job/15897572781 
None of the other tests use this variable so it should not affect them. I ran a full test just in case and it does not show a change of behavior for other tests. 

Note: Failed tests are due to multi-config issues.

generator/resources/ecs_ec2_daemon_test_matrix.json
zhihonl: Yes

bryce-carey (approved):

sethAmazon (approved):

PR #324, opened 2023-08-02
Add EMF UDP TCP Logs CA Bundle Tests

sethAmazon:
# Description of the issue
emf upd/tcp has no ca bundle tests

# Description of changes
Add tests for emf udp/tcp ca bundle

# License

# Tests
Integration test


test/ca_bundle/ca_bundle_test.go
```diff
@@ -29,20 +29,21 @@ const (
...
-	// Let the agent run for 30 seconds. This will give agent enough time to call server
-	agentRuntime         = 30 * time.Second
 	localstackS3Key      = "integration-test/ls_tmp/%s"
 	keyDelimiter         = "/"
 	localstackConfigPath = "../../localstack/ls_tmp/"
 	originalPem          = "original.pem"
 	combinePem           = "combine.pem"
 	snakeOilPem          = "snakeoil.pem"
 	tmpDirectory         = "/tmp/"
+	runEMF               = "sudo bash resources/emf.sh"
```
sethAmazon, line 39: I believe this should be a file instead of built into the ami like we assume in our other tests

util/common/agent_util_unix.go
```diff
@@ -180,16 +180,22 @@ func StopAgent() {
 	log.Printf("Agent is stopped")
 }
 
-func ReadAgentOutput(d time.Duration) string {
+func ReadAgentLogfile(logfile string) string {
```
sethAmazon, line 183: I find it easier to work with the log file than journald for debugging issues in my env. I can change this back if need be would rather not.

jefchien:
Does https://github.com/aws/amazon-cloudwatch-agent-test/pull/321 need to be closed?

test/ca_bundle/resources/with/original/bundle/emf/common-config.toml
jefchien, line 1: These are a bunch of the same files. Could we move them up a few levels and have one for the `resources/with` and one for `resources/without`?

test/ca_bundle/resources/with/combine/bundle/emf/config.json
jefchien, line 1: Do we need to have separate files for each combination? It looks like all the `emf` and `metric` configs are the same. Can we just have one copy of each?

util/common/agent_util_unix.go
```diff
@@ -180,16 +180,22 @@ func StopAgent() {
...
+	out, err := os.ReadFile(logfile)
+	if err != nil {
+		log.Fatal(fmt.Sprint(err) + string(out))
+	}
+	return string(out)
+}
+
+func RecreateAgentLogfile(logfile string) {
 	out, err := exec.Command("bash", "-c",
```
jefchien, line 192: nit: I know this was done before, but why are we using bash for something as simple as removing and recreating a file? Can't we just use golang for this?

test/ca_bundle/ca_bundle_test.go
```diff
@@ -58,28 +59,35 @@ func TestBundle(t *testing.T) {
...
-		log.Printf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget)
-		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget), func(t *testing.T) {
-			common.ReplaceLocalStackHostName(parameter.dataInput + configJSON)
-			t.Logf("config file after localstack host replace %s", string(readFile(parameter.dataInput+configJSON)))
-			common.CopyFile(parameter.dataInput+configJSON, configOutputPath)
-			common.CopyFile(parameter.dataInput+commonConfigTOML, commonConfigOutputPath)
+		pathDir := parameter.dataInput + "/" + parameter.testType
+		log.Printf("resource file location %s find target %t", pathDir, parameter.findTarget)
+		t.Run(fmt.Sprintf("resource file location %s find target %t", pathDir, parameter.findTarget), func(t *testing.T) {
+			common.RecreateAgentLogfile(logfile)
```
jefchien, line 80: Do we need to recreate it? Can't we just delete it? The agent should recreate it on start up.

test/ca_bundle/resources/with/combine/bundle/emf/config.json
sethAmazon: No since there is a http type. They are not all the same.

test/ca_bundle/ca_bundle_test.go
sethAmazon: No the agent does not do that on start up. If there is no log file it will not start.

util/common/agent_util_unix.go
sethAmazon: It is simpler to run a single command, but this also assumes we have bash access, so maybe doing it with golang commands is better.

test/ca_bundle/resources/with/original/bundle/emf/common-config.toml
sethAmazon: reduced amount of files

test/ca_bundle/ca_bundle_test.go
jefchien: The agent will create a new log file if it doesn't exist on start up. I've tested this.

>Logger opens or creates the logfile on first Write.

https://github.com/natefinch/lumberjack

util/common/agent_util_unix.go
jefchien: Ah right. Different file permissions.

test/ca_bundle/ca_bundle_test.go
sethAmazon: I remember this being an issue in the past. When was lumberjack added?

jefchien (approved):

test/ca_bundle/resources/http/metric/config.json
```diff
@@ -2,8 +2,7 @@
   "agent": {
     "metrics_collection_interval": 10,
     "run_as_user": "root",
-    "debug": true,
-    "logfile": ""
+    "debug": true
```
okankoAMZ, line 5: do we want to leave debug on ?

okankoAMZ (approved):

test/ca_bundle/resources/http/metric/config.json
sethAmazon: Yes helps with debugging

PR #326, opened 2023-08-03
Macos integration test error fix

zhihonl:
# Description of the issue
macOS test is failing because we accidentally added workflow and test path that are not needed for macOS test.

# Description of changes
Remove the unnecessary workflow.

# License

# Tests
Test run showing all macOS tests passing. https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/5745725949/job/15576333770

### DIFF
--- a/generator/test_case_generator.go
+++ b/generator/test_case_generator.go
@@ -105,7 +105,6 @@ var testTypeToTestConfig = map[string][]testConfig{
 	*/
 	"ec2_mac": {
 		{testDir: "../../../test/feature/mac"},
-		{testDir: "../../../test/run_as_user"},
 	},
 	"ec2_windows": {
 		{testDir: "../../../test/feature/windows"},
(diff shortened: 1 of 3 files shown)


generator/test_case_generator.go
```diff
@@ -105,7 +105,6 @@ var testTypeToTestConfig = map[string][]testConfig{
 	*/
 	"ec2_mac": {
 		{testDir: "../../../test/feature/mac"},
-		{testDir: "../../../test/run_as_user"},
```
zhihonl, line 108: We don't need this. macOS has to use validator so we can avoid starting up macOS instances that are more than needed due to limit macOS quota. Instead we manually run the run_as_user go test on one host,.

terraform/ec2/mac/main.tf
```diff
@@ -135,32 +135,6 @@ resource "null_resource" "integration_test" {
   }
 
 }
-
```
zhihonl, line 138: This looks like an accidental bad merge. We don't need this since the previous `resource` does the actual integration test.

PR #328, opened 2023-08-09
Fix bug in duplicate check.

jefchien:
# Description of the issue
Currently, it returns the duplicate error if it doesn't find the message for the timestamp. This means it'll fail on the first log message. It should be the opposite.

# Description of changes
Changed the condition.

# License

# Tests
Checked the log groups for a failing test.

### DIFF
--- a/util/awsservice/cloudwatchlogs.go
+++ b/util/awsservice/cloudwatchlogs.go
@@ -253,7 +253,7 @@ func AssertNoDuplicateLogs() LogEventsValidator {
 				byTimestamp[timestamp] = messages
 			}
 			_, ok = messages[message]
-			if !ok {
+			if ok {
 				return fmt.Errorf("duplicate message found at %v | message: %s", time.UnixMilli(timestamp), message)
 			}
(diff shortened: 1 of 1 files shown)


util/awsservice/cloudwatchlogs.go
```diff
@@ -253,7 +253,7 @@ func AssertNoDuplicateLogs() LogEventsValidator {
 				byTimestamp[timestamp] = messages
 			}
 			_, ok = messages[message]
-			if !ok {
+			if ok {
```
okankoAMZ, line 256: This seems like we are just skipping the test?

adam-mateen (approved):

util/awsservice/cloudwatchlogs.go
jefchien: The test checks to see if there are duplicate log events by creating a map of the timestamp to messages. If it finds a message within the same timestamp that's been previously set in this map, then it'll return this error, which means the test should fail.

okankoAMZ (approved):

PR #330, opened 2023-08-14
Xray performance test

okankoAMZ:
# Description of the issue
We were missing an automated performance gathering for CloudWatchAgent's new feature, traces. 

# Description of changes
This PR is adding  a performance test for xray traces by adding a mock http server to minimize wasted cloud resources. This is added to the already built in performance test, and does no require any user input. The test will create a mock server automatically  and run trace generation on it.

# License

# Tests
Ran Performance Test and checked on Dynamo
https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/6223948452/job/17362419385


generator/test_case_generator.go
```diff
@@ -31,14 +31,16 @@ type matrixRow struct {
 	TerraformDir        string `json:"terraform_dir"`
 	UseSSM              bool   `json:"useSSM"`
 	ExcludedTests       string `json:"excludedTests"`
+	RunMockServer       bool   `json:"run_mock_server"`
 	MetadataEnabled     string `json:"metadataEnabled"`
 }
 
 type testConfig struct {
 	// this gives more flexibility to define terraform dir when there should be a different set of terraform files
```
SaxyPandaBear, line 39: nit - the comment doesn't seem directly tied to `testDir`. I see it was like this before, but just wanted to point out that it looks kind of confusing to see this group of vars that aren't all related to terraform

generator/test_case_generator.go
```diff
@@ -121,6 +123,8 @@ var testTypeToTestConfig = map[string][]testConfig{
 		{testDir: "../../test/performance/system"},
 		{testDir: "../../test/performance/statsd"},
 		{testDir: "../../test/performance/collectd"},
+		{testDir: "../../test/performance/trace/xray",
+			runMockServer: true},
```
SaxyPandaBear, line 127: Check formatting here. I am not convinced that the linter will want this on a new line

mockserver/README
```diff
@@ -0,0 +1,3 @@
+# Mocked Servers
+
+The servers themselves are very basic. They listen for a message and tend to set a flag to success upon receiving it. Typically, they are set up with two servers: one to mock the backend and one to report on the status of that mock (if endpoint is called, how many times, etc.). There's a built-in 15ms of latency between each message that's received. The status reporter will always be on port 8080 and support the `:8080/` (returns "healthcheck") and the `:8080/check-data` (returns "success" if mock has received messages) endpoints.
```
SaxyPandaBear, line 3: nit - I don't think that this documentation is worded in a way that helps people understand how it's used. I think a more procedural guide on running the mocked tests, or a step-by-step guide on what happens would illustrate it better. Avoid words like "typically".

mockserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+		if err := daemonServer.ListenAndServeTLS(CertFilePath, KeyFilePath); err != nil {
+			log.Fatalf("HTTPS server error: %v", err)
+			err = daemonServer.Shutdown(context.TODO())
+			log.Fatalf("Shutdown server error: %v", err)
+		}
+	}(&store)
+
+	go func(ts *transactionStore) {
+		defer close(serverControlChan)
+		verifyApp.HandleFunc("/", healthCheck)
```
SaxyPandaBear, line 117: nit - Typically if you want a healthcheck endpoint, you just want a `/ping` endpoint. Doing it at the root is kind of bad code smell to me.

mockserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+	startTime    time.Time
+}
+
+type TransactionPayload struct {
+	TransactionsPerMinute float64 `json:"tpm"`
+}
+
+func healthCheck(w http.ResponseWriter, _ *http.Request) {
+	if _, err := io.WriteString(w, HealthCheckMessage); err != nil {
+		log.Printf("Unable to write response: %v", err)
```
SaxyPandaBear, line 52: mild nitpick - if this fails, it will still return a 200 status code

mockserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+// Retrieve number of transactions per minute
+func (ts *transactionStore) tpm(w http.ResponseWriter, _ *http.Request) {
+	// Calculate duration in minutes
+	duration := time.Now().Sub(ts.startTime)
+	transactions := float64(atomic.LoadUint32(&ts.transactions))
+	tpm := transactions / duration.Minutes()
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(TransactionPayload{tpm}); err != nil {
+		io.WriteString(w, err.Error())
```
SaxyPandaBear, line 87: mild nitpick - if this fails, it will still return a 200 status code

mockserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+		log.Printf("Unable to write response: %v", err)
+	}
+}
+
+// Starts an HTTPS server that receives requests for the data handler service at the sample server port
+// Starts an HTTP server that receives request from validator only to verify the data ingestion
+func StartHttpServer() chan interface{} {
+	serverControlChan := make(chan interface{})
```
SaxyPandaBear, line 95: I don't think that we should use one control channel and share it across two goroutines. What you want here is a `sync.WaitGroup` and then after you start both goroutines, call `Wait()` on the WaitGroup until both routines finish.

mockserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+	verifyApp := http.NewServeMux()
+	appServer := &http.Server{Addr: ":8080", Handler: verifyApp}
+	go func(ts *transactionStore) {
+		defer close(serverControlChan)
+		dataApp.HandleFunc("/", healthCheck)
+		dataApp.PathPrefix("/put-data").HandlerFunc(ts.dataReceived)
+		dataApp.HandleFunc("/trace/v1", ts.dataReceived)
+		dataApp.HandleFunc("/metric/v1", ts.dataReceived)
+		if err := daemonServer.ListenAndServeTLS(CertFilePath, KeyFilePath); err != nil {
+			log.Fatalf("HTTPS server error: %v", err)
```
SaxyPandaBear, line 109: Are you sure we want to do a fatal log here? If we `Fatal` here, I don't think it will execute anything after it.

mockserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+	return serverControlChan
+}
+
+func main() {
+	server := StartHttpServer()
+	select {
+	case <-server:
+		fmt.Println("server is dead")
+
+	}
```
SaxyPandaBear, line 145: TBH I don't think this is super useful if we just switch to a sync group and wait call.

util/common/traces.go
```diff
@@ -112,3 +112,13 @@ func SegmentValidationTest(t *testing.T, traceTest TraceTestConfig, segments []t
 	return nil
 
 }
+func GenerateTraces(traceTest TraceTestConfig) error {
+	CopyFile(traceTest.AgentConfigPath, ConfigOutputPath)
```
jefchien, line 116: This never starts the agent after copying the file over.

validator/validators/basic/basic_validator.go
```diff
@@ -49,6 +49,8 @@ func (s *BasicValidator) GenerateLoad() error {
 	switch dataType {
 	case "logs":
 		return common.StartLogWrite(agentConfigFilePath, agentCollectionPeriod, metricSendingInterval, dataRate)
+	case "traces":
+		return util.StartTraceGeneration(receiver,agentConfigFilePath,agentCollectionPeriod,metricSendingInterval)
```
jefchien, line 53: nit: It looks like the log generation is in common. Consider moving this there as well for consistency.

util/common/traces.go
```diff
@@ -112,3 +112,13 @@ func SegmentValidationTest(t *testing.T, traceTest TraceTestConfig, segments []t
...
+	go func() {
+		traceTest.Generator.StartSendingTraces(context.Background())
+	}()
+	time.Sleep(traceTest.AgentRuntime)
+	traceTest.Generator.StopSendingTraces()
+	time.Sleep(AGENT_SHUTDOWN_DELAY)
+	return nil
```
jefchien, line 123: Why does `GenerateTraces` return an error when it's only ever nil? Same question for the other functions in this file.

test/xray/generator.go
```diff
@@ -55,6 +56,7 @@ func newLoadGenerator(cfg *common.TraceGeneratorConfig) *XrayTracesGenerator {
 }
 func (g *XrayTracesGenerator) Generate(ctx context.Context) error {
 	rootCtx, root := xray.BeginSegment(ctx, "load-generator")
+	log.Println("\033[34mGenerated Trace\033[0m")
```
jefchien, line 59: This seems like it's going to pollute the logs especially for high loads (e.g. 5000).

terraform/performance/main.tf
```diff
@@ -109,6 +109,14 @@ resource "null_resource" "validator_linux" {
   }
   provisioner "remote-exec" {
     inline = [
+      #mock server dependencies getting transfered.
+      # todo make this optional with a env var.
+      "git clone https://github.com/aws/amazon-cloudwatch-agent-test.git",
+      "cd amazon-cloudwatch-agent-test && git checkout xray-performance-test",
```
jefchien, line 115: I know this was for testing, but for the PR, this should be checking out `main`.

terraform/performance/main.tf
```diff
@@ -109,6 +109,14 @@ resource "null_resource" "validator_linux" {
...
+      var.run_mock_server ? "cd mockserver && sudo docker build -t mockserver . && cd .." : "echo skipping mock server build",
+      var.run_mock_server ? "sudo docker run --name mockserver -d -p 8080:8080 -p 443:443  mockserver" : "echo skipping mock server",
```
jefchien, line 117: nit:
```suggestion
      var.run_mock_server ? "sudo docker run --name mockserver -d -p 8080:8080 -p 443:443  mockserver" : "echo skipping mock server run",
```

terraform/performance/main.tf
```diff
@@ -109,6 +109,14 @@ resource "null_resource" "validator_linux" {
...
+      "cd ..",
+      "cp -r amazon-cloudwatch-agent-test/test/xray/resources /home/ec2-user/",
```
jefchien, line 119: What is this for? How is it being used?

terraform/performance/main.tf
```diff
@@ -109,6 +109,14 @@ resource "null_resource" "validator_linux" {
   }
   provisioner "remote-exec" {
     inline = [
+      #mock server dependencies getting transfered.
+      # todo make this optional with a env var.
+      "git clone https://github.com/aws/amazon-cloudwatch-agent-test.git",
```
jefchien, line 114: Look at how our other terraform scripts do this. They all use a `github_test_repo` variable.

```
git clone --branch ${var.github_test_repo_branch} ${var.github_test_repo}
```

test/performance/trace/xray/parameters.yml
```diff
@@ -0,0 +1,70 @@
...
+agent_collection_period: 300 
+
+commit_hash: <commit_hash>
+commit_date: <commit_date>
+
+cloudwatch_agent_config: "<cloudwatch_agent_config>"
+
+# Metric that the test needs to validate
+metric_namespace: "CloudWatchAgentPerformance"
+metric_validation:
```
jefchien, line 18: So are we just validating that the metrics exist?

test/performance/trace/xray/agent_config.json
```diff
@@ -0,0 +1,110 @@
...
+    "traces": {
+        "endpoint_override" : "https://127.0.0.1/put-data",
+        "traces_collected": {
+          "xray": {
+          }
+        }
+      },
+      "insecure": true,
+      "metrics": {
+        "namespace": "CloudWatchAgentPerformance",
```
jefchien, line 16: Shouldn't this be in a different namespace since the metrics are specific to this test?

mockserver/main.go
```diff
@@ -0,0 +1,133 @@
...
+		io.WriteString(w, err.Error())
+		log.Printf("Unable to write response: %v", err)
+	}
+}
+
+func (ts *transactionStore) dataReceived(w http.ResponseWriter, _ *http.Request) {
+	atomic.AddUint32(&ts.transactions, 1)
+
+	// Built-in latency
+	fmt.Printf("\033[31m Time: %d | data Received \033[0m \n", time.Now().Unix())
```
jefchien, line 74: nit: Mixture of `fmt` and `log`.

mockserver/main.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+func (ts *transactionStore) checkData(w http.ResponseWriter, _ *http.Request) {
+	var message string
+	var t = atomic.LoadUint32(&ts.transactions)
+	if t > 0 {
+		message = SuccessMessage
+	}
+	fmt.Printf("\033[31m Time: %d | checkData msg: %s | %d\033[0m \n", time.Now().Unix(), message, t)
```
jefchien, line 63: Is this log necessary?

test/performance/trace/xray/agent_config.json
okankoAMZ: It wasn't for the other performance tests but I can change it

util/common/traces.go
okankoAMZ: For the other function they fail if there is an error so if you reach the end there is not error.  I understand what you mean tho. Those function could be changed to void.

util/common/traces.go
okankoAMZ: That function handles only trace generation, and the validator is the that actually starts the agent.  If it makes it more coherent  I can move the CopyFile there as well.

terraform/performance/main.tf
```diff
@@ -109,6 +109,13 @@ resource "null_resource" "validator_linux" {
   }
   provisioner "remote-exec" {
     inline = [
+      #mock server dependencies getting transfered.
+      "git clone --branch ${var.github_test_repo_branch} ${var.github_test_repo}",
+      "cd amazon-cloudwatch-agent-test && git checkout xray-performance-test", #@TODO remove before PR merge
```
SaxyPandaBear, line 114: is this TODO relevant?

util/common/traces/common/common.go
SaxyPandaBear, line 1: We should be cognizant about the file structure we are introducing. Does this need to be `util/common/traces/common/common`? What is the reason we are moving the file to begin with? Maybe a broader question, if it's supposed to be "common", why is it under `/traces`?

util/common/traces/generate.go
```diff
@@ -0,0 +1,43 @@
...
+				"custom_key": "custom_value",
+			},
+		},
+	}
+	switch receiver {
+	case "xray":
+		cfg.Generator = xray.NewLoadGenerator(&xrayGenCfg)
+		cfg.Name = "xray-performance-test"
+	case "otlp":
+		panic("Only supports xray for now.")
```
SaxyPandaBear, line 37: I can't see any good reason to panic. Just return an error instead. 
```suggestion
		return errors.New("OTLP is not supported")
```
Also nitpick - "for now" will become stale as soon as you commit this. Be more concise. The issue isn't that we only support XRay. The issue is we don't support OTLP.

mockserver/main.go
```diff
@@ -0,0 +1,134 @@
...
+		io.WriteString(w, err.Error())
+		log.Printf("Unable to write response: %v", err)
+	}
+}
+
+// Starts an HTTPS server that receives requests for the data handler service at the sample server port
+// Starts an HTTP server that receives request from validator only to verify the data ingestion
+func StartHttpServer() {
+	var wg sync.WaitGroup
+	log.Println("\033[31m Starting Server \033[0m")
```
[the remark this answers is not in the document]
okankoAMZ: It helps me distinguish when the server started easier. It is not necessary  but a nice to have imo

terraform/performance/main.tf
okankoAMZ: yup! i am waiting for the PR to be ready before I remove it so I can test the code before the merge

mockserver/main.go
```diff
@@ -0,0 +1,134 @@
...
+	daemonServer := &http.Server{Addr: ":443", Handler: dataApp}
+	verifyApp := http.NewServeMux()
+	appServer := &http.Server{Addr: ":8080", Handler: verifyApp}
+	go func(ts *transactionStore) {
+		wg.Add(1)
+		defer wg.Done()
+		dataApp.HandleFunc("/ping", healthCheck)
+		dataApp.PathPrefix("/put-data").HandlerFunc(ts.dataReceived)
+		dataApp.HandleFunc("/trace/v1", ts.dataReceived)
+		dataApp.HandleFunc("/metric/v1", ts.dataReceived)
```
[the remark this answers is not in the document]
okankoAMZ: i think it makes it slightly easier to read personally

mockserver/main.go
```diff
@@ -124,7 +125,7 @@ func StartHttpServer() {
 			log.Fatalf("Shuwdown server error: %v", err)
```
jefchien, line 125: nit: 
```suggestion
			log.Fatalf("Shutdown server error: %v", err)
```

terraform/performance/main.tf
jefchien: This should use the `github_test_repo_branch` var. You can specify it in your test workflow.

util/common/traces/generate.go
jefchien: If we're not supporting it, then we can just have the "otlp" case fallthrough to the default. There's no reason to call it out.

```suggestion
   default:
		return fmt.Errorf("%s is not supported.", receiver)
```

mockserver/main.go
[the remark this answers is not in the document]
SaxyPandaBear: Distinguish where? When looking at output logs from GitHub actions? Like I said, I'm not opposed to it. It just feels unnecessary

mockserver/main.go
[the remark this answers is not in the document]
SaxyPandaBear: We can agree to disagree on that one 😛 I'll live with it.

util/common/traces/common/common.go
SaxyPandaBear: Still would like an answer for this

util/common/traces/common/common.go
```diff
@@ -3,15 +3,15 @@ package common
...
-
 	"github.com/aws/amazon-cloudwatch-agent-test/util/awsservice"
+	"github.com/aws/amazon-cloudwatch-agent-test/util/common"
 	"github.com/aws/aws-sdk-go-v2/service/xray/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otel/attribute"
+	"reflect"
+	"testing"
+	"time"
```
SaxyPandaBear, line 14: looks like a formatting thing. Might need to lint this.

util/common/traces/base/base.go
```diff
@@ -1,17 +1,17 @@
-package common
+package base
```
SaxyPandaBear, line 1: I wasn't necessarily just suggesting to rename this 😅 but I do want you to think about the intent behind it. If you told someone to look at the path `/util/common/traces/common/common` or `/util/common/traces/base/base`, will they have a sense of what to expect in that file path? I won't block the review on it, but just want you to think about that as you add/move files

SaxyPandaBear (dismissed):

mockserver/main.go
```diff
@@ -0,0 +1,141 @@
...
+	if _, err := io.WriteString(w, message); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		io.WriteString(w, err.Error())
+		log.Printf("Unable to write response: %v", err)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (ts *transactionHttpServer) acknowledgeTransaction(w http.ResponseWriter, _ *http.Request) {
```
jefchien, line 75: nit: `acknowledgeTransaction` seems inaccurate. It's incrementing a counter. Consider something like `recordTransaction` or even `incrementTransactionCount`

mockserver/main.go
```diff
@@ -0,0 +1,141 @@
...
+	var wg sync.WaitGroup
+	log.Println("\033[31m Starting Server \033[0m")
+	store := transactionHttpServer{startTime: time.Now()}
+	//2 servers one for receiving the data , one for verify data
+	dataApp := mux.NewRouter()
+	dataReceiverServer := &http.Server{Addr: ":443", Handler: dataApp}
+	verificationRequestServer := http.NewServeMux()
+	appServer := &http.Server{Addr: ":8080", Handler: verificationRequestServer}
+	go func(ts *transactionHttpServer) {
+		wg.Add(1)
```
jefchien, line 110: Incrementing wait groups should be done outside of the go routine.

>Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for.

https://pkg.go.dev/sync#WaitGroup.Add

mockserver/main.go
```diff
@@ -0,0 +1,141 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(TransactionPayload{tpm}); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		io.WriteString(w, err.Error())
+		log.Printf("Unable to write response: %v", err)
+	}
+}
+
+// Starts an HTTP server that receives request from validator only to verify the data ingestion
+func StartHttpServer() {
```
jefchien, line 100: Is this used elsewhere? Why isn't this just main?

validator/models/validation_config.go
```diff
@@ -15,7 +15,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-var supportedReceivers = []string{"logs", "statsd", "collectd", "system", "emf"}
+var supportedReceivers = []string{"logs", "statsd", "collectd", "system", "emf","xray"}
```
jefchien, line 18: nit: Looks like you need to run the formatter.

jefchien (dismissed):
Some suggestions, but nothing blocking.

util/common/traces/base/base.go
jefchien: I don't have a problem with the file having the same name as the package. That's pretty common with go. I agree though that `common`, `util`, and `base` are interchangeably used and lose meaning very quickly especially when used together. I'm guilty of using `common` in place of `util`.

mockserver/main.go
SaxyPandaBear: I'm not opposed to having the start function separate from main. It's redundant here, though.

okankoAMZ: I was using that for testing. I also like having a separate "main" function and have the argument  flags in the func main instead.

jefchien (approved):

mockserver/README.md
```diff
@@ -0,0 +1,41 @@
...
+    - `/put-data/metrics`: Use this sub-route for sending metrics data.
+
+> [!Important]
+> Currently, both traces and metrics are handled in the same way.
+
+### The Verifier
+
+The verifier component can be accessed via a listener on port 8080. It provides information about the transactions, including:
+
+- **Transactions per Minute:** You can obtain the transactions per minute by making a request to `/tpm`.
```
SaxyPandaBear, line 35: super mild nitpick, it would be nice to have an example cURL request, like
```bash
curl localhost:8080/tpm

# the output is [ something]
```

This is really trivial, imo, but something that is really helpful to document.

SaxyPandaBear (approved):

PR #352, opened 2023-09-29
Add E2E test for APM on EC2

bjrara:
# Description of the issue
This PR adds E2E test for APM on EC2.

# Description of changes
In the test case, we
* install CloudWatch agent with APM enabled;
* start an application with ADOT auto-instrumentation java agent;
* send requests to the newly brought up application to generate traces and metrics;
* check CloudWatch to verify whether APM metrics are created.

# License

# Tests
Tested through running integration test: https://github.com/aws/private-amazon-cloudwatch-agent-staging/actions/runs/6330956859/job/17204689789

### DIFF
--- a/terraform/ec2/apm/main.tf
+++ b/terraform/ec2/apm/main.tf
@@ -1,0 +1,108 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+module "common" {
+  source = "../../common"
+}
+
+module "basic_components" {
+  source = "../../basic_components"
+
+  region = var.region
+}
+
+#####################################################################
+# Generate EC2 Key Pair for log in access to EC2
+#####################################################################
+
+resource "tls_private_key" "ssh_key" {
+  count     = var.ssh_key_name == "" ? 1 : 0
+  algorithm = "RSA"
+  rsa_bits  = 4096
+}
+
+resource "aws_key_pair" "aws_ssh_key" {
+  count      = var.ssh_key_name == "" ? 1 : 0
+  key_name   = "ec2-key-pair-${module.common.testing_id}"
+  public_key = tls_private_key.ssh_key[0].public_key_openssh
+}
+
+locals {
+  ssh_key_name        = var.ssh_key_name != "" ? var.ssh_key_name : aws_key_pair.aws_ssh_key[0].key_name
+  private_key_content = var.ssh_key_name != "" ? var.ssh_key_value : tls_private_key.ssh_key[0].private_key_pem
+}
+
+#####################################################################
+# Generate EC2 Instance and execute test commands
+#####################################################################
+resource "aws_instance" "cwagent" {
+  ami                                  = data.aws_ami.latest.id
+  instance_type                        = var.ec2_instance_type
+  key_name                             = local.ssh_key_name
+  iam_instance_profile                 = module.basic_components.instance_profile
+  vpc_security_group_ids               = [module.basic_components.security_group]
+  associate_public_ip_address          = true
+  instance_initiated_shutdown_behavior = "terminate"
+
+  tags = {
+    Name = "apm-ec2-${module.common.testing_id}"
+  }
+}
+
+data "aws_ami" "latest" {
+  most_recent = true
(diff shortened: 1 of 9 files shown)


test/apm_ec2/resources/run_java_application.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+JAVA_TOOL_OPTIONS=" -javaagent:${HOME}/aws-apm-0.3.0.jar" \
```
majanjua-amzn, line 3: Is there another option than to hardcode the java APM version? Also, if this is a CW agent test that uses the java agent, doesn't it make more sense as an E2E test?

bjrara: The problem here is that ADOT only provides numbered versions, there's no `latest` jar file available, so even if we make it configurable, we still need to come back to this repo and update it.

But your concern makes sense, making it hard coded in the script is not ideal. I would recommend we revisit this after RIV when we have an official place to download the binary.

> doesn't it make more sense as an E2E test?
Sorry, not quite following, it is E2E test, right? We use java agent to generate Pulse attributes so that we can verify that the `awsapmprocessor` in the CW agent translates these attributes correctly.

lisguo: The `amazon-cloudwatch-agent-test` repo is meant for agent integration tests, not "e2e" tests that were setup in the main repo. 

This repo tests the agent by itself by crafting an otlp json request to the agent and expects some metric/trace.

If this test uses the java agent for testing, I'd recommend having this as part of the current e2e tests. @majanjua-amzn  would you agree?

terraform/ec2/apm/main.tf
```diff
@@ -0,0 +1,108 @@
...
+  connection {
+    type        = "ssh"
+    user        = var.user
+    private_key = local.private_key_content
+    host        = aws_instance.cwagent.public_ip
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "aws s3 cp s3://apm-beta-pre-release-test/e2e/aws-apm-0.3.0.jar ./",
```
lisguo, line 97: can these paths be variables in variables.tf?

terraform/ec2/apm/main.tf
```diff
@@ -0,0 +1,108 @@
...
+      "aws s3 cp s3://apm-beta-pre-release-test/e2e/spring-petclinic-api-gateway-2.6.7.jar ./",
+      "export AWS_REGION=${var.region}",
+      "export PATH=$PATH:/snap/bin:/usr/local/go/bin",
+      "echo run pulse e2e test",
```
lisguo, line 101: nit: "echo run apm e2e test"

test/apm_ec2/apm_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	flag.StringVar(&path, "path", "./resources/run_java_application.sh", "path to java script file.")
+	flag.StringVar(&testId, "test.id", "unknown", "testing id.")
+}
+
+type APMEC2TestRunner struct {
+	test_runner.BaseTestRunner
+}
+
+func (runner *APMEC2TestRunner) SetupAfterAgentRun() error {
```
lisguo, line 43: I'm assuming this is just checking that the sample app is working? Maybe needs a more clear function name like `SetupSampleJavaApplication()` or something

test/apm_ec2/apm_test.go
```diff
@@ -0,0 +1,147 @@
...
+			log.Printf("successfully sent out the request.")
+		}
+	}
+
+	if !success {
+		return errors.New("failed to call the test service")
+	}
+
+	// Sleep 1 minute to let metrics and traces be exported.
+	time.Sleep(90 * time.Second)
```
lisguo, line 89: nit: have the timer as a const.

test/apm_ec2/resources/run_java_application.sh
majanjua-amzn: Agree with Lisa, although if this accomplishes the goal of the test we can leave it as is for now and come back to it later when we're working on moving the E2E tests to a centralized location.

PR #377, opened 2023-11-10
Add win user data test

sethAmazon:
# Description of the issue
We do not have a windows user data test. 

# Description of changes
Add a windows userdata test

# License

# Tests
Ran integration tests 
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/6801520646/job/18568980590


generator/test_case_generator.go
```diff
@@ -122,6 +122,10 @@ var testTypeToTestConfig = map[string][]testConfig{
 		{testDir: "../../../test/restart"},
 		{testDir: "../../../test/acceptance"},
 		{testDir: "../../../test/feature/windows/event_logs"},
+		{
```
sethAmazon, line 125: I don't think we need to run this on all platforms. 
The latest "win-2022" has both a test for ssm and non ssm.
Meaning this will be tested the same 2 times.
This test will only use ssm since it makes the ps1 command simpler.

terraform/ec2/win/main.tf
```diff
@@ -72,18 +73,20 @@ resource "aws_instance" "cwagent" {
   tags = {
     Name = "cwagent-integ-test-ec2-windows-${var.test_name}-${module.common.testing_id}"
   }
+  depends_on = [aws_ssm_parameter.upload_ssm]
 }
 
 # Size of windows json is too large thus can't use standard tier
 resource "aws_ssm_parameter" "upload_ssm" {
-  count = var.use_ssm == true && length(regexall("/feature/windows", var.test_dir)) > 0 ? 1 : 0
+  count = length(regexall("/feature/windows", var.test_dir)) > 0 ? 1 : 0
```
sethAmazon, line 81: We can upload the value to ssm for all feature tests since it will be deleted by terraform destroy anyway. This will allow us to use ssm for userdata test.

terraform/ec2/win/main.tf
```diff
@@ -221,3 +281,19 @@ data "aws_ami" "latest" {
...
 }
+
+#####################################################################
+# Generate template file for EC2 userdata script
+#####################################################################
+data "template_file" "user_data" {
+  template = file("install_and_start_agent.tpl")
+
+  vars = {
+    copy_object = "Copy-S3Object -BucketName ${var.s3_bucket} -Key integration-test/packaging/${var.cwa_github_sha}/amazon-cloudwatch-agent.msi -region ${var.region} -LocalFile $cwAgentInstaller"
```
sethAmazon, line 292: For some reason var replacement was not working for the internal vars of this command and started working when I added the entire command as the var.

test/feature/windows/userdata/agent_config.json
```diff
@@ -0,0 +1,216 @@
...
+          "metrics_collection_interval": 1
+        }
+      ]
+    },
+    "append_dimensions": {
+      "InstanceId": "${aws:InstanceId}"
+    },
+    "force_flush_interval": 30
+  },
+  "logs": {
```
sethAmazon, line 160: While not testing logs the validator will fail if there is no logs config.

test/feature/windows/userdata/parameters.yml
```diff
@@ -0,0 +1,157 @@
...
+values_per_minute: "2"
+# Number of seconds the agent should run and collect the metrics. In this case, 1 minutes
+agent_collection_period: 60 
+
+cloudwatch_agent_config: "<cloudwatch_agent_config>"
+
+# Metric that the test needs to validate; moreover, the stress validation already has
+# InstanceID dimension; therefore, does not need to validate it
+# https://github.com/aws/amazon-cloudwatch-agent-test/pull/109/files#diff-47c87373e751dd9fd5ce504e44b320765c8b84d6cde524a4e8a32cfa34674165R124-R135
+metric_namespace: "CloudWatchAgentWinFeature"
```
sethAmazon, line 21: Only test metrics since this is more of a sanity test that we can install and start the agent via userdata thus no need to test logs.

test/app_signals/resources/traceid_generator.go
```diff
@@ -17,4 +17,4 @@ func main() {
 	binary.BigEndian.PutUint32(r[0:4], uint32(epochNow))
 	rand.Read(r[4:])
 	fmt.Printf("%s", hex.EncodeToString(r[:]))
-}
```
sethAmazon, line 20: This come from running make fmt

Makefile
```diff
@@ -47,6 +47,7 @@ install-golang-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(TOOLS_BIN_DIR) v1.50.1
 
 fmt: install-goimports addlicense
+	terraform fmt -recursive
```
sethAmazon, line 50: I don't want to have to run this command in terminal when I get terraform fmt error. I should be able to run it via make.

terraform/ec2/win/install_and_start_agent.tpl
```diff
@@ -0,0 +1,22 @@
+<powershell>
+$installDirectory = "c:\temp\cw"
```
adam-mateen, line 2: does this need double backslash? 
Same question on the `$cwAgentInstaller` variable.

terraform/ec2/win/install_and_start_agent.tpl
```diff
@@ -0,0 +1,22 @@
...
+$cwAgentInstallPath = "C:\\Program Files\\Amazon\\AmazonCloudWatchAgent"
+
+New-Item -ItemType "directory" -Path $installDirectory
+
+Set-Location -Path $installDirectory
+
+Write-host "Installing Cloudwatch Agent"
+Install-PackageProvider NuGet -Force;
+Set-PSRepository PSGallery -InstallationPolicy Trusted
+Install-Module -Name AWS.Tools.EC2,AWS.Tools.S3 -AllowClobber
```
adam-mateen, line 15: Can you add a comment as to why we need these tools installed?

terraform/ec2/win/main.tf
```diff
@@ -213,6 +237,42 @@ resource "null_resource" "integration_test_run_validator" {
...
 
+resource "null_resource" "integration_test_run_validator_userdata" {
+  # run validator only when test_dir is not passed e.g. the default from variable.tf
+  count = length(regexall("/feature/windows/userdata", var.test_dir)) > 0 ? 1 : 0
+  depends_on = [
+    null_resource.integration_test_setup_validator,
+    null_resource.integration_test_wait,
+  ]
+
+  connection {
```
adam-mateen, line 248: If we are using EC2 "userdata" to launch the instance and install the agent, then do we still need winrm access to the instance?

In other words, after launching a Windows EC2 instance with the agent installed with user-data, what additional steps occur?

adam-mateen (dismissed):

terraform/ec2/win/install_and_start_agent.tpl
sethAmazon: Not when it is in the tpl file.

terraform/ec2/win/install_and_start_agent.tpl
sethAmazon: This is for the aws cli commands. Powershell uses this instead of the normal cli.

terraform/ec2/win/main.tf
sethAmazon: This runs the validator, it does not reinstall / restart the agent. The validator runs on the ec2 instance.

terraform/ec2/win/install_and_start_agent.tpl
sethAmazon: Added

adam-mateen (approved):

okankoAMZ (approved):

PR #400, opened 2024-04-09
Changing data type of appsignals integ test

Paramadon:
# Description of the issue
Current server_consumer.json and client_producer.json generates invalid app signal metrics. The metrics Latency, Error, and Fault should be an object but instead the request sent to the agent generated the metrics to just be zero.

### Metrics are not objects
Screenshot 2024-04-09 at 3 13 50 PM

# Description of changes
It is a really small change, the data type of each metric is sum but instead it should be the type exponential_histogram. This is the correct type and it generates correct app signal metrics.

### Verifying correct metrics were generated
Screenshot 2024-04-09 at 3 11 54 PM

# License

# Tests
Sent json's to the agent and checked the console which had the correct format of the metrics.


test/app_signals/resources/metrics/client_producer.json
```diff
@@ -34,8 +34,8 @@
           "metrics": [
             {
               "name": "Error",
-              "unit": "Milliseconds",
-              "sum": {
+              "unit": "",
+              "exponential_histogram": {
```
okankoAMZ, line 38: Whats the difference between exponential histogram and sum

test/app_signals/resources/metrics/client_producer.json
```diff
@@ -34,8 +34,8 @@
           "metrics": [
             {
               "name": "Error",
-              "unit": "Milliseconds",
-              "sum": {
+              "unit": "",
```
okankoAMZ, line 37: Why did we have Milliseconds as a unit before but not now ?

test/app_signals/resources/metrics/server_consumer.json
```diff
@@ -261,17 +261,17 @@
                     ],
                     "startTimeUnixNano": START_TIME,
                     "timeUnixNano": START_TIME,
-                    "sum": 1,
-                    "min": 1,
-                    "max": 1
+                    "sum": 0,
```
okankoAMZ, line 264: Why are we changing these just curious

test/app_signals/resources/metrics/client_producer.json
Paramadon: The units were incorrect before, there are no units for app signal error/fault

test/app_signals/resources/metrics/client_producer.json
Paramadon: exponential_histogram is the data type that app signal uses for its metrics. We had the wrong data type which is why the metrics were generating invalid metrics

test/app_signals/resources/metrics/server_consumer.json
Paramadon: This is to reflect the default app signals metrics

zhihonl:
Is this being used in one of the integration tests? If so can you add the test run so we know it's working

lisguo:
Let's run the integ test a few times and see if the apm team is still affected by this. We should probably ping Mahad to take a look at this as well

okankoAMZ (approved):

jefchien (approved):

PR #405, opened 2024-04-23
Extend CloudWatch Agent AppSignals EC2 tests for MacOS and Window

Paramadon:
# Description of the issue
Extending Linux App signal EC2 Test to work on MacOs and Windows. 

# Description of changes
Changing the validator, which is what MacOs and Winows use for their testing, to have a test for app signals. Both App Signals and Traces are being tested.

### Running Integration test 
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/8820034431/job/24213490204

### Why are some test failing?
If you look at the logs of the test, the validator is passing, those are different issues not correlated with this pr.

### Steps taken for testing
1. Add AppSignal Metrics to parameters.yml file (validator uses this to know which metrics to validate, in this case we added "Latency", "Error, and "Fault")
2. Generate AppSignals/Traces Metrics 
3. Verify AppSignals/Traces Metrics

### Additional change needed to be made
Corrected the json files to generate AppSignals metrics so that App Signal metrics are being generated properly

# License

# Tests


util/awsservice/constant.go
```diff
@@ -38,16 +39,41 @@ var (
...
-	EcsClient            = ecs.NewFromConfig(awsCfg)
-	SsmClient            = ssm.NewFromConfig(awsCfg)
-	ImdsClient           = imds.NewFromConfig(awsCfg)
-	CwmClient            = cloudwatch.NewFromConfig(awsCfg)
-	CwlClient            = cloudwatchlogs.NewFromConfig(awsCfg)
-	DynamodbClient       = dynamodb.NewFromConfig(awsCfg)
-	S3Client             = s3.NewFromConfig(awsCfg)
-	CloudformationClient = cloudformation.NewFromConfig(awsCfg)
-	XrayClient           = xray.NewFromConfig(awsCfg)
+	ctx context.Context
```
okankoAMZ, line 42: Why are we changing all of them to pointers?

util/awsservice/constant.go
```diff
@@ -38,16 +39,41 @@ var (
...
+	ImdsClient           *imds.Client
+	CwmClient            *cloudwatch.Client
+	CwlClient            *cloudwatchlogs.Client
+	DynamodbClient       *dynamodb.Client
+	S3Client             *s3.Client
+	CloudformationClient *cloudformation.Client
+	XrayClient           *xray.Client
 )
+
+func init() {
```
okankoAMZ, line 57: why do we want to initialise separately  instead as variables like before

util/common/metrics.go
```diff
@@ -27,14 +38,71 @@ func StartSendingMetrics(receiver string, duration, sendingInterval time.Duratio
...
+
 		default:
 		}
-
 	}()
 
 	return err
 }
 
+func SendAppTraceMetrics(duration time.Duration) error {
```
okankoAMZ, line 53: nit: for consistency we should name this function `SendAppSignalsTraceMetrics`

util/common/metrics.go
```diff
@@ -27,14 +38,71 @@ func StartSendingMetrics(receiver string, duration, sendingInterval time.Duratio
...
+	baseDir := getBaseDir()
+
+	for i := 0; i < int(duration/(5*time.Second)); i++ {
```
okankoAMZ, line 56: For this I think using [TimeTicker](https://pkg.go.dev/time#NewTicker) would be easier to maintain and will be less flakey.  Here is an [example](https://github.com/aws/amazon-cloudwatch-agent-test/blob/d4e556e0a638a1b8f6fd9b1f763535a08322b816/util/common/traces/xray/generator.go#L28-L41) from traces-test

util/common/metrics.go
```diff
@@ -27,14 +38,71 @@ func StartSendingMetrics(receiver string, duration, sendingInterval time.Duratio
...
+		}
+
+		time.Sleep(5 * time.Second)
+	}
+
+	return nil
+}
+
+func getBaseDir() string {
+	if runtime.GOOS == "windows" {
```
okankoAMZ, line 74: nit: you might want to use https://pkg.go.dev/path/filepath

util/common/metrics.go
```diff
@@ -27,14 +38,71 @@ func StartSendingMetrics(receiver string, duration, sendingInterval time.Duratio
...
+		return "C:\\Users\\Administrator\\amazon-cloudwatch-agent-test\\test\\app_signals\\resources\\traces"
+	}
+	return "/Users/ec2-user/amazon-cloudwatch-agent-test/test/app_signals/resources/traces"
+}
+
+func generateTraceID() [16]byte {
```
okankoAMZ, line 80: can we use UUID instead?

util/common/metrics.go
```diff
@@ -27,14 +38,71 @@ func StartSendingMetrics(receiver string, duration, sendingInterval time.Duratio
...
+func processTraceFile(filePath string, startTime int64, traceID string) error {
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+		return err
+	}
+
+	modifiedData := strings.ReplaceAll(string(data), "START_TIME", fmt.Sprintf("%d", startTime))
+	modifiedData = strings.ReplaceAll(modifiedData, "TRACE_ID", traceID)
+
+	url := "http://127.0.0.1:4316/v1/traces"
```
okankoAMZ, line 97: nit: might want to make the port a variable for later use if we want to change it

util/common/metrics.go
```diff
@@ -130,6 +198,71 @@ func SendCollectDMetrics(metricPerInterval int, sendingInterval, duration time.D
...
+	if err != nil {
+		fmt.Println("Error reading file:", err)
+		return nil
+	}
+
+	//replace START_TIME with the current time
+	modifiedData := strings.ReplaceAll(string(data), "START_TIME", fmt.Sprintf("%d", startTime))
+
+	//curl command
+	url := "http://127.0.0.1:4316/v1/metrics"
```
okankoAMZ, line 213: nit: might want to make the port a variable for later use if we want to change it

util/common/metrics.go
```diff
@@ -130,6 +198,71 @@ func SendCollectDMetrics(metricPerInterval int, sendingInterval, duration time.D
...
+	dir, err := os.Getwd()
+	if err != nil {
+		fmt.Println("Error getting current directory:", err)
+		return err
+	}
+	fmt.Println("Current Directory:", dir)
+
+	// Determine the base directory for the files based on the OS
+	var baseDir string
+	if runtime.GOOS == "windows" {
```
okankoAMZ, line 237: nit: you might want to use https://pkg.go.dev/path/filepath

util/common/metrics.go
```diff
@@ -130,6 +198,71 @@ func SendCollectDMetrics(metricPerInterval int, sendingInterval, duration time.D
...
+		baseDir = "C:\\Users\\Administrator\\amazon-cloudwatch-agent-test\\test\\app_signals\\resources\\metrics"
+	} else { // assuming macOS or Unix-like system
+		baseDir = "/Users/ec2-user/amazon-cloudwatch-agent-test/test/app_signals/resources/metrics"
+	}
+
+	for i := 0; i < 12; i++ {
```
okankoAMZ, line 243: why 12? How did we come up with this number?

validator/validators/basic/basic_validator.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"go.uber.org/multierr"
 
+	apMetrics "github.com/aws/amazon-cloudwatch-agent-test/test/metric"
```
okankoAMZ, line 17: nit: can we call this `AppSignalMetrics` apMetrics seems a bit misleading

validator/validators/basic/basic_validator.go
```diff
@@ -37,7 +39,7 @@ func NewBasicValidator(vConfig models.ValidateConfig) models.ValidatorFactory {
 
 func (s *BasicValidator) GenerateLoad() error {
 	var (
-		metricSendingInterval = time.Minute
+		metricSendingInterval = 10 * time.Second
```
okankoAMZ, line 42: this doesn't break other integration tests right?

validator/validators/basic/basic_validator.go
```diff
@@ -80,12 +82,63 @@ func (s *BasicValidator) CheckData(startTime, endTime time.Time) error {
...
+				Name:  aws.String("Operation"),
+				Value: aws.String(operationValue),
+			},
+			{
+				Name:  aws.String("Service"),
+				Value: aws.String(serviceValue),
+			},
+		}
+
+		//App Signals metric testing (This is because we want to use a different checking method (same that was done for linux test))
```
okankoAMZ, line 104: nit: might want to wrap it around a function instead of a comment for clean code practices. However, this is also fine

okankoAMZ (commented):
Left some comments

okankoAMZ:
Can you also `git-rebase` to squash the commits before you merge

util/common/metrics.go
Paramadon: Since we are running for a minute and it sleeps every 5 seconds we will generate 12 times. I'll change it in terms of words to make easier to understand

util/common/metrics.go
Paramadon: This is the same method we used previously for Linux test, so just keeping it the same as before

util/awsservice/constant.go
Paramadon: To assign region to awsConfig otherwise getMetricData method fail due to awsConfig not having a region.

util/awsservice/constant.go
Paramadon: To be able to intialize awsConfig with the right region

validator/validators/basic/basic_validator.go
```diff
@@ -80,12 +82,63 @@ func (s *BasicValidator) CheckData(startTime, endTime time.Time) error {
 				Value: aws.String(dimension.Value),
 			})
 		}
-		err := s.ValidateMetric(metric.MetricName, metricNamespace, metricDimensions, metric.MetricValue, metric.MetricSampleCount, startTime, endTime)
+		environmentValue := "Generic"
+		operationValue := "operation"
+		serviceValue := "service-name"
+
+		appSignalDimensions := []cwtypes.Dimension{
```
movence, line 89: these dimensions are already defined in `parameters.yaml`, so you can just use the existing logic to build a dim set? Just omit `InstnanceId` when metric name is for appsignals.

validator/main.go
```diff
@@ -79,6 +80,7 @@ func main() {
 func validate(vConfig models.ValidateConfig) error {
 	var err error
 	for i := 0; i < awsservice.StandardRetries; i++ {
+		basic.RetryCount = i + 1
```
movence, line 83: where is this var being used? and for what?

Paramadon: Oh nice catch! That will be removed I was using it for testing purposes

movence (approved):

musa-asad (approved):

util/common/metrics.go
okankoAMZ: I still insist by this

okankoAMZ (approved):
Nothing blocking, but there is still changes I would like to see

PR #408, opened 2024-05-13
Fix panic when metrics not yet ready in a cluster

movence:
# Description of the issue
Container Insights integration test fails with a panic when a cluster is not ready with target metrics

# Description of changes
- Add a precheck before calling `rand.Intn`

# License


Paramadon (approved):
makes sense! Preventing to panic during failure bc of the randIdx

Paramadon (approved):

lisguo (approved):

PR #410, opened 2024-05-24
Fix fluent bit terraform tests

KlwntSingh:
1. Fix timeout issue in fluent bit tests when destroying cluster
2. Fix CW template yaml with workingDir field

# Description of the issue
Fluentbit tests on Windows times out when destroying EKS cluster.  This fix in fluentbit fix that issue. Problem was with order of resources deletion by terraform.

# Description of changes
Used same terraform template as for CW agent tests on Windows.

# License

# Tests
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/9220016035


terraform/eks/daemon/default_resources/cwagent-windows.yaml
mitali-salvi, line 1: nitpick - should keep the naming of the yaml uniform. cwagent-windows```-daemonset``` and fluentbit-windows-configmap

terraform/eks/daemon/fluent/windows/main.tf
```diff
@@ -106,33 +157,365 @@ resource "aws_eks_node_group" "node_group_windows" {
...
+    template {
+      metadata {
+        labels = {
+          "name" : "cloudwatch-agent"
+        }
+      }
+      spec {
+        container {
+          name  = "cloudwatch-agent"
+          image = "public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.247359.1b252618"
```
mitali-salvi, line 370: why is an older version of the agent being used ?

terraform/eks/daemon/default_resources/cwagent-windows.yaml
KlwntSingh: Addressed

terraform/eks/daemon/fluent/windows/main.tf
KlwntSingh: This logic is similar to fluent bit test on Linux.
Since this is fluent bit test, CW agent version shouldn't matter.

mitali-salvi (approved):

terraform/eks/daemon/fluent/windows/main.tf
movence: agree that agent version doesn't really matter in the context of fluentbit/d tests, but it would be nice to update it while we are at it. I think it's also a good practice to not keep a legacy agent version in Github. 

Can you please update it to `1.300040.0b650`? The other file that has this old version is `terraform/eks/daemon/fluent/common/main.tf`.

KlwntSingh: Addressed.
Update Cw agent version to `1.300040.0b650` at both the places.

movence (approved):

terraform/eks/daemon/fluent/common/main.tf
```diff
@@ -280,7 +280,7 @@ resource "kubernetes_daemonset" "agent_daemon" {
       spec {
         container {
           name  = "cloudwatch-agent"
-          image = "public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.247359.1b252618"
+          image = "public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300040.0b650"
```
okankoAMZ, line 283: Do we want to hardcode this? don't we always want to test latest?

okankoAMZ: Sounds good! I wanted to make sure it is documented.

okankoAMZ (approved):

PR #416, opened 2024-08-09
add integration tests for AWS Neuron

aditya-purang:
# Description of the issue
Adding integration tests for AWS Neuron Metrics

# Description of changes
- Added infrastructure code for AWS Neuron metrics cluster
- Added tests for verification of log schema, cluster dimension metrics and EMF log frequency per minute for AWS neuron metrics

# License

# Tests
Tested with a `ci-neuron-integ-test` branch in agent repo

test link : https://github.com/aws/amazon-cloudwatch-agent/actions/runs/10328302776/job/28594690160


terraform/eks/daemon/awsneuron/main.tf
```diff
@@ -0,0 +1,826 @@
...
+                  operator = "In"
+                  values   = ["linux"]
+                }
+              }
+            }
+          }
+        }
+        container {
+          name  = "neuron-monitor-prometheus"
+          image = "506463145083.dkr.ecr.us-west-2.amazonaws.com/mocked-neuron-monitor:v2"
```
movence, line 398: does this need to be updated everytime there is a new version of neuron monitor? what is the process of creating a mocked image?

terraform/eks/daemon/awsneuron/main.tf
```diff
@@ -0,0 +1,826 @@
...
+resource "null_resource" "validator" {
+  depends_on = [
+    aws_eks_node_group.this,
+    kubernetes_daemonset.service,
+    kubernetes_cluster_role_binding.rolebinding,
+    kubernetes_service_account.cwagentservice,
+  ]
+  provisioner "local-exec" {
+    command = <<-EOT
+      echo "Validating EKS metrics/logs for EMF"
```
movence, line 821: need this updated?

terraform/eks/daemon/awsneuron/main.tf
```diff
@@ -0,0 +1,826 @@
...
+      cd ../../../..
+      go test -timeout 30m ${var.test_dir} -eksClusterName=${aws_eks_cluster.this.name} -computeType=EKS -v -eksDeploymentStrategy=DAEMON -eksGpuType=nvidia
```
movence, line 823: it might not be relevant to this specific test but we can probably drop this gpu type if is not needed for other types.

terraform/eks/daemon/awsneuron/variables.tf
```diff
@@ -0,0 +1,47 @@
...
+  type    = string
+  default = "1.28"
+}
+
+variable "ami_type" {
+  type    = string
+  default = "AL2_x86_64"
+}
+
+variable "instance_type" {
```
movence, line 44: remove since this is not being used?

test/awsneuron/neuron_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build !windows
+
+package emf
```
movence, line 6: change this?

terraform/eks/daemon/awsneuron/main.tf
aditya-purang: updated

terraform/eks/daemon/awsneuron/variables.tf
aditya-purang: cleaned up the variables

test/awsneuron/neuron_test.go
aditya-purang: updated

terraform/eks/daemon/awsneuron/main.tf
aditya-purang: should be possible but we wanted to keep it as close as possible to the real implementation and deployment of the neuron monitor pod

movence: feels like it's another thing to manage outside of test itself when there is already a script doing exactly the same thing. should we expect this test to still pass when we use an actual image?

aditya-purang: it would, the only problem is it wont emit some of the metrics which can be sparse like errors and error correction events so we might miss some regression there. This way we inject those metrics in the prom script and then assert that all the metrics are correct

terraform/eks/daemon/awsneuron/main.tf
aditya-purang: removed

terraform/eks/daemon/awsneuron/main.tf
movence: > should be possible but we wanted to keep it as close as possible to the real implementation and deployment of the neuron monitor pod

How does it make close to the real implementation when the extra layer of docker image is just a wrapper of the python script underneath? I still think just running the same python script directly on a container or using the real neuron image would be better from the management perspective. With the real neuron image, is that sparse metric issue a known issue with no easy workaround or fix for it?

aditya-purang: Yeah, it was a conscious decision we took to not emit continuous 0s for metrics which  rarely occur, another case is that we want to simulate multiple runtimes running on the same host which we cannot test unless we actually run multiple runtimes (neuron burns) or mock their output.

movence (approved):

sam6134 (approved):

PR #427, opened 2024-11-12
Add EKS integration test for FluentBit log emission with Entity field

zhihonl:
# Description of the issue
FluentBit has a new feature to emit CloudWatch logs with entity field which helps users navigate between their telemetry in the CloudWatch console. We want to add an integration test which helps us validate this behavior.

This PR focuses on the test scenario where FluentBit falls back to using K8sWorkload as the service name and uses default environment name. 

# Description of changes
- Add new terraform that uses helm chart to set up the cluster
- Create a worker pod that output log every second
- Add test code to validate the entity fields contain the necessary values.

# License

# Tests
Test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/11790939699/job/32939527206 (Ran after commit [861e08b](https://github.com/aws/amazon-cloudwatch-agent-test/pull/427/commits/861e08bfff6c58eaeb71df2b72459ae1c4e0588a))

### DIFF
--- a/terraform/eks/daemon/entity/main.tf
+++ b/terraform/eks/daemon/entity/main.tf
@@ -19,13 +19,6 @@ resource "aws_eks_cluster" "this" {
   name     = "cwagent-eks-integ-${module.common.testing_id}"
   role_arn = module.basic_components.role_arn
   version  = var.k8s_version
-  enabled_cluster_log_types = [
-    "api",
-    "audit",
-    "authenticator",
-    "controllerManager",
-    "scheduler"
-  ]
   vpc_config {
     subnet_ids         = module.basic_components.public_subnet_ids
     security_group_ids = [module.basic_components.security_group]
--- a/terraform/eks/daemon/entity/main.tf
+++ b/terraform/eks/daemon/entity/main.tf
@@ -182,7 +182,6 @@ resource "helm_release" "aws_observability" {
   namespace  = "amazon-cloudwatch"
   create_namespace = true
 
-  # Specify any necessary values here
   set {
     name  = "clusterName"
     value = aws_eks_cluster.this.name
@@ -251,7 +250,7 @@ resource "null_resource" "validator" {
 
   provisioner "local-exec" {
     command = <<-EOT
-      echo "Validating EKS metrics/traces for AppSignals"
+      echo "Validating EKS logs for entity fields"
       cd ../../../..
       go test ${var.test_dir} -timeout 1h -eksClusterName=${aws_eks_cluster.this.name} -computeType=EKS -v -eksDeploymentStrategy=DAEMON -instanceId=${data.aws_instance.eks_node_detail.instance_id}
     EOT
--- a/terraform/eks/daemon/entity/main.tf
+++ b/terraform/eks/daemon/entity/main.tf
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 module "common" {
-  source             = "../../../common"
+  source = "../../../common"
 }
 
 module "basic_components" {
@@ -177,9 +177,9 @@ resource "null_resource" "clone_helm_chart" {
 }
 
 resource "helm_release" "aws_observability" {
-  name       = "amazon-cloudwatch-observability"
-  chart      = "./helm-charts/charts/amazon-cloudwatch-observability"
-  namespace  = "amazon-cloudwatch"
+  name             = "amazon-cloudwatch-observability"
+  chart            = "./helm-charts/charts/amazon-cloudwatch-observability"
+  namespace        = "amazon-cloudwatch"
   create_namespace = true
 
   set {
@@ -194,7 +194,7 @@ resource "helm_release" "aws_observability" {
   depends_on = [
     aws_eks_cluster.this,
     aws_eks_node_group.this,
-    null_resource.clone_helm_chart]
+  null_resource.clone_helm_chart]
 }
 
 resource "kubernetes_pod" "log_generator" {
--- a/terraform/eks/daemon/entity/main.tf
+++ b/terraform/eks/daemon/entity/main.tf
@@ -1,0 +1,260 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+module "common" {
+  source             = "../../../common"
+}
+
+module "basic_components" {
+  source = "../../../basic_components"
+
+  region = var.region
+}
+
+data "aws_eks_cluster_auth" "this" {
+  name = aws_eks_cluster.this.name
+}
+
+resource "aws_eks_cluster" "this" {
+  name     = "cwagent-eks-integ-${module.common.testing_id}"
+  role_arn = module.basic_components.role_arn
+  version  = var.k8s_version
+  enabled_cluster_log_types = [
+    "api",
+    "audit",
+    "authenticator",
+    "controllerManager",
+    "scheduler"
+  ]
+  vpc_config {
+    subnet_ids         = module.basic_components.public_subnet_ids
+    security_group_ids = [module.basic_components.security_group]
+  }
+}
+
+# EKS Node Groups
+resource "aws_eks_node_group" "this" {
+  cluster_name    = aws_eks_cluster.this.name
+  node_group_name = "cwagent-eks-integ-node-${module.common.testing_id}"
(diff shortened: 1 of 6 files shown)


varunch77 (dismissed):

terraform/eks/daemon/entity/main.tf
```diff
@@ -0,0 +1,259 @@
...
+data "aws_eks_cluster_auth" "this" {
+  name = aws_eks_cluster.this.name
+}
+
+resource "aws_eks_cluster" "this" {
+  name     = "cwagent-eks-integ-${module.common.testing_id}"
+  role_arn = module.basic_components.role_arn
+  version  = var.k8s_version
+  enabled_cluster_log_types = [
+    "api",
```
lisguo, line 23: do we need these? Not sure what these do

terraform/eks/daemon/entity/main.tf
```diff
@@ -0,0 +1,259 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+module "common" {
```
lisguo, line 4: is there no common terraform to create an eks cluster? seems like we have duplicate code here from our EKS integ tests...can we just reuse it and then apply the helm chart?

test/entity/entity_test.go
```diff
@@ -0,0 +1,219 @@
...
+	entityPlatform          = "@entity.Attributes.PlatformType"
+	entityInstanceId        = "@entity.Attributes.EC2.InstanceId"
+	entityEKSCluster        = "@entity.Attributes.EKS.Cluster"
+	entityK8sNode           = "@entity.Attributes.K8s.Node"
+	entityK8sNamespace      = "@entity.Attributes.K8s.Namespace"
+	entityK8sWorkload       = "@entity.Attributes.K8s.Workload"
+	entityServiceNameSource = "@entity.Attributes.AWS.ServiceNameSource"
+)
+
+type expectedEntity struct {
```
lisguo, line 37: can't this just be the `cloudwatch.Entity` struct? Then we can have checks to ensure the KeyAttributes/Attributes are valid in the struct?

test/entity/entity_test.go
```diff
@@ -0,0 +1,219 @@
...
+			// check CWL to ensure we got the expected entities in the log group
+			queryString := fmt.Sprintf("fields @message, @entity.KeyAttributes.Type, @entity.KeyAttributes.Name, @entity.KeyAttributes.Environment, @entity.Attributes.PlatformType, @entity.Attributes.EKS.Cluster, @entity.Attributes.K8s.Node, @entity.Attributes.K8s.Namespace, @entity.Attributes.K8s.Workload, @entity.Attributes.AWS.ServiceNameSource, @entity.Attributes.EC2.InstanceId | filter @logStream == \"%s\"", podApplicationLogStream)
+			ValidateEntity(t, appLogGroup, podApplicationLogStream, &end, queryString, testCase.expectedEntity, string(env.ComputeType))
+		})
+	}
+}
+
+// ValidateEntity queries a given LogGroup/LogStream combination given the start and end times, and executes an
+// log query for entity attributes to ensure the entity values are correct
+func ValidateEntity(t *testing.T, logGroup, logStream string, end *time.Time, queryString string, expectedEntity expectedEntity, entityPlatformType string) {
```
lisguo, line 122:
```suggestion
func ValidateLogEntity(t *testing.T, logGroup, logStream string, end *time.Time, queryString string, expectedEntity expectedEntity, entityPlatformType string) {
```

test/entity/entity_test.go
```diff
@@ -0,0 +1,219 @@
...
+	var requiredEntityFields map[string]bool
+
+	log.Printf("Checking log group/stream: %s/%s", logGroup, logStream)
+
+	if !awsservice.IsLogGroupExists(logGroup) {
+		t.Fatalf("application log group used for entity validation doesn't exsit: %s", logGroup)
+	}
+	begin := end.Add(-sleepForFlush * 2)
+	log.Printf("Start time is " + begin.String() + " and end time is " + end.String())
```
lisguo, line 131: Let's make sure these are accurate...how are we tracking the begin time? It should be when the terraform is initiated?

test/entity/entity_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	result, err := awsservice.GetLogQueryResults(logGroup, begin.Unix(), end.Unix(), queryString)
+	assert.NoError(t, err)
+	if !assert.NotZero(t, len(result)) {
+		return
+	}
+	if entityPlatformType == "EC2" {
```
lisguo, line 138: Feel like we should have a defined constant for required KeyAttributes and Attributes for each of the platforms. 

Something like

```
ec2RequiredKeyAttributes := map[string]struct{}{} {
    "Service": {},
    "Environment":{},
    ...
}
```

And have some validator in a util func. Having a map[string]struct{}{} is just easier to do the checks...not sure if we have a List abstraction for this.

util/awsservice/cloudwatchlogs.go
```diff
@@ -197,6 +197,47 @@ func GetLogQueryStats(logGroupName string, startTime, endTime int64, queryString
 	}
 }
 
+// GetLogQueryResults for the log group between start/end (in epoch seconds) for the
+// query string.
+func GetLogQueryResults(logGroupName string, startTime, endTime int64, queryString string) ([][]types.ResultField, error) {
```
lisguo, line 202: we don't have something like this already? Something that checks for the existence of logs in a log group within a certain time period?

terraform/eks/daemon/entity/main.tf
zhihonl: There is no common terraform to create EKS cluster. All terraform setups for each test under `terraform/eks` folder have their own cluster setup. I think your point is valid though, having a common code to avoid potential duplication is better but I think it's out of scope for this PR.

test/entity/entity_test.go
zhihonl: When querying the logs, the response doesn't return in `cloudwatch.Entity` type, it's `Results [][]*ResultField` which is basically arrays of strings. We could try to use `cloudwatch.Entity` but it would require additional logics to deference the maps just to turn them into strings so thought using a custom test struct is cleaner.

util/awsservice/cloudwatchlogs.go
zhihonl: We have [GetLogQueryStats](https://github.com/aws/amazon-cloudwatch-agent-test/blob/eks-entity/util/awsservice/cloudwatchlogs.go#L161) function which does what you described but that returns the statistics of the query result which does not include log fields so we can't validate entity from that. `GetLogQueryResults` function is a copy of that function except that it returns `results.Results`

test/entity/entity_test.go
zhihonl: Will fix in next commit

terraform/eks/daemon/entity/main.tf
zhihonl: Tried running the setup without these and still works. I think these provide control plane logs from EKS under `/aws/eks` log group but we only care about application logs from Container Insights log group for these tests.

test/entity/entity_test.go
zhihonl: Going to create a validator struct that performs different logics based on platforms.

test/entity/entity_test.go
lisguo: but this assumes that the cluster was created within 6 minutes before the end time right? Wondering if there is a way to start the timer at terraform init. Not sure how we are doing it for our other tests

test/entity/entity_test.go
```diff
@@ -0,0 +1,222 @@
...
+		agentConfigPath string
+		podName         string
+		useEC2Tag       bool
+		expectedEntity  expectedEntity
+	}{
+		"Entity/K8sWorkloadServiceNameSource": {
+			agentConfigPath: filepath.Join("resources", "compass_default_log.json"),
+			podName:         "log-generator",
+			expectedEntity: expectedEntity{
+				entityType:        "Service",
```
lisguo, line 163: nit: we could have an enum for different entity types: Service, Resource

Same for platformType: AWS::EKS, AWS::K8s, etc

lisguo (approved):

varunch77 (approved):

PR #430, opened 2024-11-14
Add E2E EKS Terraform environment.

musa-asad:
This PR was moved to #434 in order to be able to call it as a feature branch from the GitHub workflow from the `amzon-cloudwatch-agent` repo.


terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,164 @@
...
+      cd helm-charts
+      git checkout ${var.helm_charts_branch}
+    EOT
+  }
+
+  provisioner "local-exec" {
+    when    = destroy
+    command = "rm -rf ${path.module}/helm-charts"
+  }
+}
```
musa-asad, line 122: Just here for testing.

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,128 @@
...
+
+resource "null_resource" "kubectl" {
+  depends_on = [
+    aws_eks_cluster.this,
+    aws_eks_node_group.this
+  ]
+  provisioner "local-exec" {
+    command = <<-EOT
+      ${local.aws_eks} update-kubeconfig --name ${aws_eks_cluster.this.name}
```
Paramadon, line 102: If the default cluster name is being used should we not add an id to make sure it is a different cluster name than a run that is concurrently running?

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,128 @@
...
+  source = "../../common"
+}
+
+module "basic_components" {
+  source = "../../basic_components"
+}
+
+locals {
+  aws_eks      = "aws eks --region ${var.region}"
+  cluster_name = var.cluster_name != "" ? var.cluster_name : "cwagent-monitoring-config-e2e-eks"
```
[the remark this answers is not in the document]
Paramadon: something like a testing id to have uniqueness to the  cluster name.

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,128 @@
...
+module "common" {
+  source = "../../common"
+}
+
+module "basic_components" {
+  source = "../../basic_components"
+}
+
+locals {
+  aws_eks      = "aws eks --region ${var.region}"
```
Paramadon, line 13: --region is redundant since we already have --region in provider.tf
```suggestion
  aws_eks      = "aws eks"
```

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,128 @@
...
+  provisioner "local-exec" {
+    command = <<-EOT
+      git clone https://github.com/aws-observability/helm-charts.git helm-charts
+      cd helm-charts
+      git checkout ${var.helm_charts_branch}
+    EOT
+  }
+
+  provisioner "local-exec" {
+    when    = destroy
```
Paramadon, line 119: Why are we only cleaning up during destroy what if there was a retry/failure? We should still cleanup

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,128 @@
...
+    command = "rm -rf ${path.module}/helm-charts"
+  }
+}
```
Paramadon, line 122: Adding a cleanup at the start might be better
```suggestion
resource "null_resource" "helm_charts" {
  provisioner "local-exec" {
    command = <<-EOT
      # Clean up any existing helm-charts directory before cloning
      rm -rf ${path.module}/helm-charts
      # Clone the Helm charts repository and checkout the specific branch
      git clone https://github.com/aws-observability/helm-charts.git helm-charts
      cd helm-charts
      git checkout ${var.helm_charts_branch}
    EOT
  }

  provisioner "local-exec" {
    when    = destroy
    command = "rm -rf ${path.module}/helm-charts"
  }
}
```

terraform/e2e/eks/main.tf
musa-asad: Yeah, we add an id when we create the cluster.

terraform/e2e/eks/main.tf
musa-asad: That's the flag to pass the region in.

musa-asad: Will remove.

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,128 @@
...
+
+resource "null_resource" "kubectl" {
+  depends_on = [
+    aws_eks_cluster.this,
+    aws_eks_node_group.this
+  ]
+  provisioner "local-exec" {
+    command = <<-EOT
+      ${local.aws_eks} update-kubeconfig --name ${aws_eks_cluster.this.name}
+      ${local.aws_eks} list-clusters --output text
```
movence, line 103: I understand other tests also do this but do we really need it?

terraform/e2e/eks/variables.tf
```diff
@@ -0,0 +1,92 @@
...
+  type    = string
+  default = "public.ecr.aws/cloudwatch-agent"
+}
+
+variable "validate_test" {
+  type    = string
+  default = "../../../test/e2e/jmx/validate_test.go"
+}
+
+variable "sample-app" {
```
movence, line 84: where are we using all these new variable including ECR repos and sample apps?

terraform/e2e/eks/providers.tf
```diff
@@ -0,0 +1,25 @@
...
+terraform {
+  required_providers {
+    docker = {
+      source  = "kreuzwerker/docker"
+      version = "3.0.2"
+    }
+  }
+}
+
+provider "docker" {
```
movence, line 13: what is docker for?

movence:
looks like you need to rebase

terraform/e2e/eks/variables.tf
musa-asad: We use them for the EKS cluster that is created. The sample app will be deployed in the cluster and instrumented with ADOT SDK libraries (e.g. java). Then the ECR images will be added to the agent custom resource and operator deployment.

terraform/e2e/eks/providers.tf
musa-asad: Will remove, that was from testing ECR images being uploaded.

terraform/e2e/eks/main.tf
movence: we just need `update-kubeconfig` for newly created cluster. we don't need to run `list-clusters` and `describe-cluster` but maybe just one of them to validate kubeconfig has been udpated correctly

musa-asad: Ok, I fixed my PR to include that.

terraform/e2e/eks/variables.tf
musa-asad: Fixed PR.

terraform/e2e/eks/main.tf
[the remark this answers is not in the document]
okankoAMZ: I agree with @Paramadon  here, why is this fully hard coded?

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,155 @@
...
+      ${local.aws_eks} describe-cluster --name ${aws_eks_cluster.this.name} --output text
+    EOT
+  }
+}
+
+resource "null_resource" "install_helm_release_and_sample_app" {
```
okankoAMZ, line 109: I though we were going to do helm install during testing portion of the code and terraform was going to handle only resource generation?

terraform/e2e/eks/main.tf
```diff
@@ -0,0 +1,155 @@
...
+        --set agent.image.tag=${var.cloudwatch_agent_tag} \
+        --set agent.image.repositoryDomainMap.public=${var.cloudwatch_agent_repository_url} \
+        --set manager.image.repository=${var.cloudwatch_agent_operator_repository} \
+        --set manager.image.tag=${var.cloudwatch_agent_operator_tag} \
+        --set manager.image.repositoryDomainMap.public=${var.cloudwatch_agent_operator_repository_url} \
+        --namespace amazon-cloudwatch \
+        --create-namespace \
+        ${var.agent-config != "" ? "--set-json agent.config='${jsonencode(jsondecode(file(var.agent-config)))}'" : ""} \
+        ${var.otel-config != "" ? "--set-file agent.otelConfig=${yamlencode(yamldecode(file(var.otel-config)))}" : ""} \
+        ${var.prometheus-config != "" ? "--set-file agent.prometheus.config=${yamlencode(yamldecode(file(var.prometheus-config)))}" : ""}
```
okankoAMZ, line 135: I think all these should be pulled from values.yaml located in test directory.  When we do it manually like this we get less control over tests and it decreases our agility in updating our test framework.

terraform/e2e/eks/variables.tf
okankoAMZ: Why are we deploying a default sample app for a testing framework? I thought this code is only suppose to generate the resources for testing

okankoAMZ (commented):
The PR aims to add an E2E EKS test environment, which is a valuable addition. However, it includes some changes that may not be directly related to this goal, such as helm install commands and sample applications.

Could you clarify the reasoning behind these additions or consider separating them into a different PR if they're not essential for the E2E EKS environment?

terraform/e2e/eks/main.tf
musa-asad: I decided on doing the initial helm install with Terraform and then do upgrades with testing.

okankoAMZ: Could you explain the reasoning behind it?

terraform/e2e/eks/main.tf
musa-asad: That's fair, but there's an issue with that approach in case we update `values.yaml` in the future. Setting it like this ensures to target values we will always maintain.

terraform/e2e/eks/main.tf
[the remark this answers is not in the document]
musa-asad: It isn't. https://github.com/aws/amazon-cloudwatch-agent-test/pull/430/files#diff-4542098ea7039d57648d4ed98ad0d0d033a5eef29ffbabe22617c02924466c92R22

I just kept the variable as is, but the actual cluster name does include the unique id.

terraform/e2e/eks/variables.tf
okankoAMZ, line 84: sample app is still here

okankoAMZ (approved):
Looks good!

PR #433, opened 2024-11-19
Add logs agent integ test

varunch77:
# Description of the issue
We are currently missing tests that ensure that the logs agent is properly sending an entity when sending application logs to CloudWatch Logs.

# Description of changes
`TestWriteLogsWithEntityInfo()` writes logs to CloudWatch and validates that the logs are associated with entities and that all values are translated properly. The three test cases are `IAMRole`, `ServiceInConfig` (which uses the added `config_log_service_name.json` configuration), and `EC2Tags`. The changes also include a suite of helper functions like `ValidateEntity()` to validate the logs and check that the expected entity information is present after querying the published logs. 

# License

# Tests
I ran all tests in the `publish_logs_test.go` file on a newly created EC2 instance with the most recent build of the cloudwatch agent and the end-to-end testing IAM role attached. I also verified that `TestWriteLogsWithEntityInfo` passed for every running order combination of the IAM, EC2, and Service test cases.

Test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12283029637


test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -119,9 +161,9 @@ func TestWriteLogsToCloudWatch(t *testing.T) {
 
 			// ensure that there is enough time from the "start" time and the first log line,
 			// so we don't miss it in the GetLogEvents call
-			time.Sleep(sleepForFlush)
+			time.Sleep(sleepForExtendedFlush)
```
lisguo, line 164: Why not just increase `sleepForFlush`? Do we use `sleepForFlush` anywhere else in the code?

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -139,6 +181,120 @@ func TestWriteLogsToCloudWatch(t *testing.T) {
...
+			common.StartAgent(configOutputPath, true, false)
+			time.Sleep(sleepForExtendedFlush)
+			writeLogLines(t, f, testCase.iterations)
+			time.Sleep(sleepForExtendedFlush)
+			common.StopAgent()
+			end := time.Now()
+
+			ValidateEntity(t, instanceId, instanceId, &end, testCase.expectedEntity)
+
+			f.Close()
```
lisguo, line 292: feel like we can `f.Close()` after we are done writing on line 285

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +488,152 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
 	)
 	assert.NoError(t, err)
 }
+
+func ValidateEntity(t *testing.T, logGroup, logStream string, end *time.Time, expectedEntity expectedEntity) {
+	log.Printf("Validating entity for log group: %s, stream: %s", logGroup, logStream)
+
+	logGroupInfo, err := getLogGroup()
+	for _, lg := range logGroupInfo {
```
lisguo, line 496: Is there no way to get a specific log group when you provide the name? 

Seems expensive to get all log groups in an account. We have a ton in our test account..

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +488,152 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
...
+		if *lg.LogGroupName == logGroup {
+			log.Println("Log group " + *lg.LogGroupName + " exists")
+			break
+		}
+	}
+	assert.NoError(t, err)
+	begin := end.Add(-sleepForExtendedFlush * 4)
```
lisguo, line 503: Instead of trying to calculate the begin time, can we just pass in the begin time?

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +488,152 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
...
+			allEntityFieldsFound = false
+		}
+	}
+	assert.True(t, allEntityFieldsFound)
+}
+
+func getLogQueryId(logGroup string, since, until *time.Time) (*string, error) {
+	var queryId *string
+	params := &cloudwatchlogs.StartQueryInput{
+		QueryString:  aws.String("fields @message, @entity.KeyAttributes.Type, @entity.KeyAttributes.Name, @entity.KeyAttributes.Environment, @entity.Attributes.PlatformType, @entity.Attributes.EC2.InstanceId"),
```
lisguo, line 553: nit: make this a const

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +488,152 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
...
+	log.Printf("Query start time is " + begin.String() + " and end time is " + end.String())
+	queryId, err := getLogQueryId(logGroup, &begin, end)
+	assert.NoError(t, err)
+	log.Printf("queryId is " + *queryId)
+	result, err := getQueryResult(queryId)
+	assert.NoError(t, err)
+	if !assert.NotZero(t, len(result)) {
+		return
+	}
+	requiredEntityFields := map[string]bool{
```
lisguo, line 513: Not sure I understand why we need this...when we pass in the query:

```
fields @message, @entity.KeyAttributes.Type, @entity.KeyAttributes.Name, @entity.KeyAttributes.Environment, @entity.Attributes.PlatformType, @entity.Attributes.EC2.InstanceId
```

Doesn't the fact that there are results mean that the logs all have the required fields?

test/cloudwatchlogs/publish_logs_test.go
varunch77: addressed in commit [84cc3d3](https://github.com/aws/amazon-cloudwatch-agent-test/pull/433/commits/84cc3d39def9c67c5a0460105bd990fe712f031e#diff-90e695612350cf075a13140798e4d267b81f04e5d9258b705f170ab76499d804R53)

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -86,8 +107,29 @@ type cloudWatchLogGroupClassTestInput struct {
...
+
 func init() {
 	environment.RegisterEnvironmentMetaDataFlags()
+	awsCfg, err := config.LoadDefaultConfig(
+		context.Background(),
+		config.WithRegion(pdxRegionalCode),
+	)
+	if err != nil {
+		fmt.Println("There was an error trying to load default config: ", err)
+		return
```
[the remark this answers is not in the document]
varunch77: It didn't, but I changed the behavior so that it will fail now if the default config fails in commit [84cc3d3](https://github.com/aws/amazon-cloudwatch-agent-test/pull/433/commits/84cc3d39def9c67c5a0460105bd990fe712f031e)

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -29,9 +37,19 @@ const (
 	logLineId2                    = "bar"
 	logFilePath                   = "/tmp/cwagent_log_test.log" // TODO: not sure how well this will work on Windows
 	sleepForFlush                 = 20 * time.Second            // default flush interval is 5 seconds
+	sleepForExtendedFlush         = 180 * time.Second           // increase flush time for the two main tests
+	retryWaitTime                 = 30 * time.Second
 	configPathAutoRemoval         = "resources/config_auto_removal.json"
 	standardLogGroupClass         = "STANDARD"
 	infrequentAccessLogGroupClass = "INFREQUENT_ACCESS"
+	cwlPerfEndpoint               = "https://logs.us-west-2.amazonaws.com"
```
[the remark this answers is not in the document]
varunch77: I think we may already be constrained by the region being hardcoded elsewhere: https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/util/awsservice/constant.go#L58C1-L60C78. We would need to make edits there but I'm a little wary of what the blast radius of that would look like for our testing

test/cloudwatchlogs/publish_logs_test.go
varunch77: Updated this for `ValidateEntity()` (commit [7950725](https://github.com/aws/amazon-cloudwatch-agent-test/pull/433/commits/795072522768ef075f3068f93a5c84642193acc3))

test/cloudwatchlogs/publish_logs_test.go
varunch77: I don't think there's a way to just search for one log group (why not is beyond me..) but I updated this so it passes in the log group as a prefix to search by (commit [006a057](https://github.com/aws/amazon-cloudwatch-agent-test/pull/433/commits/006a057bf450c4f0f243278f50bde941a01dd2d9))

varunch77:
New run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/11978587627

test/cloudwatchlogs/publish_logs_test.go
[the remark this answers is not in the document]
zhihonl: We don't need to hardcode the endpoint here. The code I wrote originally had to use a custom endpoint because it was a beta endpoint but this is the public us-west-2 endpoint which the client will automatically resolve

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -29,9 +37,20 @@ const (
...
+	pdxRegionalCode               = "us-west-2"
```
zhihonl, line 46: region hardcoding is not needed since entity feature is GA

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -86,8 +108,28 @@ type cloudWatchLogGroupClassTestInput struct {
...
 	environment.RegisterEnvironmentMetaDataFlags()
+	awsCfg, err := config.LoadDefaultConfig(
+		context.Background(),
+		config.WithRegion(pdxRegionalCode),
+	)
+	if err != nil {
+		log.Fatalf("Failed to load default config: %v", err)
+	}
+
+	cwlClient = cloudwatchlogs.NewFromConfig(awsCfg, func(o *cloudwatchlogs.Options) {
```
zhihonl, line 129: we don't need to create a new client. we have globally available client that already uses us-west-2 as the region" https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/util/awsservice/constant.go#L74

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -86,8 +108,28 @@ type cloudWatchLogGroupClassTestInput struct {
...
+		o.BaseEndpoint = aws.String(cwlPerfEndpoint)
+	})
+	ec2Client = ec2.NewFromConfig(awsCfg)
```
zhihonl, line 132: same as above

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +497,154 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
 	)
 	assert.NoError(t, err)
 }
+
+func ValidateEntity(t *testing.T, logGroup, logStream string, begin, end *time.Time, expectedEntity expectedEntity) {
+	log.Printf("Validating entity for log group: %s, stream: %s", logGroup, logStream)
+
+	logGroupInfo, err := getLogGroup(logGroup)
```
zhihonl, line 504: You can just use `IsLogGroupExists` from our common package: https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/util/awsservice/cloudwatchlogs.go#L137

The helper functions that I wrote was because we had to create a new client in us-east-1 for beta purpose but they are no longer needed

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +497,154 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
...
+			break
+		}
+	}
+	assert.NoError(t, err)
+	
+	log.Printf("Query start time is " + begin.String() + " and end time is " + end.String())
+	queryId, err := getLogQueryId(logGroup, begin, end)
+	assert.NoError(t, err)
+	log.Printf("queryId is " + *queryId)
+	result, err := getQueryResult(queryId)
```
zhihonl, line 517: `getLogQueryId` and `getQueryResult` are not needed anymore since we can just use default client now. There's a helper function here: https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/util/awsservice/cloudwatchlogs.go#L202

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -332,3 +497,154 @@ func checkData(t *testing.T, start time.Time, lineCount int) {
 	)
 	assert.NoError(t, err)
 }
+
+func ValidateEntity(t *testing.T, logGroup, logStream string, begin, end *time.Time, expectedEntity expectedEntity) {
```
zhihonl, line 501: nonblocking but recommended: This function has a lot of overlaps with this function: https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/entity/entity_test.go#L241

Maybe good idea to move `ValidateEntity` to a common package and just use for any entity validation tests in the future.

varunch77:
Addressed comments above, including merging `ValidateEntity` with `ValidateLogEntity` and moving the resulting function to the `util` package so it can be used in multiple places for future entity validation tests.

New run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12172724382

PR #434, opened 2024-12-02
Implement EKS E2E Terraform Framework

musa-asad:
# Description of the issue
We aim to enhance our testing framework to support end-to-end testing of various `helm-charts` values and allow the use of custom branches for the agent, operator, and helm-charts repositories to enable testing of different configurations and prevent regressions to verify Kubernetes resource creation/deletion and metric emission.

# Description of changes
**Co-PR:** https://github.com/aws/amazon-cloudwatch-agent/pull/1445.

- Changes brought over from #430 to be able to reference this repository from the GitHub workflow in the `amazon-cloudwatch-agent` package.
- Added new fields to `MetaData` and `MetaDataStrings` structs in `environment/metadata.go` to support EKS E2E testing parameters.
- Implemented `registerEKSE2ETestData` function to register new command-line flags for EKS E2E testing.
- Created new Terraform files (`main.tf`, `providers.tf`, `variables.tf`) in `terraform/eks/e2e` to set up EKS clusters for testing.
- Added `eks_e2e` test configuration in `generator/test_case_generator.go`. Separated it into `testTypeToTestConfigE2E` in order to avoid generating the non-E2E matrix files in the E2E workflow.
- Implemented `ConfigureAWSClients` function in `util/awsservice/constant.go` to allow custom region changes for AWS clients.

**JMX will be implemented in the following PR:** https://github.com/aws/amazon-cloudwatch-agent-test/pull/443.

# License

# Tests
```
terraform apply --auto-approve -var="test_dir=../../../test/e2e/jmx" -var="agent_config=../../../test/e2e/jmx/resources/cwagent_configs/jvm_tomcat.json" -var="sample_app=../../../test/e2e/jmx/resources/sample_apps/tomcat-deployment.yaml"
```

Successfully deploys the CWA Operator, CWA, and Sample Application:
Screenshot 2024-11-15 at 12 29 42 PM

Also utilizes the correct images.
Screenshot 2024-11-15 at 12 30 08 PM
Screenshot 2024-11-15 at 12 30 30 PM


TravisStark (approved):

terraform/eks/e2e/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+resource "null_resource" "validator" {
+  depends_on = [aws_eks_cluster.this, aws_eks_node_group.this, null_resource.helm_charts]
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      echo "Validating K8s resources and metrics"
+      go test -timeout 30m -v ${var.test_dir} \
```
musa-asad, line 115:
```suggestion
      go test -timeout 2h -v ${var.test_dir} \
```

terraform/eks/e2e/providers.tf
```diff
@@ -0,0 +1,15 @@
...
+    docker = {
+      source  = "kreuzwerker/docker"
+      version = "3.0.2"
+    }
+  }
+}
+
+provider "aws" {
+  region = var.region
+}
```
musa-asad, line 15:
```suggestion
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: MIT

provider "aws" {
  region = var.region
}

provider "kubernetes" {
  exec {
    api_version = "client.authentication.k8s.io/v1beta1"
    command     = "aws"
    args        = ["eks", "get-token", "--cluster-name", aws_eks_cluster.this.name]
  }
  host                   = aws_eks_cluster.this.endpoint
  cluster_ca_certificate = base64decode(aws_eks_cluster.this.certificate_authority.0.data)
  token                  = data.aws_eks_cluster_auth.this.token
}
```

generator/resources/eks_e2e_test_matrix.json
```diff
@@ -0,0 +1,5 @@
+[
+  {
+
+  }
+]
```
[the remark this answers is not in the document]
musa-asad: Each empty set of braces in the JSON array represents the tests under:
```
"eks_e2e": {
		{testDir: "../../../test/e2e/jmx"},
	},
```

So it can't be empty.

environment/metadata.go
```diff
@@ -101,6 +133,25 @@ func registerEKSData(d *MetaDataStrings) {
 	flag.StringVar(&(d.EksGpuType), "eksGpuType", "", "nvidia/inferentia")
 }
 
+func registerEKSE2ETestData(dataString *MetaDataStrings) {
+	flag.StringVar(&(dataString.Region), "region", "", "AWS region")
+	flag.StringVar(&(dataString.K8sVersion), "k8s_version", "", "Kubernetes version")
+	flag.StringVar(&(dataString.HelmChartsBranch), "helm_charts_branch", "", "Helm charts branch")
+	flag.StringVar(&(dataString.CloudwatchAgentRepository), "cloudwatch_agent_repository", "", "CloudWatch Agent repository")
+	flag.StringVar(&(dataString.CloudwatchAgentTag), "cloudwatch_agent_tag", "", "CloudWatch Agent tag")
+	flag.StringVar(&(dataString.CloudwatchAgentRepositoryURL), "cloudwatch_agent_repository_url", "", "CloudWatch Agent repository URL")
```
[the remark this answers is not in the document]
musa-asad: It's a good suggestion, but since we automate this with our workflow, it is probably better (and ensures accuracy) to keep it the way it is now.

util/awsservice/constant.go
```diff
@@ -77,3 +79,27 @@ func init() {
 	CloudformationClient = cloudformation.NewFromConfig(awsCfg)
 	XrayClient = xray.NewFromConfig(awsCfg)
 }
+
+// ReconfigureAWSClients reconfigures the AWS clients using a new region.
+func ReconfigureAWSClients(region string) error {
```
dricross, line 84: nit: this is basically the same thing as `init`. any reason to keep the implementation separate? also, it's a bit odd that if you want to use a region that not us-west-2, you have to create all the client for us-west-2 and then re-create them all for the desired region. is there something that prevents us from setting the correct region in `init` to start with? I'm not sure where this function is actually called from, so these may be naive questions

musa-asad: I cleaned up the code a little, but the problem is that I want to retain the initial `init()` functionality due to other tests using this package, but we also should have the option to test in different regions, which my test will have the option for.

I could remove the `init()` function and make sure to call this in the other files that use `awsservice` instead, but that seems prone to making errors and breaking tests due to the amount of places it's used.

dricross: Makes sense to me. Thanks for consolidating the logic

dricross (approved):

terraform/eks/e2e/main.tf
```diff
@@ -34,9 +34,9 @@ resource "aws_eks_node_group" "this" {
   subnet_ids      = module.basic_components.public_subnet_ids
 
   scaling_config {
-    desired_size = 2
-    max_size     = 2
-    min_size     = 2
```
okankoAMZ: So we are _always_ generating a single node group? I think it would be useful if are able to configure it

util/awsservice/constant.go
```diff
@@ -56,12 +58,23 @@ var (
...
+	err := ConfigureAWSClients("us-west-2")
+	if err != nil {
+		fmt.Println("There was an error trying to configure the AWS clients: ", err)
+	}
+}
+
+// ConfigureAWSClients configures the AWS clients using a set region.
+func ConfigureAWSClients(region string) error {
+	mu.Lock()
+	defer mu.Unlock()
```
okankoAMZ, line 71: Why do we need a lock?

terraform/eks/e2e/main.tf
```diff
@@ -0,0 +1,137 @@
...
+resource "aws_eks_node_group" "this" {
+  cluster_name    = aws_eks_cluster.this.name
+  node_group_name = "${local.cluster_name}-node"
+  node_role_arn   = aws_iam_role.node_role.arn
+  subnet_ids      = module.basic_components.public_subnet_ids
+
+  scaling_config {
+    desired_size = 1
+    max_size     = 1
+    min_size     = 1
```
okankoAMZ, line 39: So this mean we will always have only 1 node for EKS tests? It might be useful if we get to configure this

util/awsservice/constant.go
musa-asad: We need a lock of some kind because when we call `ConfigureAWSClients` inidividually, we do after we call the `init` function implicitly through another package. The lock prevents both the `init` and `ConfigureAWSClients` from changing the client values at the same time.

terraform/eks/e2e/main.tf
musa-asad: Yeah, I agree, I'll pass in a variable for this value.

terraform/eks/e2e/main.tf
```diff
@@ -34,9 +34,9 @@ resource "aws_eks_node_group" "this" {
   subnet_ids      = module.basic_components.public_subnet_ids
 
   scaling_config {
-    desired_size = 1
-    max_size     = 1
-    min_size     = 1
+    desired_size = var.nodes
+    max_size     = var.nodes
+    min_size     = var.nodes
```
okankoAMZ, line 39: This is not blocking but shouldn't we be able to configure max and min size separately?

okankoAMZ (approved):
non-blocking comments. Looks good

okankoAMZ (approved):

terraform/eks/e2e/main.tf
musa-asad: We can discuss this, but right now I don't see a use-case for doing anything beyond the desired size.

dricross (approved):

PR #443, opened 2024-12-09
Implement JMX E2E Tests in Terraform Framework

musa-asad:
# Description of the issue
The CloudWatch Agent now supports JMX metrics on EKS. We need to add testing to verify the behavior for JMX metrics collection in EKS environments.

# Description of changes
**Co-PR:** https://github.com/aws/amazon-cloudwatch-agent/pull/1463.

## [Revision 1](https://github.com/aws/amazon-cloudwatch-agent-test/compare/f92b85b567d7fc4f7f31b8070719bc42d401970f..b9ae22bdafb41bc00182e590b1bcebe1b5d8544c)
- Implemented `test/e2e/jmx/jmx_test.go` with the following functions:
  - `ApplyHelm`: Applies a helm release based on the specified `resources` passed in as variables, which sets up the CloudWatch Agent and deploys an annotated sample application.
  - `TestResources`: Tests to see if the appropriate resources have been deployed on the EKS cluster.
  - `TestMetrics`: Tests to see if metrics are being emitted properly based on the agent configuration file passed in.
- Added `test/e2e/jmx/resources/cwagent_configs/` and `test/e2e/jmx/resources/sample_apps/` for custom agent configurations and sample applications to deploy.
  - Added a check to see if `tomcat.sessions` and `tomcat.rejected_sessions` increases above 0 accordingly.
- Added `GetMetricMaximum` to `util/awsservice/cloudwatchmetrics.go` in order to find the maximum value for a metric being emitted.
- Added nodes to `generator/test_case_generator.go` and `generator/resources/eks_e2e_test_matrix.json` to be able to use this as a configurable value from the generated matrix.

## [Revision 2](https://github.com/aws/amazon-cloudwatch-agent-test/compare/b9ae22bdafb41bc00182e590b1bcebe1b5d8544c..8cb81ce05f18b584b05c7317295b6379bcd780f6)
- Use testify's `require` / `assert` methods instead of `t.Fatal` / `t.Error`.
- Add `destroy` logic for Kubernetes resources by uninstalling helm and deleting `test` namespace.
- Removed `--values` from Helm installation since it isn't required.
- Use `kubectl wait` for operator deployment instead of `time.Sleep` to continue as soon as the operator is ready.
- Removed `-deployment` from deployment names due to redundancy.
- Set the namespace once for `testXMetrics` functions due to redundancy.
- Switch to using our own ECR images to keep stable images we control.
- Replaced `GetMetricMaximum` with `CheckMetricAboveZero`, which exits as soon as it finds a metric above zero.
- Added variables for `ami_type` and `instance_type` to have control over those values for future testing.
- Added `kafka` and `kafka_producer` metrics.
- Moved test environment initialization and resource management logic into `util/common/kubernetes.go`.
- Added check for `jmx` in agent configuration.
- Added check for `jmx` environmental variables in application pod.
- Cleaned up code by adding comment blocks and helper functions.

## [Revision 3](https://github.com/aws/amazon-cloudwatch-agent-test/compare/8cb81ce05f18b584b05c7317295b6379bcd780f6..52f3a8639c9f8a1c6090c41971a39dcb08d83ddf)
- Added `STATEFUL` type for future testing.
- Added `var.eks_deployment_strategy` and `DeploymentStrategy` to select the agent mode.
- Changed `Destroy` to boolean in `metadata.go`.
- Increased nodes to 2 to test in multi-node environment.
- Created `testutils.go` function to hold generic test functionality.
- Moved `kubernetes.go` under `test/e2e` and renamed it to `envutils.go`.
- Removed checking for node names due to flakiness.

# License

# Tests
Workflow Status
[View Workflow Run #12541382360](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12541382360)


test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,405 @@
...
+			{"jvm.memory.pool.max", "JVM_TOMCAT_E2E"},
+			{"jvm.memory.pool.used", "JVM_TOMCAT_E2E"},
+			{"jvm.memory.pool.committed", "JVM_TOMCAT_E2E"},
+			{"jvm.threads.count", "JVM_TOMCAT_E2E"},
+			{"tomcat.sessions", "JVM_TOMCAT_E2E"},
+			{"tomcat.errors", "JVM_TOMCAT_E2E"},
+			{"tomcat.request_count", "JVM_TOMCAT_E2E"},
+			{"tomcat.max_time", "JVM_TOMCAT_E2E"},
+			{"tomcat.processing_time", "JVM_TOMCAT_E2E"},
+			{"tomcat.threads", "JVM_TOMCAT_E2E"},
```
varunch77, line 229: can we do this in a more elegant way if the namespace is the same for all of these metrics? do we expect this to change in the future?

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,405 @@
...
+			{"java_lang_operatingsystem_openfiledescriptorcount", "ContainerInsights/Prometheus"},
+			{"java_lang_operatingsystem_availableprocessors", "ContainerInsights/Prometheus"},
+			{"jvm_memory_bytes_used", "ContainerInsights/Prometheus"},
+			{"jvm_memory_pool_bytes_used", "ContainerInsights/Prometheus"},
+			{"catalina_manager_activesessions", "ContainerInsights/Prometheus"},
+			{"catalina_manager_rejectedsessions", "ContainerInsights/Prometheus"},
+			{"catalina_globalrequestprocessor_requestcount", "ContainerInsights/Prometheus"},
+			{"catalina_globalrequestprocessor_errorcount", "ContainerInsights/Prometheus"},
+			{"catalina_globalrequestprocessor_processingtime", "ContainerInsights/Prometheus"},
+		}
```
varunch77, line 340: same as above

varunch77 (changes requested):

test/e2e/jmx/jmx_test.go
musa-asad: Yeah good point, I can make it a constant.

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,411 @@
...
+	fmt.Println("Starting Helm installation...")
+	if err := ApplyHelm(env); err != nil {
+		fmt.Printf("Failed to apply Helm: %v\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Println("Waiting for metrics to propagate...")
+	time.Sleep(5 * time.Minute)
+
+	os.Exit(m.Run())
```
dricross, line 78: Do the helm resources ever get cleaned up?

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	return nil
+}
+
+func TestResources(t *testing.T) {
+	config, err := clientcmd.BuildConfigFromFlags("", filepath.Join(os.Getenv("HOME"), ".kube", "config"))
+	if err != nil {
+		t.Fatalf("Error building kubeconfig: %v", err)
+	}
```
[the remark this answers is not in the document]
musa-asad: Thanks, I'll make this change

test/e2e/jmx/jmx_test.go
musa-asad: With the exception of Load Balancers, they all should be removed upon the deletion of the cluster, but I'll look into handling this more elegantly.

test/e2e/jmx/resources/sample_apps/kafka-deployment.yaml
```diff
@@ -0,0 +1,122 @@
...
+    matchLabels:
+      app: zookeeper
+  template:
+    metadata:
+      labels:
+        app: zookeeper
+    spec:
+      containers:
+        - name: zookeeper
+          image: wurstmeister/zookeeper:latest
```
lisguo, line 28: Instead of using latest, can we pin a version so that it never changes?

Or maybe better yet -- we push an image to our own ECR repo?

test/e2e/jmx/resources/sample_apps/kafka-deployment.yaml
```diff
@@ -0,0 +1,122 @@
...
+    matchLabels:
+      app: kafka
+  template:
+    metadata:
+      labels:
+        app: kafka
+    spec:
+      containers:
+        - name: kafka
+          image: wurstmeister/kafka:latest
```
lisguo, line 62: similar comment to above

util/awsservice/cloudwatchmetrics.go
```diff
@@ -134,6 +134,78 @@ func GetMetricStatistics(
 	return CwmClient.GetMetricStatistics(ctx, &metricStatsInput)
 }
 
+func GetMetricMaximum(
```
lisguo, line 137: Curious to hear the reasoning behind this func...is it because we don't know the exact value of the metric? But we expect the value to never be above a certain amount?

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,411 @@
...
+	updateKubeconfig := exec.Command("aws", "eks", "update-kubeconfig", "--name", env.EKSClusterName)
+	output, err := updateKubeconfig.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to update kubeconfig: %w\nOutput: %s", err, output)
+	}
+
+	helm := []string{
+		"helm", "upgrade", "--install", "amazon-cloudwatch-observability",
+		filepath.Join("..", "..", "..", "terraform", "eks", "e2e", "helm-charts", "charts", "amazon-cloudwatch-observability"),
+		"--values", filepath.Join("..", "..", "..", "terraform", "eks", "e2e", "helm-charts", "charts", "amazon-cloudwatch-observability", "values.yaml"),
```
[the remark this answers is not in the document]
musa-asad: Yeah, I can remove it. Thanks.

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	helmUpgrade := exec.Command(helm[0], helm[1:]...)
+	helmUpgrade.Stdout = os.Stdout
+	helmUpgrade.Stderr = os.Stderr
+	if err := helmUpgrade.Run(); err != nil {
+		return fmt.Errorf("failed to install Helm release: %w", err)
+	}
+
+	fmt.Println("Waiting for CloudWatch Agent Operator to initialize...")
+	time.Sleep(300 * time.Second)
```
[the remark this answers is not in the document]
musa-asad: I'll lower the time.

test/e2e/jmx/resources/sample_apps/kafka-deployment.yaml
musa-asad: Good point, I'll add it to our own ECR repo.

test/e2e/jmx/resources/sample_apps/kafka-deployment.yaml
musa-asad: Will add.

util/awsservice/cloudwatchmetrics.go
musa-asad: The reasoning behind this function was to get the highest metric value being emitted for `tomcat.sessions` / `tomcat.rejected_sessions` in order to check if that value is greater than zero since we need to verify we're able to get these metrics above zero. The reason I went for the maximum was just to be sure we check for a value that is absolutely expected to be above zero. 

However, thinking through it, a more efficient solution would be to dedicate the function to check if it's above zero directly and exit out when we find that. I fixed the function to account for that.

As for the exact values for each metric, `tomcat.sessions` will max out to 2 since that's what I set it to in the application, but I didn't think it was necessary to directly check for this since the agent doesn't generate the metric, the application does. But it is important to make sure the agent is able to ship a non-zero value to CloudWatch. Another reason is a possible race condition where there are latency issues and it doesn't append to the right value in time compared to when we check the metric value, so it just seems more ideal to check if the value is above 0 instead of a specific value. `tomcat.rejected_sessions` is unpredictable since there isn't a max value for that, so it'll just keep increasing after the max sessions are hit.

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	deploymentName := strings.TrimSuffix(filepath.Base(env.SampleApp), ".yaml")
+
+	apply := exec.Command("kubectl", "apply", "-f", env.SampleApp)
+	output, err = apply.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to apply sample app: %w\nOutput: %s", err, output)
+	}
+
+	wait := exec.Command("kubectl", "wait", "--for=condition=available", "--timeout=300s", fmt.Sprintf("deployment/%s", deploymentName), "-n", "default")
```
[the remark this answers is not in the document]
musa-asad: Yeah, I can change it to be `-app` instead.

test/e2e/jmx/jmx_test.go
musa-asad: Added cleanup logic.

varunch77 (dismissed):

environment/metadata.go
```diff
@@ -136,6 +138,7 @@ func registerEKSData(d *MetaDataStrings) {
 func registerEKSE2ETestData(dataString *MetaDataStrings) {
 	flag.StringVar(&(dataString.Region), "region", "", "AWS region")
 	flag.StringVar(&(dataString.K8sVersion), "k8s_version", "", "Kubernetes version")
+	flag.StringVar(&(dataString.Destroy), "destroy", "false", "Whether to run in destroy mode (true/false)")
```
mitali-salvi, line 141: Can use https://pkg.go.dev/flag#BoolVar instead of using a string and later converting it

generator/resources/eks_e2e_jmx_test_matrix.json
```diff
@@ -0,0 +1,5 @@
+[
+  {
+    "nodes": 1
```
mitali-salvi, line 3: Does this imply just 1 node in the cluster ? Why just 1 ?

terraform/eks/e2e/main.tf
```diff
@@ -52,6 +52,15 @@ resource "aws_eks_node_group" "this" {
   ]
 }
 
+resource "aws_security_group_rule" "nodeport_inbound" {
```
mitali-salvi, line 55: Why is this required ?

terraform/eks/e2e/main.tf
```diff
@@ -134,4 +149,17 @@ resource "null_resource" "validator" {
...
+  provisioner "local-exec" {
+    when    = destroy
+    command = <<-EOT
+      echo "Running cleanup for K8s resources"
+      go test -timeout 30m -v ${self.triggers.test_dir} \
+      -destroy=true \
+      -region=${self.triggers.region} \
+      -eksClusterName=${self.triggers.cluster_name} \
+      -computeType=EKS \
+      -eksDeploymentStrategy=DAEMON
```
mitali-salvi, line 162: What's the alternative strategy here ?

terraform/eks/e2e/variables.tf
```diff
@@ -21,6 +21,16 @@ variable "nodes" {
   default = 1
 }
 
+variable "ami_type" {
```
mitali-salvi, line 24: I am curious do we actually need to use variables for the AMI/Instance type ? Are we testing a variety of AMIs/Instances for EKS ?

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,364 @@
...
+		require.NoError(t, err, "Error getting CloudWatch Agent Service")
+		require.NotNil(t, service, "CloudWatch Agent Service not found")
+
+		serviceAccount, err := clientset.CoreV1().ServiceAccounts("amazon-cloudwatch").Get(context.TODO(), "cloudwatch-agent", metav1.GetOptions{})
+		require.NoError(t, err, "Error getting CloudWatch Agent Service Account")
+		require.NotNil(t, serviceAccount, "CloudWatch Agent Service Account not found")
+	})
+}
+
+func testJMXResources(t *testing.T, clientset *kubernetes.Clientset) {
```
mitali-salvi, line 183: Can this functionality be made generic ? Feels like its checking pods status, env vars which could be re-used for other functionality testing as well

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,364 @@
...
+
+		require.Empty(t, requiredEnvVars, "Not all required environment variables were found in the pod")
+	})
+}
+
+//------------------------------------------------------------------------------
+// Metric Test Functions
+//------------------------------------------------------------------------------
+
+func testTomcatMetrics(t *testing.T) {
```
mitali-salvi, line 228: Similar comment to `testJMXResources`

test/e2e/jmx/jmx_test.go
```diff
@@ -0,0 +1,364 @@
...
+		resp, err := http.Get(fmt.Sprintf("http://%s:30080/webapp/index.jsp", nodeIP))
+		if err != nil {
+			t.Logf("Request attempt %d failed: %v", i+1, err)
+			continue
+		}
+		require.NoError(t, resp.Body.Close(), "Failed to close response body")
+	}
+}
+
+func verifyMetricAboveZero(t *testing.T, metricName, namespace string, containerInsights bool) {
```
mitali-salvi, line 349: So these tests are just checking the existence of metrics, not a value.

musa-asad: Most of the tests in `jmx_test.go` check for the existence of metrics to ensure the agent properly emits them. However, for `tomcat.sessions` and `tomcat.rejected_sessions`, the `verifyMetricAboveZero` function does more: it verifies if these metrics increase above 0 when we send load to the Tomcat application. This additional check was an important part of manual testing to ensure these specific metrics can emit non-zero values, which wasn't a concern for other metrics.

test/e2e/jmx/jmx_test.go
musa-asad: Yeah, I agree that it could be made more generic. I can adjust for that.

terraform/eks/e2e/variables.tf
musa-asad: You're right to question this. I don't think we need this, but I just thought it was worth adding this variable in case we wanted to configure it in the future. I could remove it since AMI testing for EKS doesn't seem necessary for the purpose of these tests. What do you think?

terraform/eks/e2e/main.tf
musa-asad: `REPLICA` or `SIDECAR` are the current options. I can add a `STATEFUL` type too and make this configurable for future testing. I didn't include it now because we don't change the deployment strategy from the default as of now.

terraform/eks/e2e/main.tf
musa-asad: We need this to be able to access the service that the Tomcat application uses in order to send load to it.

ref: https://github.com/aws/amazon-cloudwatch-agent-test/blob/8cb81ce05f18b584b05c7317295b6379bcd780f6/test/e2e/jmx/resources/sample_apps/tomcat.yaml#L18

generator/resources/eks_e2e_jmx_test_matrix.json
musa-asad: Yeah, we can re-configure this, but I just went for 1 since that seemed sufficient for these JMX tests. I can change the number if you think it's needed.

test/e2e/jmx/jmx_test.go
musa-asad: I moved `ValidateMetrics()` into `testutils.go`.

varunch77 (approved):

dricross (approved):

PR #450, opened 2025-01-16
Add pod identity credential integration test

zhihonl:
# Description of the issue
Pod identity credential support is added to CloudWatch Agent and FluentBit, however we lack integration test.

# Description of changes
- Add new terraform folder for credential related testing
- Add new test case when deployment strategy is pod identity. This mode only checks for basic Container Insight metrics that are present on any EKS cluster.

# License

# Tests
- Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12797706927/job/35689119613. Note that other test failures were there before this change and it is not in scope for fixing as part of this PR.
- integ test run rev2: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12952146544

### DIFF
--- a/environment/eksdeploymenttype/eks_deployment_type.go
+++ b/environment/eksdeploymenttype/eks_deployment_type.go
@@ -8,18 +8,20 @@ import "strings"
 type EKSDeploymentType string
 
 const (
-	DAEMON   EKSDeploymentType = "DAEMON"
-	REPLICA  EKSDeploymentType = "REPLICA"
-	SIDECAR  EKSDeploymentType = "SIDECAR"
-	STATEFUL EKSDeploymentType = "STATEFUL"
+	DAEMON      EKSDeploymentType = "DAEMON"
+	REPLICA     EKSDeploymentType = "REPLICA"
+	SIDECAR     EKSDeploymentType = "SIDECAR"
+	STATEFUL    EKSDeploymentType = "STATEFUL"
+	PODIDENTITY EKSDeploymentType = "PODIDENTITY"
 )
 
 var (
 	eksDeploymentTypes = map[string]EKSDeploymentType{
-		"DAEMON":   DAEMON,
-		"REPLICA":  REPLICA,
-		"SIDECAR":  SIDECAR,
-		"STATEFUL": STATEFUL,
+		"DAEMON":      DAEMON,
+		"REPLICA":     REPLICA,
+		"SIDECAR":     SIDECAR,
+		"STATEFUL":    STATEFUL,
+		"PODIDENTITY": PODIDENTITY,
 	}
 )
 
--- a/terraform/eks/daemon/credentials/pod_identity/main.tf
+++ b/terraform/eks/daemon/credentials/pod_identity/main.tf
@@ -1,0 +1,324 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+module "common" {
+  source = "../../../../common"
+}
+
+module "basic_components" {
+  source = "../../../../basic_components"
+
+  region = var.region
+}
+
+locals {
+  aws_eks = "aws eks --region ${var.region}"
+}
+
+data "aws_eks_cluster_auth" "this" {
+  name = aws_eks_cluster.this.name
+}
+
+resource "aws_eks_cluster" "this" {
+  name     = "cwagent-eks-integ-${module.common.testing_id}"
+  role_arn = module.basic_components.role_arn
+  version  = var.k8s_version
+  vpc_config {
+    subnet_ids         = module.basic_components.public_subnet_ids
+    security_group_ids = [module.basic_components.security_group]
+  }
+}
+
+# EKS Node Groups
+resource "aws_eks_node_group" "this" {
+  cluster_name    = aws_eks_cluster.this.name
+  node_group_name = "cwagent-eks-integ-node-${module.common.testing_id}"
+  node_role_arn   = aws_iam_role.node_role.arn
+  subnet_ids      = module.basic_components.public_subnet_ids
+
+  scaling_config {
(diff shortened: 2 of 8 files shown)


varunch77 (approved):

environment/eksdeploymenttype/eks_deployment_type.go
```diff
@@ -8,18 +8,20 @@ import "strings"
...
 const (
-	DAEMON   EKSDeploymentType = "DAEMON"
-	REPLICA  EKSDeploymentType = "REPLICA"
-	SIDECAR  EKSDeploymentType = "SIDECAR"
-	STATEFUL EKSDeploymentType = "STATEFUL"
+	DAEMON      EKSDeploymentType = "DAEMON"
+	REPLICA     EKSDeploymentType = "REPLICA"
+	SIDECAR     EKSDeploymentType = "SIDECAR"
+	STATEFUL    EKSDeploymentType = "STATEFUL"
+	PODIDENTITY EKSDeploymentType = "PODIDENTITY"
```
lisguo, line 15: Seems kinda interesting to put PodIdentity here...it's not necessarily a deployment type but a credential provider. Technically we can have podidentity enabled with any of these deployment options right?

terraform/eks/daemon/credentials/pod_identity/main.tf
```diff
@@ -0,0 +1,324 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+module "common" {
```
lisguo, line 4: just curious -- is there a way we can re-use existing terraform modules from our other eks tests? So that we aren't adding all this terraform code for every new eks test we add

terraform/eks/daemon/credentials/pod_identity/main.tf
```diff
@@ -0,0 +1,324 @@
...
+  description              = "Allow worker Kubelets and pods to receive communication from the cluster control plane"
+  from_port                = 1025
+  protocol                 = "tcp"
+  security_group_id        = aws_security_group.eks_nodes_sg.id
+  source_security_group_id = aws_security_group.eks_cluster_sg.id
+  to_port                  = 65535
+  type                     = "ingress"
+}
+
+resource "null_resource" "clone_helm_chart" {
```
lisguo, line 197: Is there a reason we are using helm over the eks addon?

Wondering what it is this integ test is testing -- I would assume we would want to install the addon + pod identity enabled + the release artifact of the agent = verify functionality works / calls to cw work

terraform/eks/daemon/credentials/pod_identity/main.tf
```diff
@@ -0,0 +1,324 @@
...
+    kubernetes_pod.log_generator,
+  ]
+
+  triggers = {
+    always_run = timestamp()
+  }
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      echo "Validating CloudWatch Agent and FluentBit with pod identity credential"
```
lisguo, line 317: technically...we are testing against cloudwatch agent, not fluentbit. Or are we pulling the latest fluentbit version in this test? 

And if we are...I don't think we should block our agent release if fluentbit fails.

test/metric_value_benchmark/eks_daemonset_test.go
```diff
@@ -31,7 +32,9 @@ type EKSDaemonTestRunner struct {
 
 func (e *EKSDaemonTestRunner) Validate() status.TestGroupResult {
 	var testResults []status.TestResult
-	testResults = append(testResults, metric.ValidateMetrics(e.env, "", eks_resources.GetExpectedDimsToMetrics(e.env))...)
+	if e.env.EksDeploymentStrategy != eksdeploymenttype.PODIDENTITY {
```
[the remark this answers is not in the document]
lisguo: > For more context, Container Insights provides different set of metrics based on what's present on the cluster. When testing Pod Identity I see a different set of metrics from this list: 

Is that a problem? What is the diff?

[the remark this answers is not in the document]
zhihonl: Not a problem, CI only shows certain metrics under certain situations. For example some error metrics are only shown when the failure case occur for pods. iirc for pod identity I saw additional API server related metrics being there, I'm guessing it's because of API interactions from pod identity addon.

environment/eksdeploymenttype/eks_deployment_type.go
zhihonl: Correct. I added here just as an easy way to have a different way of running EKS tests with basics without adding another variable in the class which felt redundant.

terraform/eks/daemon/credentials/pod_identity/main.tf
zhihonl: I used helm chart for easier testing. For example, I can check out a new branch from helm chart repo with custom changes, then run terraform on my local with the said changes. Using add-on have the same effect but will be harder to test custom changes.

terraform/eks/daemon/credentials/pod_identity/main.tf
zhihonl: I think should be possible but probably need to refactor some code. Don't think we should do it in this PR since that adds additional code that's not necessarily related to the goal of this PR.

terraform/eks/daemon/credentials/pod_identity/main.tf
zhihonl: Fixed in new commit

zhihonl: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12952146544 new integration test run. random surprise that all EKS tests passing 

Alt Text

lisguo (approved):

PR #455, opened 2025-01-24
Fix flaky cloudwatchlogs test

varunch77:
# Description of the issue
`TestLogGroupClass` uses fixed log group names in the tests. When multiple integration tests are run concurrently in the same account, they're all trying to operate and validate with the same log groups, leading to flaky behavior.

# Description of changes
The log groups created now have the instance id appended to the end of the name, ensuring there is no overlap. I've also added a 7-day log group retention policy in the agent configs.

# License

# Tests
Integration Test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12958277627


musa-asad (approved):

test/cloudwatchlogs/resources/config_log_infrequent_access.json
```diff
@@ -9,7 +9,7 @@
         "collect_list": [
           {
             "file_path": "/tmp/cwagent_log_test.log",
-            "log_group_name": "infrequent_access",
+            "log_group_name": "infrequent_access-{instance_id}",
             "log_stream_name": "{instance_id}",
             "timezone": "UTC",
             "log_group_class": "INFREQUENT_ACCESS"
```
jefchien, line 15: As @sky333999 mentioned, now that we're creating separate log groups per instance, we want to make sure to set the `retention_in_days` so they get cleaned up.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html

> Valid values are 1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, 2192, 2557, 2922, 3288, and 3653.

dricross: I think the log groups get deleted up at the end of the test anyways: https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/cloudwatchlogs/publish_logs_test.go#L101. I recall trying to find the infrequent_access log group to debug some issues and it didn't exist.

dricross: Oh, actually we would need to update that line to make sure the test deletes the right log group.

test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -249,6 +249,9 @@ func TestLogGroupClass(t *testing.T) {
 
 	for _, param := range cloudWatchLogGroupClassTestParameters {
 		t.Run(param.testName, func(t *testing.T) {
+			// add instance id to ensure that running integration tests concurrently doesn't cause tests
+			// to operate and validate with the same log groups and lead to flaky results
+			param.logGroupName += "-" + instanceId
 			defer awsservice.DeleteLogGroupAndStream(param.logGroupName, instanceId)
```
dricross, line 255: Mentioned in another comment, but we should update the logGroupName passed here too

musa-asad (approved):

test/cloudwatchlogs/publish_logs_test.go
musa-asad: Good catch.
```suggestion
			defer awsservice.DeleteLogGroupAndStream(param.logGroupName += "-" + instanceId, instanceId)
```

musa-asad (changes requested):

test/cloudwatchlogs/resources/config_log_infrequent_access.json
varunch77: Added the retention policy as a good-to-have for all of the agent config files in this folder, but we do call `defer awsservice.DeleteLogGroupAndStream()` in our code so this should handle it

test/cloudwatchlogs/publish_logs_test.go
varunch77: This should have already been handled since I was using `+=` here but I rewrote it to make it more clear now

varunch77:
Changes:
- Added log group retention policy in agent configs for all of our logs tests
- Checked that log group deletion was working as intended by commenting/uncommenting `defer awsservice.DeleteLogGroupAndStream()`
- New integration test run:

test/cloudwatchlogs/publish_logs_test.go
dricross: Ah good point, I missed that. Thanks Varun!

I think with that we don't really need to change log retention at all then since the log group and stream gets deleted. Though it also doesn't hurt to set it.

dricross (approved):

musa-asad (approved):

test/cloudwatchlogs/publish_logs_test.go
musa-asad: Ah missed that too. Yeah, keeping the log retention doesn't hurt.

jefchien (approved):

PR #469, opened 2025-02-22
Unblock China region integration tests

varunch77:
# Description of the issue
The EC2LinuxCN tests are not able to start up properly due to errors while fetching dependencies. This creates a gap in our testing for this region.

# Description of changes
This change adds a step to the existing terraform code to pull the test repo and dependencies from S3 instead of attempting to fetch them directly. This only runs for StartLocalStackCN and tests in the China region.

This PR is in conjunction with: https://github.com/aws/amazon-cloudwatch-agent/pull/1564

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. ✅ 

# Tests
Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13595857669

PR #478, opened 2025-03-13
Fix CN tests

varunch77:
# Description of the issue
We need to check to see if the test repo and dependencies are already present so we can avoid doing this step in the CN region (which will fail). More context: https://github.com/aws/amazon-cloudwatch-agent-test/pull/469

# Description of changes
This change re-adds a check to see if the repo exists. It also sets enforcing mode for selinux to `1` (enforcing mode) in order for selinux tests to run accurately and not silently fail

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. ✅ 

# Tests
Integration Test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13840967261


Paramadon (approved):
LGTM!

musa-asad (dismissed):

Paramadon (approved):

terraform/ec2/linux/main.tf
```diff
@@ -129,7 +135,7 @@ resource "null_resource" "integration_test_run" {
         "sudo yum install amazon-cloudwatch-agent -y",
         "echo Running SELinux test setup...",
         "sudo yum install selinux-policy selinux-policy-targeted selinux-policy-devel -y",
-        "sudo setenforce 0",
+        "sudo setenforce 1",
```
jefchien, line 138: Why are we changing this?

Paramadon: I recommended this change, this needed to be set to 1 which means enforcing mode and that test will fail if permissions are denied. If it is in set to 0 than it is permissive and even with permission denials the test will still pass since the operation will not be denied only logged. We can have a seperate pr for this but just thought we can add this in here since it has already been tested.

jefchien (approved):

PR #486, opened 2025-03-20
Updating AL2 ami and selinux tests 

Paramadon:
# Description of the issue
In the selinux installation script we added the ability for the customer to restart their agent which requires customer input. 

Selinux pr : https://github.com/aws/amazon-cloudwatch-agent-selinux/commit/8ec69f647f09594a394c6459a83651c9c81f7874
Which is why we need to update our tests to have -y when calling selinux script to pass the prompt.

As for the ami change, previous ami did not have the resources the instance needed or some directories already created example /mnt/lvm which was causing lvm failure for al2, this ami makes sure that all resources are like our original al2 integ test image just with selinux enabled. 

# Description of changes
This fixes selinux installation script issue.

# License

# Tests
SELinux test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13976059769/job/39130224897

### DIFF
--- a/terraform/ec2/linux/main.tf
+++ b/terraform/ec2/linux/main.tf
@@ -126,15 +126,13 @@ resource "null_resource" "integration_test_run" {
     inline = concat(
       [
         "echo Preparing environment...",
-        "sudo setenforce 0",
-        "sudo yum install -y audit policycoreutils-python-utils go --allowerasing",
       ],
 
       # SELinux test setup (if enabled)
       var.is_selinux_test ? [
         "sudo yum install amazon-cloudwatch-agent -y",
         "echo Running SELinux test setup...",
-        "sudo yum install selinux-policy selinux-policy-targeted selinux-policy-devel -y",
+        "sudo yum install selinux-policy selinux-policy-targeted policycoreutils-python-utils selinux-policy-devel -y",
         "sudo setenforce 1",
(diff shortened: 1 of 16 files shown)


TravisStark (approved):

terraform/ec2/linux/main.tf
varunch77, line 1: do you need the `--allowerasing` flag here?

terraform/ec2/linux/main.tf
```diff
@@ -126,15 +126,13 @@ resource "null_resource" "integration_test_run" {
...
-        "sudo setenforce 0",
-        "sudo yum install -y audit policycoreutils-python-utils go --allowerasing",
       ],
 
       # SELinux test setup (if enabled)
       var.is_selinux_test ? [
         "sudo yum install amazon-cloudwatch-agent -y",
         "echo Running SELinux test setup...",
-        "sudo yum install selinux-policy selinux-policy-targeted selinux-policy-devel -y",
+        "sudo yum install selinux-policy selinux-policy-targeted policycoreutils-python-utils selinux-policy-devel -y",
```
varunch77, line 135: do you need the --allowerasing flag here?

varunch77 (approved):

phugay (approved):

PR #503, opened 2025-04-11
Add integration tests for EBS NVMe metrics on EC2

duhminick:
# Description of the issue
Adding tests for EBS NVMe metrics when using EC2.

# Description of changes
1. Updating the Terraform to have a new instance disk size of 200gb. This is not particularly ideal. Some of the test cases resulted in out of space errors resulting in failures. As for why 200gb: 
```
$ aws ec2 describe-images --owners self --filters 'Name=state,Values=available' 'Name=manifest-location,Values=506463145083/cloudwatch-agent-integration-test*' --query 'Images[*].BlockDeviceMappings[*].Ebs.VolumeSize' | grep -E '[0-9]+' | sort -nr | head -n 10
        200
        140
        140
        140
        140
        140
        140
        140
        140
        100,
```
The disk size has to be greater than or equal to the snapshot size. For one of the AMIs used in testing, the snapshot has a volume size of 200.

2. Added a new VolumeIdDimensionProvider which will retrieve the volume ID of any EBS NVMe device attached to the host. Any usually means the one EBS volume attached.
3. New DiskIOEBSTestRunner to test the EBS metrics. This will run for all EC2 AMIs **except** when using SELinux. The SELinux policies that are required is way too permissive to be a default. We do not want to vend that out. 
 i. Checks that the metric is emitted
 ii. Checks that entity information is being properly sent
4. The SELinux check is done by checking `/sys/fs/selinux/enforce` for the value of 1 which is `Enforced`. I initially had passed the `is_selinux_test` flag from Terraform into the test arguments, but some of tests were not "registering" the flag. As for why... not sure. Spent the past 2 days trying to figure it out (a lot of it spent waiting on the tests to finish) 😭 

# License

# Tests
1. https://github.com/aws/amazon-cloudwatch-agent/actions/runs/14412522892 passed all the test artifacts
2. Tests that all of the EBS metrics are emitted to CloudWatch
3. Tests that all of the metrics have relevant entity information (EC2 instance)

### DIFF
--- a/test/metric_value_benchmark/diskio_ebs_test.go
+++ b/test/metric_value_benchmark/diskio_ebs_test.go
@@ -173,8 +173,8 @@ func (m *DiskIOEBSTestRunner) validateEBSEntity(metricName string) status.TestRe
 		return testResult
 	}
 
-	if len(response.Entities) != 0 {
-		log.Printf("Response contains no entities for metric '%s'", metricName)
+	if len(response.Entities) != 1 {
+		log.Printf("Response does not contain the correct number of entities for metric '%s'", metricName)
 		return testResult
 	}
 
--- a/test/metric_value_benchmark/diskio_ebs_test.go
+++ b/test/metric_value_benchmark/diskio_ebs_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"time"
 
 	"github.com/aws/amazon-cloudwatch-agent-test/environment"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/metric"
@@ -76,6 +77,10 @@ func (m *DiskIOEBSTestRunner) GetMeasuredMetrics() []string {
 		"diskio_ebs_ec2_instance_performance_exceeded_tp",
 		"diskio_ebs_volume_queue_length",
 	}
+}
+
+func (m *DiskIOEBSTestRunner) GetAgentRunDuration() time.Duration {
+	return 4 * time.Minute
 }
 
 func (m *DiskIOEBSTestRunner) validateEBSMetric(metricName string) status.TestResult {
--- a/test/metric_value_benchmark/diskio_ebs_test.go
+++ b/test/metric_value_benchmark/diskio_ebs_test.go
@@ -1,0 +1,102 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build !windows
+
+package metric_value_benchmark
+
+import (
+	"log"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/test/metric"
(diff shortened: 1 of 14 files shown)


util/common/agent_util_unix.go
```diff
@@ -268,3 +268,14 @@ func DownloadFromS3(bucket string, key string, destPath string) error {
 
 	return nil
 }
+
+func SELinuxEnforced() (bool, error) {
+	status, err := os.ReadFile("/sys/fs/selinux/enforce")
```
varunch77, line 273: nit: make the file path a const

test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -132,6 +133,12 @@ func getEc2TestRunners(env *environment.MetaData) []*test_runner.TestRunner {
 		if os.Getenv("AWS_REGION") == "us-west-2" {
 			ec2TestRunners = append(ec2TestRunners, &test_runner.TestRunner{TestRunner: &EntityMetricsTestRunner{test_runner.BaseTestRunner{DimensionFactory: factory}}})
 		}
+
+		// Only add the Disk IO EBS test if not running on SELinux
+		runningOnSELinux, _ := common.SELinuxEnforced()
```
varunch77, line 138: If there's an error when calling SELinuxEnforced(), this will just default into `runningOnSELinux` being false while losing awareness that we had an error. Is that what we're expecting to do if we can't read the file?

test/metric_value_benchmark/diskio_ebs_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if !metric.IsAllValuesGreaterThanOrEqualToExpectedValue(metricName, values, 0) {
+		return testResult
+	}
+
+	testResult.Status = status.SUCCESSFUL
+	return testResult
+}
+
+func (m *DiskIOEBSTestRunner) validateEBSEntity(metricName string) status.TestResult {
```
varunch77, line 130: nit put this in a util package so we can reuse this in the future

test/metric_value_benchmark/metrics_value_benchmark_test.go
duhminick: So the error is if the file doesn't exist, that can be if the AMI does not have support for it or if it's (maybe) not setup for SELinux. 

This is really only useful for the SELinux tests

util/common/agent_util_unix.go
duhminick: Most of the strings in this file are not const which is why I leaned towards not using one

test/metric_value_benchmark/diskio_ebs_test.go
duhminick: Yeah, that's fair. I think that would be a good thing to do. But it's already been done this way a couple different instances -- not really looking to refactor atm. Maybe it can be done as a follow up

varunch77 (approved):

PR #511, opened 2025-04-28
Adding retries to Prometheus EMF tests

Paramadon:
# Description of the issue
Currently test validates before metrics appear sometimes so in order to fix this issue adding a retry to make sure that test are not flaky:

Here is the flaky test fail: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/14673647451/job/41185739077 (test retried)

Here is the test run that shows test passing 10+ times: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/14710259879/job/41280752001

### DIFF
--- a/terraform/ec2/linux/main.tf
+++ b/terraform/ec2/linux/main.tf
@@ -126,11 +126,11 @@ resource "null_resource" "integration_test_run" {
     inline = concat(
       [
         "echo Preparing environment...",
+        "sudo yum install amazon-cloudwatch-agent -y",
       ],
 
       # SELinux test setup (if enabled)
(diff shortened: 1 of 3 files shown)


terraform/ec2/linux/main.tf
```diff
@@ -126,11 +126,11 @@ resource "null_resource" "integration_test_run" {
     inline = concat(
       [
         "echo Preparing environment...",
```
Paramadon, line 128: This move ensures agent dir exists.

musa-asad (approved):

varunch77 (approved):

PR #523, opened 2025-05-22
Add integration test for pushing histograms to PMD

dricross:
# Description of the issue
The CloudWatch Agent supports pushing OTLP histograms to CloudWatch via PMD but there are no integration tests for this functionality.

# Description of changes
This PR adds an integration test for pushing OTLP histograms to CloudWatch via PMD. Most of the work was done by @agarakan. I'm just pushing it over the finish line

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
* Abbreviated integration test (with most tests removed) passed successfully: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/15169490318
* Running full integration test now: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/15290831839

PR #529, opened 2025-05-30
Histogram test via EMF

Paramadon:
Creating test for histogram via otel->emf pipeline, test for pmd implemented here: [#523 (files)](https://github.com/aws/amazon-cloudwatch-agent-test/pull/523/files#diff-b0d3d7ec6a09c058253a5ce6599c18923114bbbfd8bbc5a67c6fb4dcd588e4c6).

Test run: [aws/amazon-cloudwatch-agent/actions/runs/15220612676](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/15220612676)

## Description
This PR adds integration tests for histogram metrics through the OTLP to EMF pipeline. The tests verify the correct processing and transformation of both delta and exponential histogram types from OTLP format to CloudWatch metrics via EMF. Tests are verified by instance id which is added to the metrics as an additional attribute.

## Test Coverage
Tests include:
- Delta Histogram metrics with aggregation temporality 1
  - Verifies correct bucket distribution
  - Validates all statistics (min, max, sum, average, sample count)
  - Tests attribute preservation
- Cumulative Exponential Histogram metrics with aggregation temporality 2
  - Tests exponential bucket calculations
  - Validates zero count handling
  - Verifies scale and threshold processing

## Implementation Details
- Uses OTLP JSON format for metric generation
- Sends metrics via HTTP to the local OTLP receiver (port 1234)
- Validates metrics in CloudWatch after EMF transformation


test/otlp_histograms/otlp_histograms_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+func TestOTLPMetrics(t *testing.T) {
+	startAgent(t)
+	instanceID := awsservice.GetInstanceId()
+	err := runOTLPPusher(instanceID)
+	assert.NoError(t, err)
+
+	time.Sleep(2 * time.Minute)
+
+	metrics := []struct {
```
dricross, line 38: I think we want a test case cumutalive histogram as well

test/otlp_histograms/otlp_histograms_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/aws/amazon-cloudwatch-agent-test/test/metric"
+	"github.com/aws/amazon-cloudwatch-agent-test/util/awsservice"
+	"github.com/aws/amazon-cloudwatch-agent-test/util/common"
+)
+
+func init() {
+	environment.RegisterEnvironmentMetaDataFlags()
+}
+
+func TestOTLPMetrics(t *testing.T) {
```
dricross, line 30: Any reason youre not using the standard test suite?

Paramadon: The test suite is good if you're going to have mutlple test runners, but since this is just one test, I think just having this singular test is fine. This is also seen in other tests already as well, like the agent_otel_merging.

PR #536, opened 2025-06-11
Prometheus->Emf stress test

Paramadon:
## Description
Currently the cloudwatch agent is using telegraf prometheus receiver which does not have full support of histograms, so in order to give full prometheus support to customers, the agent will use the otel prometheus receiver. In order to make sure that there are no degradation in performance we are creating this stress test that goes over these cases:

Note: The scrape interval and prometheus metrics are configurable, so if there are a different scenario that needs to be tested, the parameters of the stress test can be modified and tested manually, this is just the base tests that will run to ensure no degradation are happening with any commits and if ever there needs to be a seperate stress that with different intervals or # of metric that needs to be done, is possible.

### Test parameters
Scrape inverval: 60s
~1000 prometheus metrics(guages,summary,counters)
~5000 prometheus metrics(guages,summary,counters)
~10000 prometheus metrics(guages,summary,counters)
~50000 prometheus metrics(guages,summary,counters)

### Test run
All test thresholds were decided in this run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/15620252133/job/44003091000


util/common/metrics.go
```diff
@@ -26,13 +32,28 @@ import (
...
 const SleepDuration = 5 * time.Second
 
 const TracesEndpoint = "4316/v1/traces"
 const MetricEndpoint = "4316/v1/metrics"
+const TMPAGENTPATH = "/tmp/agent_config.json"
+
+//go:embed prometheus/prometheus.yaml
+var prometheusTemplate string
+
+type PrometheusConfig struct {
```
agarakan, line 50: It seems a little confusing that you have a metrics util file but you're actually setting up your prometheus config and configuring metric collection. Is there a reason this logic isnt kept in the specific test folder? Is there a scope issue where you're accessing these methods from beyond the test/stress/prometheus directory?

util/common/metrics.go
```diff
@@ -76,6 +106,254 @@ func SendAppSignalsTraceMetrics(duration time.Duration) error {
...
+}
+
+func cleanupPortPrometheus(port int) {
+	killCmd := exec2.Command("sudo", "fuser", "-k", fmt.Sprintf("%d/tcp", port))
+	if err := killCmd.Run(); err != nil {
+		log.Printf("[Prometheus] Failed to kill port %d: %v", port, err)
+	}
+}
+
+func getAvalancheParams(metricPerInterval int) (counter, gauge, summary, series, label int) {
```
agarakan, line 271: Definitely doesnt belong in a common/metrics.go file. Can you please split these into separate files, where one defines prometheus config with avalanche separately from your log and metric validation logic?

validator/validators/basic/basic_validator.go
```diff
@@ -55,7 +55,13 @@ func (s *BasicValidator) GenerateLoad() error {
 	case "traces":
 		return traces.StartTraceGeneration(receiver, agentConfigFilePath, agentCollectionPeriod, metricSendingInterval)
 	default:
-		// Sending metrics based on the receivers; however, for scraping plugin  (e.g prometheus), we would need to scrape it instead of sending
+		if receiver == "prometheus" {
+			if s.vConfig.GetScrapeInterval() == 0 {
```
agarakan, line 60: please refactor this to be cleaner. Should read something like

```
if s.vConfig.GetScrapeInterval() != 0 {
  metricSendingInterval = time.Duration(s.vConfig.GetScrapeInterval()) * time.Second
  log.Printf("Using defined scrape interval for metric sending interval: %v seconds", metricSendingInterval.Seconds())
} 
```

Also I think you can print the log statement outside the if/else so that the scrapeInterval is always printed out

util/common/metrics.go
```diff
@@ -76,6 +106,254 @@ func SendAppSignalsTraceMetrics(duration time.Duration) error {
...
+	if err != nil {
+		log.Printf("[Prometheus] Failed to write config: %v", err)
+		return err
+	}
+
+	log.Printf("[Prometheus] Successfully wrote config file")
+	return nil
+}
+
+// Behavior:
```
agarakan, line 305: Block quotes read cleaner, and should always be preferred for multiline comments

```
/*
Example block quote
1. line one
  - sub text
  - sub text

2. line two
  - sub text
  - I am a chicken

*/
```

util/common/metrics.go
```diff
@@ -76,6 +106,254 @@ func SendAppSignalsTraceMetrics(duration time.Duration) error {
...
+// 1. On the first run:
+//   - Replaces "CloudWatchAgentStress/Prometheus".
+//   - Replaces it with "CloudWatchAgentStress/Prometheus/{instanceID}/1".
+//
+// 2. On subsequent runs (if stress test retried):
+//   - Detects an existing namespace in the form "CloudWatchAgentStress/Prometheus/{instanceID}/{index}".
+//   - Increments the {index} by 1 and replaces it with the new value.
+//
+// The function writes the updated config back to the file and returns the new namespace used.
+func updateAgentConfigNamespacePrometheus(configPath string, instanceID string) string {
```
agarakan, line 315: I dont follow this method exactly, maybe we can go over it and what its doing. Glad we're using regex but curious why the method seems to to have debug-style logs

util/common/metrics.go
```diff
@@ -76,6 +106,254 @@ func SendAppSignalsTraceMetrics(duration time.Duration) error {
...
+		Dimensions [][]string `json:"Dimensions"`
+		Metrics    []struct {
+			Name              string `json:"Name"`
+			Unit              string `json:"Unit"`
+			StorageResolution int    `json:"StorageResolution"`
+		} `json:"Metrics"`
+	} `json:"CloudWatchMetrics"`
+}
+
+func countMetricsInLogs(logGroupName string) (int, error) {
```
agarakan, line 121: Can you simplify this method by leveraging common methods in  `/util/awsservice/cloudwatchlogs.go` such as `AssertLogContainsSubstring`  with an expected minimum count?

agarakan (commented):
Left some comments for you

validator/validators/basic/basic_validator.go
Paramadon: Cleaned it up, the log statement needs to stay within the if statement since if it is not defined than there is no change in the metric sending interval.

util/awsservice/cloudwatchlogs.go
```diff
@@ -244,7 +244,7 @@ func GetLogStreams(logGroupName string) []types.LogStream {
 			LogGroupName: aws.String(logGroupName),
 			OrderBy:      types.OrderByLastEventTime,
 			Descending:   aws.Bool(true),
-			Limit:        aws.Int32(10),
+			Limit:        aws.Int32(10000),
```
Paramadon, line 247: Limit needs to be increase since we can get a lot more metrics than just 10, 10000 is the limit of the number of log events we can pull.

util/common/metrics.go
agarakan: capturing notes from syncup: 

We need to introduce a clean up that will delete the loggroups/logstreams after the tests are over

util/common/metrics.go
agarakan: Reviewed together, it makes sense this is following the existing pattern. Gave suggestions for improvement, not a blocking issue

validator/validators/stress/stress_validator.go
```diff
@@ -231,7 +255,7 @@ var (
 				"net_packets_sent":     float64(100),
 			},
 			"emf": {
-				"procstat_cpu_usage":   float64(165),
+				"procstat_cpu_usage":   float64(130),
```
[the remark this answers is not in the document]
Paramadon: no this should have not changed, going to revert this.

validator/validators/stress/stress_validator.go
```diff
@@ -406,8 +438,8 @@ func (s *StressValidator) ValidateStressMetric(metricName, metricNamespace strin
 
 	// Validate if the metrics are not dropping any metrics and able to backfill within the same minute (e.g if the memory_rss metric is having collection_interval 1
 	// , it will need to have 60 sample counts - 1 datapoint / second)
-	if ok := awsservice.ValidateSampleCount(metricName, metricNamespace, metricDimensions, startTime, endTime, metricSampleCount-5, metricSampleCount, int32(boundAndPeriod)); !ok {
-		return fmt.Errorf("\n metric %s is not within sample count bound [ %d, %d]", metricName, metricSampleCount-5, metricSampleCount)
+	if ok := awsservice.ValidateSampleCount(metricName, metricNamespace, metricDimensions, startTime, endTime, metricSampleCount-10, metricSampleCount, int32(boundAndPeriod)); !ok {
+		return fmt.Errorf("\n metric %s is not within sample count bound [ %d, %d]", metricName, metricSampleCount-10, metricSampleCount)
```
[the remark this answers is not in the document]
Paramadon: The sample count is flaky for other stress tests too, we get 293 sample count sometimes where we should get 295-300 as that what the -5 was for but since we get 293/294  sometimes, changing it to -10 makes more sense bounds will be [290-300] instead of [295-300]

agarakan (approved):

util/common/metrics.go
```diff
@@ -76,6 +102,77 @@ func SendAppSignalsTraceMetrics(duration time.Duration) error {
...
+
+	if err != nil {
+		return fmt.Errorf("E! error counting metrics: %v", err)
+	}
+
+	// This is just to generally verify that we are getting metrics in the emf logs
+	if count < config.MetricCount {
+		return fmt.Errorf("insufficient metrics generated: expected ~%d, got %d", config.MetricCount, count)
+	}
+	awsservice.DeleteLogGroup(namespaceAndLogGroup)
```
dricross, line 171: nit: this should be `defer`'d so that it runs if the test returns an error

agarakan (approved):

agarakan (approved):

validator/validators/stress/stress_validator.go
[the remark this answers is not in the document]
TravisStark: Good idea to adjust this, the variance usually fluctuated > 5 on prior stress tests runs.

PR #548, opened 2025-06-26
Adding prometheus metrics via PMD test

Paramadon:
# Description of the issue
Currently customers cannot send prometheus metrics via pmd so in order to be able to do that we need an integ test to verify that we get prometheus metrics via pmd.  

# Description of changes
Added agent change to add cloudwatch as a destination in prometheus pipeline in this pr: https://github.com/aws/amazon-cloudwatch-agent/pull/1742. 

Ran an integ test against the agent branch that is in the above pr (prometheusPMDSupport) on this test branch here is the passing test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/15905800422/job/44860425431


test/metric_value_benchmark/prometheus_pmd_test.go
```diff
@@ -0,0 +1,174 @@
...
+		Name:   metricName,
+		Status: status.FAILED,
+	}
+
+	dims, failed := t.DimensionFactory.GetDimensions([]dimension.Instruction{
+		{
+			Key:   "InstanceId",
+			Value: dimension.ExpectedDimensionValue{aws.String(awsservice.GetInstanceId())},
+		},
+	})
```
dricross, line 107: There are some more dimensions we should probably check the values of as part of this test

cmd/prometheus-generator/prometheus-generator.go
dricross, line 1: Very cool tool! Great addition to have

test/metric_value_benchmark/prometheus_pmd_test.go
```diff
@@ -0,0 +1,174 @@
...
+			log.Printf("Failed to fetch %s for metric %s: %v", stat, metricName, err)
+			return testResult
+		}
+		statValues[stat] = values
+	}
+
+	// Validate based on metric type
+	switch metricName {
+	case "prometheus_test_histogram":
+		// For histogram, values should be between 0-10
```
dricross, line 130: Is it possible to be more strict on the histogram values rather than just checking min/max? I think we'd at least want to check sum/sample count too so those are all parts of the StatisticsSet

PR #567, opened 2025-08-13
Add environment variable expansion test

dricross:
# Description of the issue
We want to ensure that regex replacement variable syntax (e.g. `$1` and `${2}`) in the prometheus configuration files is not interpreted as environment variables when processed by the translator/agent.

# Description of changes
Update the emf_prometheus relabel test to include a new metric relabel configuration which uses both forms of the replacement variable syntax. The new metric relabel config also uses the `$` syntax as part of the `regex` field to ensure that isn't interpreted as an environment variable either.

# License

# Tests
Ran the test locally and passes.

Kicking off full integration test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16943457816
Passing emf_prometheus test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16943457816/job/48018598148


test/emf_prometheus/common.go
```diff
@@ -35,12 +36,21 @@ func setupPrometheus(prometheusConfig, prometheusMetrics string, jobName string)
...
+	if err := os.WriteFile("/tmp/prometheus.yaml", []byte(configContent), os.ModePerm); err != nil {
+		return fmt.Errorf("unable to write to /tmp/prometheus.yaml: %w", err)
+	}
+	if err := os.WriteFile("/tmp/metrics", []byte(prometheusMetrics), os.ModePerm); err != nil {
+		return fmt.Errorf("unable to write to /tmp/metrics: %w", err)
+	}
 	commands := []string{
-		fmt.Sprintf("cat <<EOF | sudo tee /tmp/prometheus.yaml\n%s\nEOF", configContent),
-		fmt.Sprintf("cat <<EOF | sudo tee /tmp/metrics\n%s\nEOF", prometheusMetrics),
 		"sudo python3 -m http.server 8101 --directory /tmp &> /dev/null &",
```
dricross, line 52: Ran into some strange behavior here while writing this test. Previously, we were using `cat <<EOF | sudo tee ...` to write a file on disk because the integration test process runs as `ec2-user` which does not have permissions to write to `/tmp` (only root has permissions). When I added the `$1` and `${2}` syntax to the prometheus yaml, this bash commands actually interprets them as environment variables and replaces the values in the final output. So on disk, the metric relabel config looks something like:
```
      - source_labels: ['prom_type','include']
        separator: '@'
        regex: '^(.*)@(.*)$'
        target_label: 'my_replacement_test'
        action: replace
        replacement: '/'
```

Re-factored how we write the files to avoid this by:
* Allow anyone to write to /tmp using `sudo chmod 777 /tmp`
* Using `os.WriteFile` to write the contents directly to the output file.

movence (approved):

PR #575, opened 2025-08-28
Paramadon/instance store test

Paramadon:
# Description of the issue
Currently we only have integration tests for EBS metric and nothing for Instance Store metrics.

# Description of changes
This change adds integration test for Instance store metrics following a similar pattern of the ebs integration test. 

# Tests
Testing in a test instance with instance store metrics available:

Screenshot 2025-08-28 at 11 03 06 AM

### DIFF
--- a/terraform/ec2/common/linux/main.tf
+++ b/terraform/ec2/common/linux/main.tf
@@ -75,6 +75,11 @@ resource "aws_instance" "cwagent" {
     volume_size = 200
   }
 
+  ephemeral_block_device {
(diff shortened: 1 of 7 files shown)


terraform/ec2/common/linux/main.tf
```diff
@@ -75,6 +75,11 @@ resource "aws_instance" "cwagent" {
     volume_size = 200
   }
 
+  ephemeral_block_device {
```
duhminick, line 78: So this will get replaced at some point?

PR #671, opened 2026-04-21
Add OTEL MetricsV2 standard cluster integration tests

spanaik:
Port OTEL integration tests into the agent test framework for the
  standard EKS cluster (2x t3.medium). Tests validate metric correctness
  by querying the CloudWatch PromQL API with SigV4 auth and cross-
  validating against K8s API ground truth.

  Components:
  - util/otelmetrics/: Shared PromQL client library with SigV4 signing, query cache, ZIP-0006 label parsing, and rate limiting
  - terraform/eks/daemon/otel/: Ephemeral EKS cluster with Helm chart (release-6.1.0), Pod Identity, and nginx test workload
  - test/otel/standard/: 145 integration tests covering cadvisor, node_exporter, kubeletstats, KSM, control plane, dedup, resolution, host validation, and cross-source label correctness
  - generator/test_case_generator.go: Added to eks_daemon matrix

  Validated in CI: 145/145 PASS via existing EKSIntegrationTest workflow.

Workflow Run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/24726297813

### DIFF
--- a/terraform/eks/daemon/otel/main.tf
+++ b/terraform/eks/daemon/otel/main.tf
@@ -1,0 +1,242 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+module "common" {
+  source             = "../../../common"
+  cwagent_image_repo = var.cwagent_image_repo
+  cwagent_image_tag  = var.cwagent_image_tag
+}
+
+module "basic_components" {
+  source = "../../../basic_components"
+  region = var.region
+}
+
+locals {
+  aws_eks = "aws eks --region ${var.region}"
+}
+
+resource "aws_eks_cluster" "this" {
+  name     = "cwagent-eks-integ-${module.common.testing_id}"
+  role_arn = module.basic_components.role_arn
+  version  = var.k8s_version
+  vpc_config {
+    subnet_ids         = module.basic_components.public_subnet_ids
+    security_group_ids = [module.basic_components.security_group]
+  }
+}
+
+# EKS Node Group — 2x t3.medium with node-color=blue label
+resource "aws_eks_node_group" "this" {
+  cluster_name    = aws_eks_cluster.this.name
+  node_group_name = "cwagent-otel-integ-node-${module.common.testing_id}"
+  node_role_arn   = aws_iam_role.node_role.arn
+  subnet_ids      = module.basic_components.public_subnet_ids
+
+  scaling_config {
+    desired_size = 2
+    max_size     = 2
+    min_size     = 2
+  }
+
+  ami_type       = var.ami_type
+  capacity_type  = "ON_DEMAND"
+  disk_size      = 20
+  instance_types = [var.instance_type]
+
+  labels = {
+    "ci-test.example.com/node-color" = "blue"
+  }
+
+  depends_on = [
+    aws_iam_role_policy_attachment.node_AmazonEC2ContainerRegistryReadOnly,
+    aws_iam_role_policy_attachment.node_AmazonEKS_CNI_Policy,
+    aws_iam_role_policy_attachment.node_AmazonEKSWorkerNodePolicy,
+  ]
+}
+
+# EKS Node IAM Role
+resource "aws_iam_role" "node_role" {
+  name = "cwagent-otel-eks-Worker-Role-${module.common.testing_id}"
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Effect    = "Allow"
+      Principal = { Service = "ec2.amazonaws.com" }
+      Action    = "sts:AssumeRole"
+    }]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "node_AmazonEKSWorkerNodePolicy" {
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy"
+  role       = aws_iam_role.node_role.name
+}
+
+resource "aws_iam_role_policy_attachment" "node_AmazonEKS_CNI_Policy" {
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy"
+  role       = aws_iam_role.node_role.name
+}
+
+resource "aws_iam_role_policy_attachment" "node_AmazonEC2ContainerRegistryReadOnly" {
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
+  role       = aws_iam_role.node_role.name
+}
+
+
+# Pod Identity IAM Role
+resource "aws_iam_role" "pod_identity_role" {
+  name = "cwagent-otel-pod-identity-${module.common.testing_id}"
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Effect    = "Allow"
+      Principal = { Service = "pods.eks.amazonaws.com" }
+      Action    = ["sts:AssumeRole", "sts:TagSession"]
+    }]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "pod_identity_CloudWatchAgentServerPolicy" {
+  policy_arn = "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy"
+  role       = aws_iam_role.pod_identity_role.name
+}
+
+# --- EKS Addon: Pod Identity agent ---
+
+resource "aws_eks_addon" "pod_identity_agent" {
+  depends_on   = [aws_eks_node_group.this]
+  cluster_name = aws_eks_cluster.this.name
+  addon_name   = "eks-pod-identity-agent"
+}
+
+# --- Update kubeconfig ---
+
+resource "null_resource" "kubectl" {
+  depends_on = [aws_eks_cluster.this, aws_eks_node_group.this]
+  provisioner "local-exec" {
+    command = "${local.aws_eks} update-kubeconfig --name ${aws_eks_cluster.this.name}"
+  }
+}
+
+# --- Helm chart install ---
+
+data "external" "clone_helm_chart" {
+  program = ["bash", "-c", <<-EOT
+    rm -rf ./helm-charts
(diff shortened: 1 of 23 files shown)


louisall:
I know these tests run in us-west-2 but could be worth putting a region guard. If anyone adds a different region to the matrix that doesn't support metrics v2 it'll all just silently fail

util/otelmetrics/query_cache.go
```diff
@@ -0,0 +1,213 @@
...
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strings"
+	"sync"
+)
+
+// DefaultMaxConcurrency is the default number of concurrent in-flight queries.
+const DefaultMaxConcurrency = 1
```
louisall: Why 1?

test/otel/standard/setup_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	cfg = otelmetrics.TestConfig{
+		Region:         region,
+		Endpoint:       fmt.Sprintf("https://monitoring.%s.amazonaws.com", region),
+		Timeout:        30 * time.Second,
+		MaxRetries:     3,
+		ClusterName:    clusterName,
+		AccountID:      *identity.Account,
+		SigningService: "monitoring",
+		QueryDelay:     500 * time.Millisecond,
```
louisall: Do we really need this? Is it just waiting for metrics to land?

test/otel/standard/assertions_test.go
```diff
@@ -0,0 +1,118 @@
...
+	t.Helper()
+	if len(results) == 0 {
+		if len(msgAndArgs) > 0 {
+			t.Fatalf(msgAndArgs[0].(string), msgAndArgs[1:]...)
+		}
+		t.Fatalf("expected non-empty results, got 0")
+	}
+}
+
+func filterByHostType(results []otelmetrics.MetricResult, instanceType string) []otelmetrics.MetricResult {
```
louisall: │ `filterByHostType`, `uniqueValues`, `uniqueDatapointValues`, `uniqueAnyValues`, and `getAnyValue` are declared in this file but not referenced anywhere in the PR. They were ported from the reference repo where they had callers. Either port the tests that use them or drop them to keep this first PR lean

util/otelmetrics/query_cache.go
spanaik: Removed

test/otel/standard/setup_test.go
spanaik: It was creating a delay between two Query Calls, removed it now.

test/otel/standard/assertions_test.go
spanaik: Removed will bring it back with the other other tests.

util/otelmetrics/client.go
```diff
@@ -0,0 +1,283 @@
...
+// OtelMetricsClient queries the OTLP PromQL API with SigV4 authentication.
+type OtelMetricsClient struct {
+	httpClient     *http.Client
+	signer         *v4.Signer
+	creds          aws.CredentialsProvider
+	queryURL       string
+	region         string
+	signingService string
+	maxRetries     int
+	mu             sync.Mutex
```
sky333999: I dont see this `mu` being used anywhere - is it supposed to be?

terraform/eks/daemon/otel/main.tf
```diff
@@ -0,0 +1,252 @@
...
+    }]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "pod_identity_CloudWatchAgentServerPolicy" {
+  policy_arn = "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy"
+  role       = aws_iam_role.pod_identity_role.name
+}
+
+resource "aws_iam_role_policy_attachment" "pod_identity_AWSXrayWriteOnlyAccess" {
```
sky333999: nit: Dont need this.. should be included with `CloudWatchAgentServerPolicy`.

terraform/eks/daemon/otel/variables.tf
```diff
@@ -0,0 +1,42 @@
...
+}
+
+variable "cwagent_image_tag" {
+  type    = string
+  default = "latest"
+}
+
+variable "helm_chart_branch" {
+  type    = string
+  default = "release-6.1.0"
```
sky333999: nit: Id default to `main` here, but we should always pass a `release` branch when invoking this from the GH workflow.

On a related note, we should think through how this gets updated on every new release.

terraform/eks/daemon/otel/variables.tf
```diff
@@ -0,0 +1,42 @@
...
+}
+
+variable "k8s_version" {
+  type    = string
+  default = "1.31"
```
sky333999: Can we use the current latest version 1.35 as the default?

util/otelmetrics/client.go
spanaik: Removed, it was part of the throttle on the client.

terraform/eks/daemon/otel/variables.tf
spanaik: Makes sense, i will update the GH workflow to add the helm chart branch input, also automate picking up the latest release-* branch if not provided.

image

Something like this.

spanaik: PR for this: https://github.com/aws/amazon-cloudwatch-agent/pull/2099

Test Run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/24829129318

terraform/eks/daemon/otel-multi-efa/helm-charts
sky333999, line 1: Are we adding `helm-charts` as a git submodule? Dont think this should be here?

terraform/eks/daemon/otel/main.tf
```diff
@@ -0,0 +1,247 @@
...
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy"
+  role       = aws_iam_role.node_role.name
+}
+
+resource "aws_iam_role_policy_attachment" "node_AmazonEC2ContainerRegistryReadOnly" {
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
+  role       = aws_iam_role.node_role.name
+}
+
+resource "aws_iam_role_policy_attachment" "node_CloudWatchAgentServerPolicy" {
```
sky333999: Dont attach this to the node role so that your PodIdentity stuff is actually validated and if it fails, it wont silently fall back to the node role.

util/otelmetrics/query_cache.go
```diff
@@ -0,0 +1,213 @@
...
+// Get returns cached results for a metric filtered by cluster name.
+func (qc *QueryCache) Get(ctx context.Context, metricName string) ([]MetricResult, error) {
+	qc.mu.RLock()
+	if entry, ok := qc.filtered[metricName]; ok {
+		qc.mu.RUnlock()
+		return entry.results, entry.err
+	}
+	qc.mu.RUnlock()
+
+	qc.mu.Lock()
```
sky333999, line 100: 🤖 : 
```
  The write lock is held for the entire miss path, which includes the blocking HTTP calls. While the tests run sequentially today (no t.Parallel()), this design is fragile: if anyone enables t.Parallel() on these subtests, the cache becomes a single-flight serialization point across all metrics, not just per metric.
  The intra-metric fan-out via qc.sem ends up executing inside the held lock anyway, so the semaphore's inter-metric budget is effectively unused.

  Fix options:

  - Double-check the miss path with per-metric sentinels (e.g. x/sync/singleflight) so different metric names proceed concurrently.
  - Or release the lock after adding a placeholder entry, do the HTTP work, then re-acquire to store.

  Either approach keeps the RLock fast-path on cache hits intact.
```

util/otelmetrics/models.go
```diff
@@ -0,0 +1,180 @@
...
+
+// EscapePromQLValue escapes a string for use as a PromQL label value inside double quotes.
+func EscapePromQLValue(value string) string {
+	value = strings.ReplaceAll(value, `\`, `\\`)
+	value = strings.ReplaceAll(value, `"`, `\"`)
+	return value
+}
+
+// EscapeAt escapes literal @ in attribute values for PromQL (ZIP-0006: @@ escaping).
+func EscapeAt(value string) string {
```
sky333999: Are `EscapeAt` and `UnescapeAt` actually being used anywhere besides their own unit tests?

test/otel/standard/assertions_test.go
```diff
@@ -0,0 +1,53 @@
...
+package standard
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/util/otelmetrics"
+)
+
+func requireEqual(t *testing.T, got, want any, msgAndArgs ...any) {
```
sky333999: Do we need these? Any reason we cant use  github.com/stretchr/testify/require?

util/otelmetrics/query_cache.go
sky333999: 🤖 :
```
Once ^ is addressed, adding t.Parallel() at the outer TestXxx level would significantly cut wall-clock time — almost every test is cache-backed reads of the same metric set, and the ground-truth client is already guarded by sync.Once.
```

terraform/eks/daemon/otel-multi-efa/helm-charts
spanaik: Was a miss from previous testing

terraform/eks/daemon/otel/main.tf
spanaik: ACK

util/otelmetrics/query_cache.go
spanaik: Fixed, lemme know if this works

util/otelmetrics/models.go
spanaik: Yep removed

test/otel/standard/assertions_test.go
spanaik: Done migrated to use testify/require.

sky333999 (approved):

PR #674, opened 2026-04-29
fix(eks): migrate all AMI types from AL2 to AL2023 for K8s 1.33+

musa-asad:
# Description of the issue

EKS test clusters use deprecated AL2 AMI types which are no longer supported on K8s 1.33+. Additionally, AL2023 managed node groups default to IMDS hop limit 1, blocking pod IMDS access and breaking Container Insights.

# Description of changes

1. **Migrate AMI types to AL2023:**
   - `AL2_x86_64` → `AL2023_x86_64_STANDARD`
   - `AL2_ARM_64` → `AL2023_ARM_64_STANDARD`
   - `AL2_x86_64_GPU` → `AL2023_x86_64_NVIDIA`
   - Updated in: test matrix JSON, generator, all variables.tf defaults
   - Replaced hardcoded ami_type in main.tf with var.ami_type

2. **Fix IMDS access for pods on AL2023:**
   - Added `aws_launch_template` with `http_put_response_hop_limit = 2`
   - Removed `disk_size` from node groups (incompatible with launch templates)

3. **Add host_network = true to daemonset pod specs:**
   - AL2023 hop limit 1 blocks IMDS from pods without hostNetwork
   - Production helm/add-on already sets hostNetwork; test terraform was missing it

# License

# Tests

https://github.com/aws/amazon-cloudwatch-agent/actions/runs/25810039514

### DIFF
--- a/generator/resources/eks_addon_test_matrix.json
+++ b/generator/resources/eks_addon_test_matrix.json
@@ -1,7 +1,7 @@
 [
   {
     "k8sVersion": "1.31",
-    "ami": "AL2_x86_64_GPU",
+    "ami": "AL2023_x86_64_NVIDIA",
     "terraform_dir": "terraform/eks/addon/gpu",
     "test_dir": "./test/gpu",
     "instanceType":"g4dn.xlarge"
--- a/generator/resources/eks_daemon_test_matrix.json
+++ b/generator/resources/eks_daemon_test_matrix.json
@@ -1,13 +1,13 @@
 [
   {
     "k8sVersion": "1.35",
-    "ami": "AL2_x86_64",
+    "ami": "AL2023_x86_64_STANDARD",
     "instanceType":"t3.medium",
     "arc": "amd64"
   },
   {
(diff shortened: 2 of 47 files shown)


Paamicky:
1. I think a little explanation to why efa cluster uses a different mechanism for access entries `authentication_mode = "API"` rather than `access_config ` used by all other clusters would be helpful for the future
2. Multiple files have 3-4 blank lines added at the end. Cosmetic but worth cleaning.

musa-asad:
Thanks for the review!

1. **`authentication_mode` vs `access_config`**: They're actually the same thing — `authentication_mode = "API"` is set *inside* the `access_config` block on all clusters. The EFA cluster uses the same pattern:
   ```hcl
   access_config {
     authentication_mode = "API"
   }
   ```
   If the formatting looks different, it's just HCL style variation — functionally identical across all clusters. I'll double-check the EFA file to make sure it's consistent.

2. **Trailing blank lines**: Will clean those up. Thanks for catching it.

PR #690, opened 2026-05-11
Implement ApplicationSignals Logs integ tests for dynamic placeholders for otlp headers, and add Metrics tests for routing & OTLP endpoint

jj22ee:
# Description of the issue and changes

Add tests for changes in: https://github.com/aws/amazon-cloudwatch-agent/pull/2111

# License

# Tests
- `TestAppSignalsLogsDynamicRouting` — Multiple services route to separate log groups with correct stream names; verifies unknown_service:java truncation and internal attribute cleanup
- `TestAppSignalsLogsNoisyNeighbor` — Invalid service name (colons) doesn't block healthy services; validates provisioner failure is logged
- `TestAppSignalsLogsDefaultPlaceholder` — Missing service.name defaults to unknown_service; missing attrs default to unknown; no invalid log group creation attempted
- `TestAppSignalsLogsRouting` — Routing connector splits logs between batch and no-batch pipelines based on event.name == "aws.service_events.aggregate_profile"
- `TestAppSignalsMetricsRouting` — Routing connector splits metrics between EMF (Latency/Error/Fault) and OTLP monitoring endpoint (ServiceEvents); validates via PromQL query


test/app_signals/service_events_test.go
```diff
@@ -0,0 +1,235 @@
...
+printf '[default]\naws_access_key_id=%s\naws_secret_access_key=%s\naws_session_token=%s\n' "$AKID" "$SAK" "$TOK" | sudo tee ` + onPremCredsFile + `>/dev/null
+# Region for onPrem mode is read from a config file next to the credentials file.
+printf '[default]\nregion = us-east-1\n' | sudo tee ` + onPremCredsDir + `/config>/dev/null`
+	require.NoError(t, common.RunCommands([]string{writeCredsScript}), "Failed to write credentials file")
+
+	// common-config.toml points the agent at the credentials file.
+	require.NoError(t, common.RunCommands([]string{
+		"printf '[credentials]\\n  shared_credential_profile = \"default\"\\n  shared_credential_file = \"" +
+			onPremCredsFile + "\"\\n' | sudo tee " + commonConfigOutput + ">/dev/null",
+	}), "Failed to write common-config.toml")
```
jefchien: Take a look at https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/credential_chain/util/setup_util.go. It does a lot of what this is trying to do.

test/app_signals_service_events/logs_routing_test.go
```diff
@@ -0,0 +1,349 @@
...
+	// Verify no failed log group creation — unhandled nil would produce an invalid name
+	// that the provisioner would fail to create
+	agentLog, _ := os.ReadFile(common.AgentLogFile)
+	assert.NotContains(t, string(agentLog), "Failed to create log group/stream",
+		"Agent should not attempt to create a log group with an invalid name from unhandled nil")
+}
+
+// TestAppSignalsLogsRouting verifies that the routing connector correctly
+// splits logs between the batch and no-batch pipelines:
+//   - 4 regular logs → batch pipeline (same ingestion time)
```
jefchien, line 252: nit: This seems like it could be flaky. Is there something else we can look at instead of ingestion time?

test/app_signals/agent_configs/config_logs_dynamic_placeholders.json
```diff
@@ -0,0 +1,22 @@
+{
+  "agent": {
+    "debug": true,
+    "region": "us-east-1"
```
jefchien: nit: Should we stick with us-west-2 to be consistent with the rest of the tests?

test/app_signals/service_events_test.go
```diff
@@ -0,0 +1,235 @@
...
+	t.Helper()
+	_, err := common.RunCommand(fmt.Sprintf("sudo iptables -A OUTPUT -d %s -j REJECT", imdsEndpoint))
+	require.NoError(t, err, "Failed to block IMDS")
+}
+
+// unblockIMDS removes the iptables rule added by blockIMDS.
+func unblockIMDS(t *testing.T) {
+	t.Helper()
+	_, _ = common.RunCommand(fmt.Sprintf("sudo iptables -D OUTPUT -d %s -j REJECT", imdsEndpoint))
+}
```
jefchien: Instead of blocking with IP tables, consider using the existing `SetupSystemdOverride` with `AWS_EC2_METADATA_DISABLED` set to true.

https://github.com/aws/amazon-cloudwatch-agent-test/blob/c3bb64185e982702101949242419da7721764f50/test/credential_chain/util/setup_util.go#L100-L111

That may not help with the initial translation (during the fetch), but it'll impact the agent startup and runtime.

test/app_signals/service_events_test.go
```diff
@@ -0,0 +1,235 @@
...
+		"printf '[ssl]\\n  ca_bundle_path = \"" + caBundlePath + "\"\\n' | " +
+			"sudo tee " + commonConfigOutput + ">/dev/null",
+	}), "Failed to write common-config.toml")
+
+	common.CopyFile(logsConfigPath, common.ConfigOutputPath)
+
+	defer common.StopAgent()
+	defer common.RunCommand("sudo " + agentCtl + " -a remove-config -c all")
+
+	fetchConfig(t, "ec2")
```
jefchien: Any reason not to use `common.StartAgent`?

test/app_signals/service_events_test.go
```diff
@@ -0,0 +1,235 @@
...
+	// env-config.json, which the agent service picks up.
+	require.NoError(t, common.RunCommands([]string{
+		"printf '[ssl]\\n  ca_bundle_path = \"" + caBundlePath + "\"\\n' | " +
+			"sudo tee " + commonConfigOutput + ">/dev/null",
+	}), "Failed to write common-config.toml")
+
+	common.CopyFile(logsConfigPath, common.ConfigOutputPath)
+
+	defer common.StopAgent()
+	defer common.RunCommand("sudo " + agentCtl + " -a remove-config -c all")
```
jefchien: nit: Don't think this is necessary. It also doesn't clean up the `common-config.toml`, so can create a false sense of a reset.

test/app_signals/service_events_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+	fetchConfig(t, "ec2")
+	time.Sleep(10 * time.Second)
+
+	// The provisioner extension must not fail to start due to custom root CAs.
+	// (Scoped to fatal startup errors — a non-fatal RootCAs warning from other
+	// components, e.g. ec2tagger, does not crash the agent.)
+	agentLog := common.ReadAgentLogfile(common.AgentLogFile)
+	assert.NotContains(t, agentLog, "failed to create CW Logs client",
```
jefchien: This is a very specific point-in-time failure. The `NotContains` test passes even if the CA bundle is ignored. The test's "custom" CA is the same as the system CA, so unless we can see the path that it's reading from, it's hard to know if it's using the custom CA. Is there a stronger signal we can use to verify it works? Take a look at how the CA bundle tests do their assertions.

test/app_signals/service_events_test.go
jj22ee: No, updated to use `common.StartAgent`.

test/app_signals/service_events_test.go
jj22ee: Updated to use `SetupSystemdOverride`, and replaced block/unblock with just `disableIMDS` that returns a cleanup function.

test/app_signals/service_events_test.go
jj22ee: Addressed in https://github.com/aws/amazon-cloudwatch-agent-test/pull/690/changes/52be9907adf2ec57c5443dfd4fb5b249b1422456

test/app_signals/service_events_test.go
jj22ee: Tried to address in: https://github.com/aws/amazon-cloudwatch-agent-test/pull/690/changes/67b9b50e78fd5c7113ef3dc9be71ef68c1e86bcc

test/app_signals_service_events/logs_routing_test.go
jj22ee: Couldn't find a better indicator instead of ingestion time. Since these logs are sent in quick succession, there is a very low chance for the flakiness issue (e.g. logs are in different batches) to occur. At least I haven't run into it yet...

test/app_signals_service_events/logs_routing_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+// --- Helpers ---
+
+func sendOTLPLogs(t *testing.T, serviceName string, numLogs int) {
+	t.Helper()
+	cmd := exec.Command("/bin/bash", "resources/send_otlp_logs.sh",
+		serviceName, fmt.Sprintf("%d", numLogs), testHostName, testInstanceID)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Logf("Warning: send_otlp_logs.sh for %s returned error: %v\nOutput: %s", serviceName, err, string(output))
```
jefchien, line 316: Is there a reason we don't want to fail if the script fails? Why is this just a warning? Won't all the following assertions fail?

test/app_signals_service_events/resources/send_otlp_logs.sh
jefchien, line 1: Have we considered making this go-native instead of using a shell script? Would allow us to handle the errors better. Could even centralize them into a shared package.

jefchien (approved):
Ran the test successfully with the current release of the agent: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/27989918707/job/82840181504

PR #695, opened 2026-05-13
Add integration test for journald

Paamicky:
# Description of the issue
Adds integration tests for the journald log collection feature. Tests validate that the journald pipeline correctly collects, filters, and delivers logs to CloudWatch. 

# Description of changes
- Feature test (test/feature/linux/journald_logs/) - Consolidated test file with 4 test functions validating units, priority, matches, and regex filtering. Uses a single EC2 instance with all 4 collect_list entries. Validates that only matching entries appear in each CloudWatch log stream.
  - Log state test (test/log_state/journald/) - Tests cursor persistence across agent restarts to validate no missing logs
  - Test case generator - Registered journald tests in ec2_linux and ec2_selinux sections with AL2/AL2023 targets.

# License

# Tests
All Journald tests passes:
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/26290578857/job/78162376974


generator/test_case_generator.go
```diff
@@ -421,57 +424,7 @@ var testTypeToTestConfig = map[string][]testConfig{
 			testDir:      "./test/liscsi",
 			terraformDir: "terraform/eks/daemon/liscsi",
 			targets:      map[string]map[string]struct{}{"arc": {"amd64": {}}},
-			instanceType: "i7i.xlarge",
-		},
-		{
-			testDir:      "./test/otel/standard",
```
TravisStark, line 433: should these be deleted?

Paamicky: I rebased, I think I mistakenly deleted

validator/validators/basic/basic_validator.go
```diff
@@ -208,6 +208,16 @@ func (s *BasicValidator) ValidateLogs(logStream, logLine, logLevel, logSource, l
 					if logEventID != "" && strings.Contains(message, logLine) && strings.Contains(message, logEventID) {
 						actualEventCount += 1
 					}
+				case "Journald":
+					if logPriority != "" && strings.Contains(message, logLine) && strings.Contains(message, logPriority) {
+						actualEventCount += 1
```
TravisStark: Should this increment once per match? If all are present this would add 3 to actualEventCount.

validator/validators/basic/basic_validator.go
```diff
@@ -208,6 +208,16 @@ func (s *BasicValidator) ValidateLogs(logStream, logLine, logLevel, logSource, l
 					if logEventID != "" && strings.Contains(message, logLine) && strings.Contains(message, logEventID) {
 						actualEventCount += 1
 					}
+				case "Journald":
```
TravisStark: should this be lowercase (`journald`)? or is `Journald` the expected value

TravisStark: Will this case be reached by the current tests?

Paamicky: `case Journald` block has been removed since no journald test routes through
  BasicValidator.ValidateLogs. All journald tests call awsservice.ValidateLogs. I initially modelled after the windows event but then I realised linux test didn't need the BasicValidator. It is dead code and has been removed entirely

validator/validators/basic/basic_validator.go
Paamicky: fixed, code has been removed entirely

generator/test_case_generator.go
```diff
@@ -94,6 +94,8 @@ var testTypeToTestConfig = map[string][]testConfig{
 			testDir: "./test/log_state/logfile",
 			targets: map[string]map[string]struct{}{"os": {"al2": {}}},
 		},
+		{testDir: "./test/log_state/journald", targets: map[string]map[string]struct{}{"os": {"al2": {}, "al2023": {}}}},
```
okankoAMZ: what is the difference between these 2 tests journald vs journald_logs? Why did we divide it to 2 tests?

test/feature/linux/journald_logs/journald_logs_test.go
```diff
@@ -0,0 +1,188 @@
...
+	environment.RegisterEnvironmentMetaDataFlags()
+}
+
+func TestJournaldUnitsLogs(t *testing.T) {
+	common.CopyFile("agent_config.json", configOutputPath)
+
+	err := common.StartAgent(configOutputPath, true, false)
+	assert.NoError(t, err)
+
+	// Journal entries for cwagent-unit-test.service are reated by Terraform provisioner.
```
okankoAMZ: Typo here, + why not have the entries get generated inside the go code instead of relying on the terraform?

test/feature/linux/journald_logs/journald_logs_test.go
```diff
@@ -0,0 +1,188 @@
...
+	instanceId := awsservice.GetInstanceId()
+
+	if err := exec.Command("logger", "-t", "priority-test", "-p", "user.err", "Database connection error occurred").Run(); err != nil {
+		t.Logf("warning: logger command failed: %v", err)
+	}
+	if err := exec.Command("logger", "-t", "priority-test", "-p", "user.err", "Authentication failed for user").Run(); err != nil {
+		t.Logf("warning: logger command failed: %v", err)
+	}
+
+	time.Sleep(120 * time.Second)
```
okankoAMZ: 2 minutes sound long?

test/feature/linux/journald_logs/journald_logs_test.go
```diff
@@ -0,0 +1,188 @@
...
+	assert.NoError(t, err)
+}
+
+func TestJournaldMatchesLogs(t *testing.T) {
+	common.CopyFile("agent_config.json", configOutputPath)
+
+	err := common.StartAgent(configOutputPath, true, false)
+	assert.NoError(t, err)
+
+	// Wait for journald receiver to initialize
```
okankoAMZ: I understand this is suppose to be a pattern, if it is a pattern we should move it to a function. I also don't understand why we have 2 separate sleeps

generator/test_case_generator.go
Paamicky: This log_state tests cursor persistence. It verifies that after an agent restart, no logs are missed and no logs are duplicated.  They have different infrastructure needs and can't be merged. The log state test needs a continuous log generator and restart logic. The the feature test needs specific log entries and validates filtering behavior.

test/feature/linux/journald_logs/journald_logs_test.go
Paamicky: Typo fixed! I generate logs in the go test and the agent kept missing them in ITAR/CN regions due timing. The units test is different because it needs a real `systemd` service unit name in `_SYSTEMD_UNIT`, which requires an actual registered service, logger alone doesn't set that field. That's why Terraform creates and starts `cwagent-unit-test.service.`

Those entries need to exist in the journal before the agent starts, so the agent can read them at startup. When generated in go after StartAgent, the `journald receiver` may have already advanced past the cursor position and miss them. Terraform runs during provisioning before the test binary executes, guaranteeing the entries are already in the journal when the agent starts.

test/feature/linux/journald_logs/journald_logs_test.go
Paamicky: Agreed! I will extract this to a helper function.

test/feature/linux/journald_logs/journald_logs_test.go
Paamicky: This was when I had timing issues  with ITAR/CN regions that were failing, I will revert to 60s and run test to see it it's sufficient

test/feature/linux/journald_logs/journald_logs_test.go
Paamicky: The first sleep is to allow the journal receiver time to initialize before logs are written. The second one is to allow some time for the agent to send logs to cloudwatch before validation. I did that when I was dealing with timing issue

generator/test_case_generator.go
Paamicky: Also, this is consistent with already existing test structure.

okankoAMZ (dismissed):

TravisStark (dismissed):

generator/test_case_generator.go
```diff
@@ -208,6 +210,8 @@ var testTypeToTestConfig = map[string][]testConfig{
 	testTypeKeyEc2SELinux: {
 		{testDir: "./test/ca_bundle"},
 		{testDir: "./test/cloudwatchlogs"},
+		{testDir: "./test/log_state/journald", selinuxBranch: "mcommey/journald-selinux-policy", targets: map[string]map[string]struct{}{"os": {"al2": {}, "al2023": {}}}},
+		{testDir: "./test/feature/linux/journald_logs", selinuxBranch: "mcommey/journald-selinux-policy", targets: map[string]map[string]struct{}{"os": {"al2": {}, "al2023": {}}}},
```
dricross: Make sure we change these branches before merging to main.

dricross (dismissed):

generator/test_case_generator.go
Paamicky: change made.

dricross (approved):

PR #720, opened 2026-07-09
[test/otel/pernode] Add per-node + zero-step CRD E2E suite and EKS harness

wenegiemepraise:
## Summary
Add the **`test/otel/pernode` integration suite** plus a Terraform EKS harness that validates
the ServiceMonitor/PodMonitor scraping path end to end on a live cluster: per-node target
locality, zero-step CRD bundling, and Target Allocator resilience to missing CRDs.

## What's included
**Suite (`test/otel/pernode/`, build tag `integration`):**
- `per_node_test.go` — every SM/PM series is node-local: `target_node == @resource.k8s.node.name`; targets span ≥2 nodes.
- `crd_bundling_test.go` — SM + PM CRDs are served/established after a plain chart install (harness never installs them separately).
- `ta_resilience_test.go` — TA Deployment Available, no CrashLoopBackOff, 0 restarts on a fresh cluster; and it actually discovers monitors.
- `setup_test.go`, `k8s_helpers_test.go`, `metrics_test.go` — shared clientset/ground-truth, CRD discovery, otelmetrics client bootstrap.
- `resources/workload.yaml` — sm-app/pm-app workloads + SM/PM with the `target_node` relabel + load generator.

**Harness (`terraform/eks/daemon/otel-pernode/`):** provisions EKS, installs the chart with the
per-node-capable operator/TA images, applies the workloads, and runs the suite. Because it
exercises unreleased code it takes `operator_image_*`, `ta_image`, `helm_chart_repo/branch` vars.

## Verification method (per-node)
```
target_node             = scraped POD's node   (SM/PM relabel from __meta_kubernetes_pod_node_name)
@resource.k8s.node.name = scraping AGENT's node (from ${env:K8S_NODE_NAME})
per-node holds  ⇔  for every series, target_node == @resource.k8s.node.name
```

## Dependencies / notes
- Validates the feature PRs (operator per-node + CRD resilience, helm bundling) — most useful
  once those and their custom images are available; tests are skip-guarded so they don't break
  the broader suite where fixtures are absent.
- `go vet -tags integration ./test/otel/pernode/...` is clean against current `main`.


test/otel/pernode/ta_resilience_test.go
```diff
@@ -0,0 +1,92 @@
...
+			if cs.State.Waiting != nil {
+				assert.NotEqualf(t, "CrashLoopBackOff", cs.State.Waiting.Reason,
+					"TA pod %s container %s is in CrashLoopBackOff -- it did not tolerate the CRD state", p.Name, cs.Name)
+			}
+		}
+	}
+
+	// Informational: on a freshly provisioned cluster (the otel-pernode harness)
+	// this should be 0, evidencing the TA never restarted to pick up the CRDs.
+	t.Logf("Target Allocator lifetime container restarts: %d (expected 0 on a fresh harness cluster)", totalRestarts(pods))
```
musa-asad, line 64: The harness installs the chart with CRDs already bundled and then `rollout restart`s the `target-allocator`, so this never opens the missing CRD window it's meant to test, and the restart zeroes the one signal (lifetime restart count). Could we relabel it a smoke test on a bundled install rather than CRD resilience?

test/otel/pernode/crd_bundling_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestServiceMonitorPodMonitorCRDsBundled verifies G1 (zero-step install): after
+// installing the chart with otelContainerInsights enabled, the community
+// ServiceMonitor and PodMonitor CRDs are present and established (served by the
+// API server) with no manual prerequisite. The otel-pernode harness installs the
+// chart onto a cluster that does NOT pre-install these CRDs, so their presence
+// here is attributable to the chart's bundling.
```
musa-asad: `crdServed` only checks the CRDs are served, not that this chart installed them, so a rerun or a preexisting `prometheus-operator` passes even when the chart bundles nothing. Could we assert the CRD carries this release's `app.kubernetes.io/managed-by=Helm` and treat CRD absence as a precondition?

test/otel/pernode/resources/workload.yaml
```diff
@@ -0,0 +1,163 @@
...
+  selector:
+    matchLabels: { app: sm-app }
+  template:
+    metadata:
+      labels: { app: sm-app }
+    spec:
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: kubernetes.io/hostname
+          whenUnsatisfiable: ScheduleAnyway
```
musa-asad: This workload only soft prefers spreading (`ScheduleAnyway`), but `TestPerNodeCoverageAcrossNodes` needs two `target_node`s, so both replicas can land on one node and flake it. Could we force the spread with `DoNotSchedule` or a `kubernetes.io/hostname` `podAntiAffinity`? Same for `pm-app` at line 94.

test/otel/pernode/per_node_test.go
```diff
@@ -0,0 +1,132 @@
...
+// TestPerNodeAllocation verifies the per-node strategy end to end: every scraped
+// series must be collected by the agent running on the SAME node as the scraped
+// pod, i.e. the relabeled target_node equals the scraping agent's
+// @resource.k8s.node.name.
+func TestPerNodeAllocation(t *testing.T) {
+	gt := getGroundTruth(t)
+
+	var results []otelmetrics.MetricResult
+	var usedMetric string
+	for _, m := range perNodeMetrics {
```
musa-asad, line 51: The callers `break` on the first `perNodeMetrics`, and `sm-app`/`pm-app` share metric names, so one path passing greens the test while the other is broken, and `queryWorkloadMetric` never checks the test context so it can hang. Could we filter on `app`/`job` for one series per path and thread a `context.Context` through?

terraform/eks/daemon/otel-pernode/variables.tf
```diff
@@ -0,0 +1,97 @@
...
+
+# --- Allocation strategy under test. ---
+variable "allocation_strategy" {
+  type    = string
+  default = "per-node"
+}
+
+variable "k8s_version" {
+  type    = string
+  default = "1.35"
```
musa-asad: A few defaults are footguns: `k8s_version` `1.35` may be ahead of what EKS offers in a region (so a default apply fails), and the `helm_chart_*` defaults point at upstream `main`. Could we pin a GA version and a real chart ref, and drop the dev registry path in `operator_image_repo`, before this goes public?

terraform/eks/daemon/otel-pernode/main.tf
```diff
@@ -0,0 +1,253 @@
...
+}
+
+# --- Clone the helm-charts checkout that BUNDLES the SM/PM CRDs (G1). ---
+# NOTE: no prometheus-operator CRD install step exists in this harness on
+# purpose -- the SM/PM CRDs must come from the chart alone.
+
+data "external" "clone_helm_chart" {
+  count = var.local_chart_path == "" ? 1 : 0
+  program = ["bash", "-c", <<-EOT
+    rm -rf ./helm-charts
```
musa-asad: `data.external.clone_helm_chart` runs `rm -rf` then `git clone` on every plan, which is a side effect in a data source and can wipe a local checkout in the cwd. Could we move it to a `null_resource` with a `triggers` guard that clones into `path.module` behind a `[ -d ]` check instead?

terraform/eks/daemon/otel-pernode/main.tf
```diff
@@ -0,0 +1,253 @@
...
+#     value, so it is patched onto the AmazonCloudWatchAgent CR directly
+#     (mirrors scripts/deploy-all.sh CUSTOM_TA). ---
+
+resource "null_resource" "patch_cr" {
+  depends_on = [helm_release.aws_observability, null_resource.kubectl]
+  triggers   = { timestamp = timestamp() }
+  provisioner "local-exec" {
+    command = <<-EOT
+      set -e
+      sleep 30
```
musa-asad: `null_resource.patch_cr` waits on fixed `sleep`s instead of readiness (same with `validator` and the workload apply racing CRD creation), so on a slow cluster `kubectl patch` fails not found before the CR exists. Could we switch to `kubectl wait` on the CR and `Established` CRDs plus `kubectl rollout status`?

test/otel/pernode/k8s_helpers_test.go
```diff
@@ -0,0 +1,191 @@
...
+	groundTruthOnce.Do(func() {
+		groundTruth, groundTruthErr = buildGroundTruth()
+	})
+	if groundTruthErr != nil {
+		t.Fatalf("failed to build K8s ground truth: %v", groundTruthErr)
+	}
+	return groundTruth
+}
+
+func buildGroundTruth() (*k8sGroundTruth, error) {
```
musa-asad, line 78: The `groundTruth` and `TestMain` helpers here duplicate the `standard` suite, so any fallback fix has to land twice, and `buildGroundTruth`'s pod `List` and `nodeNames()` are dead code. Could we pull the shared helpers into a common package, drop the dead code, and collapse `crdServed`'s redundant not found branch?

musa-asad (approved):

Aakash-Dantre:
Checklist posted for the PR author to move into the PR description under a `## PR Checklist` heading.

## PR Checklist

- [ ] Commits are squashed into a logical, reviewable set (one commit for a single change)
- [x] Commits and PR description comply with Amazon internal guidelines
- [ ] `make` passes locally (build, unit tests, lint)
- [ ] All GitHub Actions checks on the PR are passing
- [ ] Integration test evidence: link to a passing run of this suite
- [x] New or updated integration test coverage: N/A — this PR is the test coverage
- [x] New functionality has unit tests; bug fixes have a reproducing test
- [x] Config translation changes include updated golden files — N/A, test-only change
- [x] Breaking or customer-visible changes are called out in the PR description — none, test-only

> Note for the description: `helm_chart_branch` defaults to `main`, which the file's own comment says
> to avoid ("Pin to a fixed ref (not a moving branch) for CI"). Until the chart work merges, `main`
> installs a chart without the feature under test.

PR #727, opened 2026-07-20
Sync ManageAgent test document: add set-env action

sky333999:
Adds a 'set-env' action and 'optionalEnvironmentVariable' parameter to the test ManageAgent document, delegating to the ctl set-env action added in aws/amazon-cloudwatch-agent#2211.

Do not merge before #2211 is released — integration tests exercising set-env require an agent build whose ctl supports the action.


sky333999:
Ran the full `ssm_document` suite (all 23 ec2_linux rows + all 7 ec2_windows rows) using a targeted subset workflow run:
- **Agent build**: `add-set-env-ctl-action` (aws/amazon-cloudwatch-agent#2211, merge commit 4efa2d0f)
- **Test branch**: `sync-ssm-document-20260720` (this PR, incl. commit e8bea7b3)
- **Run**: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/29800532672

test/ssm_document/helper.go
```diff
@@ -88,3 +89,122 @@ func VerifyAgentAction(out *ssm.SendCommandOutput, instanceId, documentName stri
...
+}
+
+// VerifyEnvConfigContent reads the agent's env-config.json on the instance by running
+// readCommand through runCommandDocument (AWS-RunShellScript or AWS-RunPowerShellScript)
+// and asserts that the file contains every expected key/value pair. Additional keys in
+// the file are ignored.
+func VerifyEnvConfigContent(runCommandDocument, readCommand, instanceId string, expected map[string]string) error {
+	log.Printf("Verifying env-config.json content via %s", runCommandDocument)
+
+	out, err := awsservice.RunSSMDocument(runCommandDocument, []string{instanceId}, map[string][]string{"commands": {readCommand}})
```
jefchien: Do we need to use a run command to cat the file? Isn't this running on the same instance? Couldn't we save the SSM round trip and just read the file directly?

test/ssm_document/constants.go
jefchien, line 1: Are we going to treat this as our primary `ctl -a set-env` test? If that's the case, should we verify a few more test cases like overwriting a key/value pair in the `env-config.json` and invalid input patterns?

test/ssm_document/ssm_document_unix.go
jefchien, line 1: nit: The unix and windows tests are almost identical. Could we just define the consts in the platform specific files and have the shared test be defined in the helper or a some other shared file?

test/ssm_document/helper.go
sky333999: Yup updated it to just read directly.

jefchien (approved):

PR #730, opened 2026-07-23
Add Azure VM and AKS integration tests for the default OTel config

movence:
# Description of the issue
The agent supports running on non-AWS hosts (Azure VM and AKS) with the default OTel config, authenticating to AWS via web-identity federation — but there was no integration test coverage proving the credential chain and end-to-end telemetry delivery (metrics, logs, traces) to CloudWatch from either environment.

# Description of changes
Adds two new test suites, each provisioned by terraform and dispatched from the companion workflow PR (aws/amazon-cloudwatch-agent#2216):

**Azure VM** (`terraform/azure/vm`, `test/azure/vm`)
- Provisions an Azure VM, installs the built `.deb`, and starts the agent with the default OTel config via `amazon-cloudwatch-agent-ctl -a set-env` + `-a fetch-config -m auto -s -c default:otel`.
- Agent authenticates via the Azure IMDS web-identity chain (`AssumeRoleWithWebIdentity`), with the role's trust policy pinned on both `:aud` and `:sub` (the VM's system-assigned identity principal) — the audience is a tenant-wide Azure resource, so it is not sufficient alone.
- Go test runs on the VM and validates metrics (label-filtered), logs, and traces (Transaction Search `aws/spans` via Logs Insights — the OTLP endpoint requires `xray:PutSpans` and X-Ray trace IDs with Unix seconds in the first 4 bytes).

**AKS** (`terraform/azure/aks`, `test/azure/aks`)
- Provisions an AKS cluster, creates an AWS IAM OIDC provider trusting the cluster issuer (`sub`/`aud` conditions), and deploys the agent as a DaemonSet from the integration-test ECR image using a projected service-account token (audience `sts.amazonaws.com`).
- A load-generator Job sends OTLP metrics/logs/traces to the agent; the Go test runs on the runner and validates all three signals in CloudWatch. Metric validation scopes on a per-run `test_id` datapoint attribute and asserts `cloud.platform=azure_aks`, so it isolates the run *and* covers resource detection rather than echoing back an injected attribute.
- Registers the new `AZUREVM` and `AKS` compute types in `environment/computetype`, and adds `-aksClusterName` to the environment metadata.

# License

# Tests
Latest green run: [30584238599](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30584238599) — test-repo `38d1c2a5`, agent `f039ce4c`, first attempt. Both Azure jobs passed:

| Job | Result | Metrics | Logs | Traces |
|---|---|---|---|---|
| [AzureVM-default-otel](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30584238599/job/91011913246) | PASS (221.66s) | 0.66s | 0.71s | 10.26s |
| [AKS-default-otel](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30584238599/job/91011913315) | PASS (10.92s) | 0.16s | 0.47s | 10.29s |

This run specifically proves the two changes no earlier run could:
- **Traces pass with `CloudWatchAgentServerPolicy` alone.** `xray:PutSpans` was removed from both roles, and both trace assertions pass with zero `AccessDenied` in either job — so that action is not required for the OTLP trace path. The AKS role now has no inline policy at all.
- **`runner_ip` is required in both modules** and both applies succeeded, confirming no caller depended on the old `""` default.

Resolved providers: `azurerm v4.81.0`, `kubernetes v2.38.0` (under `~> 2.0`), `aws v6.57.1`. The **run-level** conclusion reads `failure` for a pre-existing reason unrelated to these jobs — see the note in aws/amazon-cloudwatch-agent#2216; job-level results are the meaningful signal.

The preceding run surfaced a latent ordering bug: four Kubernetes resources named the namespace by string rather than by resource, so `terraform apply` launched them alongside it and whichever lost the race failed with `namespaces not found`. All four now reference `kubernetes_namespace.cwagent`, and this run shows the secret starting 10 ms *after* the namespace completes rather than 3 ms before.

Prior runs: [30459712300](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30459712300), [30113003649](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30113003649), [30048181167](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30048181167), [30029802668](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30029802668), [30018222183](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/30018222183).
- Every assertion demonstrably discriminates: each one both failed (for a verified physical reason) and passed during bring-up. Logs use `AssertLogsNotEmpty` to prevent vacuous passes.
- `go vet` / `gofmt` clean on new packages; `terraform validate` and `terraform fmt -check` clean on both stacks.

# Worth noting
- **`us-east-2` is load-bearing, not arbitrary.** `aws/spans` is only populated where the X-Ray trace segment destination is CloudWatch Logs, which is a per-region setting. Moving these to us-west-2 would fail the trace assertion *and* disturb the App Signals suite that relies on the `XRay` destination there.
- **The VM/AKS asymmetry is intentional.** AKS omits `CWAGENT_ROLE_ARN` so sigv4auth falls through to the default chain; the VM sets it because the translator only wires the web-identity token file when `role_arn` is non-empty. The IAM shape differs for a separate reason: both attach `CloudWatchAgentServerPolicy`, but the VM adds an inline policy for the validation reads because the test runs on the VM under that role, while the AKS test runs on the runner and needs no inline policy.
- Known follow-ups, out of scope: full attribute-shape validation across all three signals, k8sattributes pod-mapping enrichment (load-gen uses hostNetwork), and the no-attribute fallback log routing.


terraform/azure/vm/main.tf
```diff
@@ -0,0 +1,150 @@
...
+    ]
+  }
+
+  # Persist env vars to env-config.json: the agent loads this at startup, making them available to OTel
+  # expandconverter which resolves ${AWS_REGION} and ${CWAGENT_ROLE_ARN} placeholders in the translated YAML.
+  provisioner "remote-exec" {
+    inline = [
+      "export PATH=$PATH:/usr/local/go/bin",
+      "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent -setenv 'AWS_REGION=${var.region}' -envconfig /opt/aws/amazon-cloudwatch-agent/etc/env-config.json",
+      "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent -setenv 'CWAGENT_ROLE_ARN=${aws_iam_role.cwagent.arn}' -envconfig /opt/aws/amazon-cloudwatch-agent/etc/env-config.json",
```
jefchien: nit: Use the ctl script. There's a new `set-env` action https://github.com/aws/amazon-cloudwatch-agent/pull/2211

terraform/azure/vm/main.tf
```diff
@@ -0,0 +1,150 @@
...
+      "sudo USE_DEFAULT_CONFIG=otel /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m auto -s -c default:otel",
```
jefchien: I don't think `USE_DEFAULT_CONFIG=otel` should be necessary or does anything in this case.

terraform/azure/vm/main.tf
```diff
@@ -0,0 +1,150 @@
...
+      # The test binary validates delivery via AWS reads; give it the same web-identity chain the agent uses.
+      "curl -s -H Metadata:true \"http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=${var.azure_token_audience}\" | python3 -c \"import sys,json; print(json.load(sys.stdin)['access_token'])\" > /tmp/azure-identity-token",
+      "export AWS_WEB_IDENTITY_TOKEN_FILE=/tmp/azure-identity-token AWS_ROLE_ARN=${aws_iam_role.cwagent.arn} AWS_REGION=${var.region}",
```
jefchien: nit: I wonder if the token file could just point to the token file that the agent will manage or will the agent remove it during shutdown.

terraform/azure/vm/variables.tf
```diff
@@ -0,0 +1,112 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+#####################################################################
+# AWS side
+#####################################################################
+
+variable "region" {
+  type    = string
+  default = "us-west-2"
```
jefchien: Any reason why we default to `us-west-2` for `azure/vm` and `us-east-2` for `azure/aks`? Don't they both need to be a non `us-west-2` region so we can avoid the transaction search enablement clashing with our existing tests?

test/azure/aks/aks_test.go
```diff
@@ -0,0 +1,168 @@
...
+		if attempt < maxRetries {
+			log.Printf("[AKS_Traces] %v — retrying in %v", testResult.Reason, retryInterval)
+			time.Sleep(retryInterval)
+		}
+	}
+	return testResult
+}
+
+// filterLogLines returns lines containing any of the given substrings (case-insensitive).
+func filterLogLines(text string, substrs ...string) []string {
```
jefchien: Doesn't look like this is ever used. Do we need it for our validation?

test/azure/aks/aks_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func TestAKS(t *testing.T) {
+	t.Run("Metrics", func(t *testing.T) {
+		// host.id is overwritten by resourcedetection with the node's VMSS instance ID,
+		// so scope by the cluster-name resource attribute the load generator sends.
+		group := otlpvalidation.ValidateOtlpMetricsWithLabels(
+			"AKSDefaultOtel", env.Region, []string{"aks_otlp_counter"},
+			map[string]string{
+				"@resource.k8s.cluster.name": env.InstanceId,
```
jefchien: nit: Why is cluster name in the `env.InstanceId`? It looks like we have an `EKSClusterName` (https://github.com/aws/amazon-cloudwatch-agent-test/blob/55232718dac25c8e15a50b47f05ae2ac2d3235e5/environment/metadata.go#L41).

test/azure/aks/aks_test.go
```diff
@@ -0,0 +1,168 @@
...
+	flag.Parse()
+	env = environment.GetEnvironmentMetaData()
+	if env.InstanceId == "" {
+		fmt.Fprintln(os.Stderr, "instanceId flag is required (AKS cluster name) to scope telemetry")
+		os.Exit(1)
+	}
+	os.Exit(m.Run())
+}
+
+func TestAKS(t *testing.T) {
```
jefchien, line 48: Once this is working, we'll likely want to validate the entire shape of the metrics, logs, and traces and verify that the same attributes exist on each.

terraform/azure/aks/variables.tf
```diff
@@ -0,0 +1,107 @@
...
+variable "azure_token_audience" {
+  type    = string
+  default = "https://management.azure.com/"
+}
+
+variable "runner_ip" {
+  type        = string
+  description = "Runner public IP for NSG and kubectl access."
+  default     = ""
+}
```
jefchien, line 41: Doesn't look like this is ever used. Were we going to restrict the [`api_server_access_profile`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#api_server_access_profile-1) using this?

terraform/azure/aks/variables.tf
```diff
@@ -0,0 +1,107 @@
...
+
+variable "cwagent_image_tag" {
+  type        = string
+  description = "Image tag (build_id / commit SHA)."
+}
+
+variable "ecr_docker_config_json" {
+  type        = string
+  sensitive   = true
+  description = "Unused. The pull secret is generated in terraform via aws_ecr_authorization_token; kept so callers passing this variable do not error."
```
jefchien: Not sure I follow. Aren't we the only callers of this module?

terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,469 @@
...
+  curl -sf -X POST "$ENDPOINT/v1/traces" -H "Content-Type: application/json" \
+    -d "{\"resourceSpans\":[{\"resource\":{\"attributes\":[{\"key\":\"service.name\",\"value\":{\"stringValue\":\"$SERVICE_NAME\"}},{\"key\":\"host.id\",\"value\":{\"stringValue\":\"$INSTANCE_ID\"}},{\"key\":\"k8s.cluster.name\",\"value\":{\"stringValue\":\"$INSTANCE_ID\"}},{\"key\":\"k8s.namespace.name\",\"value\":{\"stringValue\":\"amazon-cloudwatch\"}}]},\"scopeSpans\":[{\"scope\":{\"name\":\"aks-otlp-test\"},\"spans\":[{\"traceId\":\"$TRACE_ID\",\"spanId\":\"$SPAN_ID\",\"name\":\"aks-otlp-test-span\",\"kind\":2,\"startTimeUnixNano\":\"$START_NS\",\"endTimeUnixNano\":\"$NOW_NS\",\"attributes\":[{\"key\":\"cluster_name\",\"value\":{\"stringValue\":\"$INSTANCE_ID\"}}]}]}]}]}" || true
+  sleep 10
+done
```
jefchien: Could we separate this out into a separate file for the script so it's more parseable and then reuse the same script for both the AKS and Azure VM tests?

terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,469 @@
...
+      }
+
+      spec {
+        host_network   = true
+        dns_policy     = "ClusterFirstWithHostNet"
+        restart_policy = "Never"
+
+        container {
+          name    = "load-gen"
+          image   = "curlimages/curl:8.8.0"
```
jefchien, line 342: Any concerns with using this image?

terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,469 @@
...
+  ]
+}
+
+#####################################################################
+# Load generator: pushes OTLP to localhost:4318 for 3 min via hostNetwork
+#####################################################################
+# The payloads carry k8s.cluster.name and k8s.namespace.name resource attributes so the
+# agent's k8s logs-routing template produces a deterministic per-cluster destination
+# (/aws/cwagent/<cluster>/otlp, stream amazon-cloudwatch/<service>); resourcedetection
+# only overrides keys it detects (e.g. host.id), so these pass through intact.
```
jefchien: Does that mean we aren't validating that the resource detection is consistent?

terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,469 @@
...
+      identifiers = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"]
+    }
+
+    condition {
+      test     = "ArnEquals"
+      variable = "aws:PrincipalArn"
+      values   = [local.cwagent_role_arn]
+    }
+  }
+}
```
jefchien: I don't think we need to set `CWAGENT_ROLE_ARN` on AKS. `AWS_ROLE_ARN` + `AWS_WEB_IDENTITY_TOKEN_FILE` and the default credential chain should be enough. There isn't a reason to assume a role on top of the role we've already assumed through web identity. Should be able to just omit `CWAGENT_ROLE_ARN`.

terraform/azure/aks/providers.tf
```diff
@@ -0,0 +1,47 @@
...
+    tls = {
+      source  = "hashicorp/tls"
+      version = "~> 4.0"
+    }
+    local = {
+      source  = "hashicorp/local"
+      version = "~> 2.4"
+    }
+  }
+}
```
jefchien, line 29: It looks like we don't typically pin to specific versions. Should we try to match the versions of the ones that do https://github.com/search?q=repo%3Aaws%2Famazon-cloudwatch-agent-test%20required_providers&type=code?

terraform/azure/aks/variables.tf
```diff
@@ -0,0 +1,107 @@
...
+variable "github_test_repo_branch" {
+  type    = string
+  default = "main"
+}
+
+variable "agent_deb_path" {
+  type        = string
+  description = "Local path to the pre-built agent .deb (uploaded into the DaemonSet container image)."
+  default     = ""
+}
```
jefchien: Doesn't look like this is used by the AKS setup. Any reason we want to keep it?

terraform/azure/aks/variables.tf
```diff
@@ -0,0 +1,107 @@
...
+variable "azure_oidc_provider_arn" {
+  type        = string
+  description = "Pre-created AWS IAM OIDC provider ARN for the Azure AD issuer (VM test uses this; AKS creates its own)."
+  default     = ""
+}
+
+variable "azure_token_audience" {
+  type    = string
+  default = "https://management.azure.com/"
+}
```
jefchien: Doesn't seem like we need this for AKS.

terraform/azure/vm/iam.tf
```diff
@@ -0,0 +1,78 @@
...
+      # destination = CloudWatchLogs), which stores 100% of spans there.
+      "logs:StartQuery",
+      "logs:GetQueryResults",
+      "xray:PutSpans",
+      "xray:PutTraceSegments",
+      "xray:PutTelemetryRecords",
+    ]
+    resources = ["*"]
+  }
+}
```
jefchien, line 80: Does this differ from the policy we end up using for AKS? Could we use the agent's managed policy + the permissions we need for validation in a separate inline policy document?

test/azure/vm/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+3. **The VM must have a managed identity and reach IMDS.** Verified: with a system-assigned identity, IMDS
+   mints a JWT with `iss=https://sts.windows.net/<tenant-id>/` (this is the exact `azure_oidc_issuer_url`);
+   without one, the token endpoint returns `Identity not found`. `main.tf` sets `identity { type = "SystemAssigned" }`.
+
+4. **`host.id` == the VM's IMDS `vmId` == ARM `virtual_machine_id`.** Verified equal on a real VM. The azure
+   resourcedetection detector sets `host.id = compute.VMID`, so the test's `-instanceId=<virtual_machine_id>`
+   correctly matches the `@resource.host.id` metric label.
+
+5. **Logs path depends on agent PR #2197 (OPEN).** Until it merges, `awscloudwatchlogsprovisioner` cannot
```
jefchien: This is outdated. 2197 is merged.

test/azure/vm/README.md
jefchien, line 1: Might want to have another pass at this README if we want to keep it. Some of the information is outdated.

terraform/azure/vm/iam.tf
```diff
@@ -0,0 +1,78 @@
...
+      "cloudwatch:ListMetrics",
+      "cloudwatch:GetMetricData",
+      "logs:PutLogEvents",
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:DescribeLogGroups",
+      "logs:DescribeLogStreams",
+      "logs:GetLogEvents",
+      # StartQuery/GetQueryResults let the test binary validate OTLP trace delivery via the aws/spans
+      # log group; the X-Ray OTLP endpoint requires account-level Transaction Search (trace segment
```
jefchien: Is it account-level? I thought it was regional?

test/azure/vm/payloads_test.go
```diff
@@ -0,0 +1,120 @@
...
+// goroutine and read by the test goroutine after the load window closes.
+var traceMu sync.Mutex
+
+// generatedTraceIDs collects the OTLP trace IDs (32 hex chars) emitted during the load window.
+// validateTraces queries the aws/spans log group (Transaction Search) for these exact IDs.
+var generatedTraceIDs []string
+
+// buildTracesPayload emits an OTLP span with X-Ray-compatible trace IDs.
+// X-Ray requires the first 4 bytes of the 16-byte trace ID to be a Unix epoch timestamp (seconds);
+// IDs that violate this are silently dropped during ingestion.
```
jefchien: Is this still true for transaction search?

terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,430 @@
...
+
+  # Terraform drives the cluster over the public API server, so restrict it to the runner that created
+  # it. The block is omitted entirely when runner_ip is unset rather than emitted with an empty list --
+  # an empty authorized_ip_ranges means "open to all", which would read as restricted while being open.
+  dynamic "api_server_access_profile" {
+    for_each = var.runner_ip != "" ? [var.runner_ip] : []
+    content {
+      authorized_ip_ranges = [api_server_access_profile.value]
+    }
+  }
```
jefchien: nit: Should `runner_ip` just be required then?

terraform/azure/aks/main.tf
jefchien: nit: Consider moving the comment into the script instead of in the module.

terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,430 @@
...
+
+# Unlike the VM test, the AKS test binary runs on the runner under the runner's own credentials, so this
+# role needs agent writes only -- no validation reads.
+data "aws_iam_policy_document" "cwagent_permissions" {
+  statement {
+    effect = "Allow"
+    actions = [
+      # Agent write omitted from CloudWatchAgentServerPolicy: the X-Ray OTLP endpoint needs PutSpans,
+      # which is a different action from PutTraceSegments.
+      "xray:PutSpans",
```
jefchien: I don't think this is true. Can you verify this? I've been testing with just the `CloudWatchAgentServerPolicy` and have been able to send traces via OTLP with the permissions it has. Can see traces from my AKS cluster in `aws/spans` with `resource.attributes.k8s.cluster.name` set to my cluster name.

test/azure/vm/README.md
```diff
@@ -0,0 +1,115 @@
...
+the Azure web-identity exchange succeeded. The `cloud.provider=azure` resource label additionally proves the
+telemetry traversed the Azure detection path rather than being attributed to stray AWS credentials.
+
+## Agent prerequisites
+
+- `#2179` Add default OTel config — **merged** (`USE_DEFAULT_CONFIG=otel`)
+- `#2183` Azure platform detection — **merged** (IMDS detection → translation)
+- `#2010` OIDC auth (`oidctoken`) — **merged**
+- `#2197` web-identity for `awscloudwatchlogsprovisioner` — **merged** (log-group creation on Azure VM)
+- `#2211` `set-env` ctl action — **merged** (used to persist `AWS_REGION` / `CWAGENT_ROLE_ARN`)
```
jefchien: Doesn't seem necessary to include this in a README.

test/azure/vm/README.md
```diff
@@ -0,0 +1,115 @@
+# Azure VM integration test
+
+Validates the CloudWatch Agent on a **real Azure VM** running the default OTel configuration
+(`USE_DEFAULT_CONFIG=otel` / `default:otel`). It exercises the full generated OTel pipeline
+(Azure IMDS detection → translation → collector startup) and verifies that metrics, logs, and traces
+reach CloudWatch through the Azure → AWS credential chain (`oidctoken` extension → `sigv4auth` web identity).
+
+## What it checks
+
+The agent is started with no JSON config and `USE_DEFAULT_CONFIG=otel`, so it emits the default OTel
```
jefchien: On Azure VM, we don't use `USE_DEFAULT_CONFIG=otel`.

test/azure/aks/aks_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// k8s.namespace.name resource attributes the load generator sends, so it is unique to
+	// this cluster. The stream is {k8s.namespace.name}/{service.namespace}/{service.name},
+	// where the agent's identity transform fills service.namespace from k8s.namespace.name.
+	// AssertLogsNotEmpty guards against a vacuous pass on an empty window.
+	logGroup := fmt.Sprintf("/aws/cwagent/%s/otlp", env.AKSClusterName)
+	logStream := fmt.Sprintf("amazon-cloudwatch/amazon-cloudwatch/%s", serviceName)
+	marker := fmt.Sprintf("aks_otlp_log_%s", env.AKSClusterName)
+	const maxRetries = 4
+	const retryInterval = 30 * time.Second
+	for attempt := 1; attempt <= maxRetries; attempt++ {
```
jefchien, line 96: nit: Is there a reason we retry 4 times for the AKS validation and only have one attempt for the VM one? Should the VM one also have retries?

test/azure/vm/azurevm_test.go
```diff
@@ -0,0 +1,281 @@
...
+				otlpLogGroup, *stream.LogStreamName, &since, &until,
+				awsservice.AssertLogsNotEmpty(),
+				awsservice.AssertPerLog(awsservice.AssertLogContainsSubstring(marker)),
+			)
+			if err == nil {
+				testResult.Status = status.SUCCESSFUL
+				return testResult
+			}
+			testResult.Reason = err
+		}
```
jefchien: Why are we checking all log streams? Shouldn't we filter it down by the `host.id`? The log stream names are deterministic. The number of log streams will continue to grow for the same log group as new VMs are spun up. Locking it down to a fixed log stream also validates that the log routing is working correctly.

jefchien (approved):

PR #748, opened 2026-08-26
Scope ssm_document to SSM instances; add non-SSM OS variants

bhavya76:
## What
Run the ssm_document integration test only on SSM-enabled instances, and add non-SSM variants of the OSes that previously had no SSM agent.

## Why
ssm_document drives the CloudWatch agent via an SSM document and therefore requires a registered SSM agent. rocky-linux-8, debian-12 and debian-13 image recipes strip the SSM agent at bake time, so the readiness wait times out and the test fails during setup (never reaching the code under test).

## Changes
- Image recipes (out of band, EC2 Image Builder): rocky-linux-8, debian-12, debian-13 now keep the SSM agent (uninstallAfterBuild=false + install component), making SSM the default; new -withoutSSM recipes/AMIs preserve the non-SSM configuration.
- Matrix: add rocky-linux-8-withoutSSM, debian-12-withoutSSM, debian-13-withoutSSM.
- rocky-linux-8 AMI filter tightened so it resolves the SSM image, not the -withoutSSM image.
- ssm_document: excludedOs for the three -withoutSSM variants.

### DIFF
--- a/generator/resources/ec2_linux_test_matrix.json
+++ b/generator/resources/ec2_linux_test_matrix.json
@@ -147,10 +147,43 @@
     "username": "rocky",
     "instanceType":"t3a.medium",
     "installAgentCommand": "go run ./install/install_agent.go rpm",
-    "ami": "cloudwatch-agent-integration-test-rocky-linux-8*",
+    "ami": "cloudwatch-agent-integration-test-rocky-linux-8 20*",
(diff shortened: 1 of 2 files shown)


AshuSrv (commented):
Can we run this branch against the actual agent to test the changes

AshuSrv (approved):

PR #756, opened 2026-09-03
Add EKS Addon install path to OTelCI test

Paamicky:
### Description of the issue
The OTel Container Insights EKS e2e test only covered the Helm install path, not the (now released) EKS add-on path. These test also needed testing with IPv6 which is included in this test.

### Description of changes
Add EKS Addon support to the Container Insights e2e test:
- `ci_node.json`: enable `otelContainerInsights` so the add-on creates the cluster-scraper CR.
- `containerinsights_test.go` (add-on-gated): patch the node CR config with the `test.run.id`-stamped config and set the cluster-scraper image to the build under test.

### License

### Tests
- Full OTELContainerInsights Add-on and IPv6 e2e runs: [OTELContainerInsightsTests](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/34507361916)

>[!Note]
> This PR should merge after we merge this [OCI Impementation](https://github.com/aws/amazon-cloudwatch-agent/pull/2271) agent main

### DIFF
--- a/test/e2e/containerinsights/containerinsights_test.go
+++ b/test/e2e/containerinsights/containerinsights_test.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/aws/amazon-cloudwatch-agent-test/environment"
+	"github.com/aws/amazon-cloudwatch-agent-test/environment/eksinstallationtype"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/e2e"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/e2e/utils"
 	"github.com/aws/amazon-cloudwatch-agent-test/test/otel_collect/otlpvalidation"
@@ -153,6 +154,21 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	// EKS add-on installs the agent via Terraform before the per-run test.run.id
+	// exists, and its image patching only covers the node CR. The Helm path delivers
+	// the stamped node config and the build image through chart values, so mirror
+	// that here by patching the add-on-managed CRs directly.
+	if env.EKSInstallationType == eksinstallationtype.EKS_ADDON {
+		if err := applyNodeConfig(env); err != nil {
+			fmt.Printf("Failed to apply node config: %v\n", err)
+			os.Exit(1)
+		}
+		if err := patchClusterScraperImage(env); err != nil {
+			fmt.Printf("Failed to patch cluster-scraper image: %v\n", err)
+			os.Exit(1)
+		}
+	}
+
 	// Clear pre-rendered otelConfig on the node CR so agent translates json config
 	if err := clearOtelConfig(env, nodeCRName); err != nil {
 		fmt.Printf("Failed to clear node otelConfig: %v\n", err)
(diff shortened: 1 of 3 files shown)


test/e2e/containerinsights/resources/keda_karpenter.yaml
```diff
@@ -64,7 +64,12 @@ spec:
                       self.wfile.write(body)
                   def log_message(self, *a):
                       pass
-              socketserver.TCPServer(("", 80), H).serve_forever()
+              class Server(socketserver.TCPServer):
+                  address_family = socket.AF_INET6
+                  def server_bind(self):
+                      self.socket.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
+                      super().server_bind()
+              Server(("::", 80), H).serve_forever()
```
olowosulu: **[non-blocking]** This dual-stack switch applies to every run, not just IPv6 — `applyKedaKarpenterStubs` is called unconditionally in `TestMain`, so existing IPv4 Helm runs now bind `AF_INET6` too. That's fine on nodes with kernel IPv6 enabled (IPv4-mapped addresses keep it backward-compatible, and the IPv4 Helm job passed in run 33828184956), but if a node AMI ever ships with IPv6 disabled, `socket(AF_INET6, ...)` will fail and the stub serves nothing, breaking KedaMetrics/KarpenterMetrics for all variants. Consider a fallback:

```python
try:
    server = Server(("::", 80), H)
except OSError:
    server = socketserver.TCPServer(("", 80), H)
server.serve_forever()
```

(with `address_family` handling adjusted accordingly), or a comment noting the kernel-IPv6 assumption. Same applies to the second stub below.

test/e2e/containerinsights/containerinsights_test.go
```diff
@@ -193,6 +209,72 @@ func resolveClusterName(env *environment.MetaData) string {
...
+	})
+	if err != nil {
+		return fmt.Errorf("marshaling node CR patch: %w", err)
+	}
+
+	k8ctl := utils.NewK8CtlManager(env)
+	if err := k8ctl.UpdateKubeConfig(resolveClusterName(env)); err != nil {
+		return err
+	}
+	return k8ctl.PatchResource(
```
olowosulu: **[non-blocking]** Both new patches (`applyNodeConfig` here and `patchClusterScraperImage` below) are single-shot with no retry, and a failure exits the suite. On the add-on path the `cloudwatch-agent-cluster-scraper` CR is created when the addon's chart manifests are applied, so if the patch ever runs before the addon has finished applying them, `PatchResource` fails and the whole run dies. The evidence run passed, so this is a flakiness hardening suggestion rather than a bug: a small bounded retry (or a wait for the CR to exist before patching) would make the add-on path resilient to addon-activation timing.

test/e2e/containerinsights/containerinsights_test.go
```diff
@@ -193,6 +209,72 @@ func resolveClusterName(env *environment.MetaData) string {
...
+		"amazoncloudwatchagent",
+		clusterScraperCRName,
+		agentNamespace,
+		utils.PatchTypeMerge,
+		fmt.Sprintf(`{"spec":{"image":"%s"}}`, image),
```
olowosulu: **[nit]** `applyNodeConfig` builds its patch with `json.Marshal`; this one uses `fmt.Sprintf` into a JSON literal. Safe today since the image is `repoURL/repo:tag` with no JSON-special characters, but `json.Marshal(map[string]any{"spec": map[string]any{"image": image}})` would be consistent with the sibling helper and immune to any future value that needs escaping.

olowosulu (commented):
Reviewed the add-on install path in depth (config unwrap, CR patch targets, companion workflow wiring, and the evidence run) — logic checks out. Three non-blocking notes inline.

test/e2e/containerinsights/resources/keda_karpenter.yaml
Paamicky: Make sense fixed.

test/e2e/containerinsights/containerinsights_test.go
Paamicky: Done

test/e2e/containerinsights/containerinsights_test.go
Paamicky: Done.

olowosulu (approved):
All review comments addressed and verified (retry wrapper, IPv4 fallback for the stub listeners, json.Marshal patch). CI green on this head. LGTM.

PR #758, opened 2026-09-10
fix(otel/performance): measure only live agent pods and recalibrate thresholds

bhavya76:
### Issue

`eks_daemon:otel_performance_test` never passed. Two causes:
1. After the image patch + rollout restart, stale/old-ReplicaSet and just-restarted pods linger in the 5-min query window, so per-pod checks and the regression maxima ran on churned data.
2. Thresholds didn't match the current agent, and per-pod CPU is too variable for a tight two-sided band.

### Changes

- Skip series with no data (all-zero/empty) in the window.
- Measure only pods currently Running in `amazon-cloudwatch` (k8s API), dropping stale/old-ReplicaSet pods. Fail-soft: on API error, proceed unfiltered.
- Recalibrate thresholds to observed clean values. Thresholds support a two-sided band (`threshold` +/- per-metric or per-pod `error_bound`) and, alternatively, an explicit per-pod `max` for a one-sided ceiling `[0, max]`.
- Treat CPU as a `max` ceiling for both pod classes. Per-pod CPU swings widely run-to-run on this small cluster, so a two-sided floor false-fails on the low side. Bloat is still caught by the ceiling, and drops are caught by `TestRegressionCheck`; memory stays two-sided.
- Drop cluster-scraper CPU from `TestRegressionCheck` (too noisy cross-commit; still bounded by its ceiling).
- Log a per-class `CALIBRATION` summary for future re-tuning.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the MIT license.


test/otel/performance/performance_test.go
```diff
@@ -218,11 +238,16 @@ func TestPerformanceThresholds(t *testing.T) {
...
 			require.NotEmpty(t, podName, "series is missing the k8s.pod.name resource label")
 
-			// Get the threshold for this specific pod type
-			threshold, podType := getThresholdForPod(podName, metric.PodThresholds)
+			// Get the threshold and effective error bound for this pod type
+			threshold, eb, podType := getThresholdForPod(podName, metric, thresholds.ErrorBound)
 			if podType == "" {
 				continue
 			}
+			if !slices.ContainsFunc(series.Values, func(v float64) bool { return v != 0 }) {
```
Paamicky, line 263: Should we mark the class as seen before skipping? cos an idle live pod could trip the require.True below

terraform/eks/daemon/otel-performance/main.tf
```diff
@@ -363,7 +363,7 @@ resource "null_resource" "validator" {
       echo "Waiting 6.5 minutes for the agent to settle and metrics to propagate (covers the 5-min query window)..."
       sleep 390
 
-      go test -tags integration -timeout 1h -v ${var.test_dir} \
+      go test -tags integration -timeout 10m -v ${var.test_dir} \
```
Paamicky, line 366: This looks like a big drop, does it safely corver the regression test dynamo backoff?

test/otel/performance/k8s_pods_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/util/otelmetrics"
+)
+
+const agentNamespace = "amazon-cloudwatch"
```
Paamicky, line 25: it seems the literal string: 'amazon-cloudwatch' appears twice in the file. 
should we make `agentNSFilter` use the same constant
```suggestion
  // setup_test.go
  agentNSFilter = fmt.Sprintf(`"@resource.k8s.namespace.name"=%q`, agentNamespace)
```

test/otel/performance/performance_test.go
bhavya76: Done: Moved `expectedClasses[podType] = true` above the zero/empty skip, so a live-but-idle pod still marks its class present and won't trip the "no <class> pod series observed" check.

test/otel/performance/k8s_pods_test.go
bhavya76: Done

terraform/eks/daemon/otel-performance/main.tf
bhavya76: Safe — StandardExponentialBackoff caps at MaxElapsedTime=30s (util/awsservice/constant.go), and the test otherwise runs in ~1-2s, so 10m is ~20x headroom over the DynamoDB backoff. The 10m only exists to bound a hung query so it can't hold the local state lock for an hour (the teardown-lock issue).

PR #759, opened 2026-09-10
Add GCE and GKE integration tests for the default OTel config

MarcoDN:
# Description of the issue

The agent supports running on GCP hosts (GCE and GKE) with the default OTel config, authenticating to AWS via web-identity federation — but there is no integration test coverage proving the credential chains and end-to-end telemetry delivery (metrics, logs, traces) to CloudWatch from either environment.

# Description of changes

Adds two new test suites, each provisioned by terraform (companion workflow jobs to follow in aws/amazon-cloudwatch-agent):

**GCE** (`terraform/gcp/gce`, `test/gcp/gce`)
* Provisions a GCE instance with a per-run service account, uploads the runner-built `.deb` over SSH, and starts the agent via `amazon-cloudwatch-agent-ctl -a set-env` + `-a fetch-config -m auto -s -c default:otel`.
* The agent authenticates via the GCP web-identity chain (`AssumeRoleWithWebIdentity`). Google is a built-in AWS web-identity provider, so the per-run role trusts `accounts.google.com` directly — no IAM OIDC provider resource — pinning all three Google condition keys: `:aud` and `:sub` (the service account's unique ID) and `:oaud` (the requested audience).
* The Go test runs on the instance and validates metrics (label-filtered), logs (asserting the derived `{host.id}/{service.name}` stream, so routing is covered rather than just delivery), and traces (the exact generated trace IDs in `aws/spans` via Logs Insights).

**GKE** (`terraform/gcp/gke`, `test/gcp/gke`)
* Provisions a zonal GKE cluster and registers its OIDC issuer as an AWS IAM OIDC provider. The issuer URL is constructed — GKE serves it at a deterministic `container.googleapis.com/v1/.../clusters/...` path that the terraform resource does not export as an attribute.
* Deploys the agent as a DaemonSet from the integration-test ECR image with a projected service-account token (audience `sts.amazonaws.com`); a load-generator Job pushes OTLP metrics/logs/traces to the agent over hostNetwork.
* The Go test runs on the runner and validates all three signals in CloudWatch. Metric validation scopes on a per-run `test_id` datapoint attribute and asserts `cloud.platform=gcp_kubernetes_engine`, so it isolates the run *and* covers resource detection rather than echoing back an injected attribute.
* Registers the new `GCE` and `GKE` compute types in `environment/computetype`, adds `-gkeClusterName` to the environment metadata, and adds `terraform/gcp/README.md` documenting the one-time project/account setup (APIs, CI service-account roles, CI authentication, GitHub configuration, Transaction Search).

# License

# Tests

Both modules verified with complete local `terraform apply` runs — the apply runs the suite via provisioners, so its exit code is the test verdict:

* **GCE**: `TestGCE` PASS in 362.5s — Metrics 0.55s / Logs 0.67s / Traces 151.27s, all generated trace IDs found in `aws/spans` on the first attempt.
* **GKE**: apply fully green — 15 resources created, `TestGKE` Metrics/Logs/Traces all PASS on the first complete run of the module.
* Failure propagation confirmed in an earlier run: a failing subtest exits nonzero through the provisioner and fails the apply.
* `gofmt` and `go vet -tags integration` clean on the new packages; `terraform validate` and `terraform fmt -check` clean on both stacks.

# Worth noting

* **`us-east-2` is load-bearing, not arbitrary.** `aws/spans` is only populated where the X-Ray trace segment destination is CloudWatch Logs, which is a per-region setting; us-west-2 deliberately keeps the `XRay` destination for the App Signals suite that validates through the X-Ray query APIs.
* **The GCE/GKE asymmetry is intentional.** GCE sets `CWAGENT_ROLE_ARN` (the translator only wires the web-identity token file when `role_arn` is non-empty) and carries an inline policy of validation reads because its test runs on the instance under that role. GKE omits `CWAGENT_ROLE_ARN` so sigv4auth falls through to the default credential chain and the projected token, and its role has no inline policy because the test runs on the runner under the runner's credentials.
* **The IAM shapes differ by provider design.** GCE federates the built-in `accounts.google.com` principal with per-run conditions; GKE registers a per-run IAM OIDC provider for the cluster issuer.


test/gcp/gce/payloads_test.go
jefchien, line 1: This is almost entirely the same as https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/test/azure/vm/payloads_test.go. Can we move it into a utility under `test/otel_collect/otlpvalidation` or something like that instead of replicating the entire file and update both the Azure VM and GCE tests to use it?

terraform/otlp_load_generator.sh
jefchien, line 1: This is also practically identical to the AKS `otlp_load_generator.sh` aside from the `aks`/`gke` prefix. It already renders via `templatefile()`, so could we move that one to a shared location and add one `prefix` template variable for those? The same prefix is used for validation so consider using the lowercase `computeType` for the prefix for both generation and validation.

test/gcp/gce/gce_test.go
```diff
@@ -0,0 +1,284 @@
...
+				result = append(result, line)
+				break
+			}
+		}
+	}
+	if len(result) > 50 {
+		result = result[len(result)-50:]
+	}
+	return result
+}
```
jefchien, line 246: These functions are the same as the Azure VM test. Consider extracting to a shared utility package.

terraform/gcp/gce/main.tf
```diff
@@ -0,0 +1,127 @@
...
+      "sudo dpkg -i -E amazon-cloudwatch-agent.deb",
+    ]
+  }
+
+  # Persist env vars with the ctl set-env action: the agent loads env-config.json at startup, making them
+  # available to OTel expandconverter which resolves ${AWS_REGION} and ${CWAGENT_ROLE_ARN} in the
+  # translated YAML. set-env runs before fetch-config so both are set on the first agent start.
+  provisioner "remote-exec" {
+    inline = [
+      "export PATH=$PATH:/usr/local/go/bin",
```
jefchien: nit: Is this necessary? I see we're installing go via apt-get, so won't it be in `/usr/bin` instead?

terraform/gcp/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+Enable the APIs:
+
+```sh
+gcloud services enable compute.googleapis.com iam.googleapis.com \
+  container.googleapis.com --project "$PROJECT_ID"
+```
+
+An existing VPC network and subnetwork must be passed as `gcp_network_name` /
+`gcp_subnetwork_name`; the auto-created `default` network works.
```
jefchien: nit: Seems like we'll want to avoid using the default network if we want to lock it down to only the `runner_ip` since there are some pre-populated firewall rules that will allow any IP (`default-allow-ssh`).

https://docs.cloud.google.com/firewall/docs/firewalls#more_rules_default_vpc

PR #761, opened 2026-09-14
Fix flakey fluent host loggroup for windows

Paamicky:
### Description of the issue
TestFluentLogs is flaky on EKS Windows, failing with fluent log group doesn't exist. The ~55s log-group existence gate is shorter than Fluent Bit's 60s winlog poll interval that creates the Windows host group, so it expires before the group exists.

### Description of changes
Merged the separate 55s existence gate and 5-min validation loop into one poll loop sharing a single 5-min budget (maxRetries = 30). Group creation now gets the full window; existence and validation now share a single 5-min budget. Removed the now unused logStreamRetry const.

### License

### Tests
- go vet ./test/fluent/ passes. Linux behavior unchanged.


olowosulu (approved):
**Summary: LGTM on the code. A few nits, mostly in the PR description.**

I confirmed the root cause independently. Two recent Windows fluent jobs failed with
`fluent_test.go:77: fluent log group doesn't exist: .../host` ([run 34463648785](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/34463648785/job/102830498748), [run 34501591449](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/34501591449/job/103295619710)).
On Windows the `/host` group is fed only by the `winlog` input with `Interval_Sec 60`, and the old gate slept 1+2+...+10 = 55 s, so 55 s < 60 s holds. The merged loop is logically sound: `validated` is reset per group, set only on `err == nil`, and the post-loop `Fatalf` covers all three exhaustion paths.

One bonus fix worth calling out in the description: in the old code, if `GetLogStreams` returned empty for all 30 iterations, the loop exited without any `Fatalf` and the test moved on to the next group (false pass). The new `if !validated` check closes that hole.

### Nits

1. **Trailing sleep on the last iteration** (L74, L81, L122). When `retry == maxRetries-1` every
   non-success branch still sleeps 10 s before falling through to the `Fatalf` at L126. The old code
   fataled before the sleep in the validation branch. Cost is 10 s per failing group. Optional guard:
   ```go
   if retry < maxRetries-1 {
       time.Sleep(10 * time.Second)
   }
   ```
   or switch to a wall-clock deadline (`time.Now().Add(5 * time.Minute)`), which also makes CI cost
   predictable since `GetLogStreams` can itself block up to ~200 s per call on empty results.

2. **"validation budget is unchanged"** is not quite right. Existence, stream, and validation polling now
   share one 30-iteration budget. If the group first appears at iteration k, validation gets 30-k attempts
   (1 attempt if k = 29). The old code guaranteed 30 validation attempts after the gate. In practice the
   Windows harness sleeps 120 s after `kubectl rollout status` before `go test`, so the group usually exists
   at loop start. Suggest rewording the description to "existence and validation now share a single 5-min budget."

3. **"Linux behavior unchanged"** is only true on the success path. `TestFluentLogs` is one function for both
   platforms (`-instancePlatform` only selects the fields map at L59-61), so a never-created group now fails
   after ~300 s instead of ~55 s on Linux too. That is fine, just worth stating.

4. **"Removed the unused logStreamRetry const"**: it was used at L69/L76 by the gate this PR deletes, so it is
   unused only after the change. Suggest "Removed the now-unused ..." Note the same-named
   `logStreamRetry = 20` in `util/awsservice/cloudwatchlogs.go` is a separate const and correctly untouched.

5. **Final `Fatalf` drops the last error** (L126). The per-attempt `t.Logf` at L121 already includes `%v`
   and the harness runs `-v`, so it is still visible. Adding the last `err` to the `Fatalf` would give a
   one-line diagnosis:
   ```go
   var lastErr error
   ...
   lastErr = err
   ...
   t.Fatalf("failed validation for log group %s within %d retries: %v", group, maxRetries, lastErr)
   ```

None of these block merge.

PR #763, opened 2026-09-17
Fix win-11 integration tests

Paamicky:
### Description of the issue
Windows 11 integration tests have failed since the win-11 build AMI was rebuilt (2026-09-07) onto a newer client Windows base. On client SKUs(desktop editions) the Network Interface Card boots into the Public firewall profile, where inbound WinRM (5985) is blocked, so Terraform's `winrm` provisioner can't connect. The test harness has no `WinRM/firewall` setup of its own and relies on the AMI, so every win-11 run fails at connect. Windows Server editions are not affected. give 

### Description of changes
- Add first-boot EC2 user_data (gated to win-11 runs via `var.ami`) in `terraform/ec2/win/main.tf` that sets the network profile to Private, that allows 5985 inbound on all firewall profiles. All other Windows OS runs are unchanged.
- JMX GC fix: `jvm_gc_collection_seconds_sum` now reports after forcing GC with a 32MB heap. Test was flakey and now passes.

### License

### Tests
WinRM now connects and the tests pass:


terraform/ec2/win/main.tf
```diff
@@ -96,7 +116,7 @@ resource "aws_instance" "cwagent" {
   vpc_security_group_ids               = [module.basic_components.security_group, aws_security_group.winrm_runner.id]
   associate_public_ip_address          = true
   instance_initiated_shutdown_behavior = "terminate"
-  user_data                            = length(regexall("/feature/windows/custom_start/userdata", var.test_dir)) > 0 ? data.template_file.user_data.rendered : ""
+  user_data                            = length(regexall("/feature/windows/custom_start/userdata", var.test_dir)) > 0 ? data.template_file.user_data.rendered : (length(regexall("win-11", var.ami)) > 0 ? local.winrm_bootstrap_userdata : "")
```
jefchien, line 119: nit: So this only works because `/feature/windows/custom_start/userdata` isn't run against `win-11` tests (currently only `win-2019`). If it ever was, the win-11 userdata wouldn't be selected and could fail again.

terraform/ec2/win/main.tf
```diff
@@ -88,6 +88,26 @@ module "validator" {
...
+Enable-NetFirewallRule -DisplayGroup "Windows Remote Management" -ErrorAction SilentlyContinue
+Set-NetFirewallRule -Name WINRM-HTTP-In-TCP-PUBLIC -RemoteAddress Any -Enabled True -ErrorAction SilentlyContinue
+Set-Service -Name WinRM -StartupType Automatic
+Get-NetConnectionProfile | Set-NetConnectionProfile -NetworkCategory Private -ErrorAction SilentlyContinue
+$act = New-ScheduledTaskAction -Execute powershell.exe -Argument '-NoProfile -WindowStyle Hidden -Command "Get-NetConnectionProfile | Set-NetConnectionProfile -NetworkCategory Private"'
+$t1  = New-ScheduledTaskTrigger -AtStartup
+$t2  = New-ScheduledTaskTrigger -Once -At (Get-Date).AddMinutes(1) -RepetitionInterval (New-TimeSpan -Minutes 1) -RepetitionDuration (New-TimeSpan -Hours 2)
+Register-ScheduledTask -TaskName "PinPrivateNetworkProfile" -Action $act -Trigger $t1,$t2 -User "SYSTEM" -RunLevel Highest -Force -ErrorAction SilentlyContinue
+</powershell>
+EOT
```
jefchien, line 108: nit: Would prefer if this was fixed in the AMI itself so we don't rely on this runtime fix.

jefchien (approved):

terraform/ec2/win/main.tf
Paamicky: Yes I will create a backlog item and investigate the pipeline build failure so we can fix in the AMI

